Issue #35, opened 2017-12-05
DOC: add SOP for when segmentation fails


jcohenadad:
how to correct, what software, etc.

alzaia:
Done in commit 9ba6dd7ac2c1df9ad4c9d4e24327caada3f9e787.

jcohenadad:
the commit reference to the notebook modif is missing. Please add

alzaia:
Commit for the notebook: 42810475481ee0c0bd51b8879879db412cbb37a2

jcohenadad:
reopened because of this:
https://github.com/neuropoly/axondeepseg/commit/42810475481ee0c0bd51b8879879db412cbb37a2#r26452234

from now on: open a specific PR and we will take it from there

jcohenadad:
fixed in https://github.com/neuropoly/axondeepseg/pull/42

jcohenadad:
my bad, i thought that https://github.com/neuropoly/axondeepseg/pull/42 was merged but it is not. Issue is reopened until PR is merged.

alzaia:
Done In #42

PR #42, opened 2018-01-24
Dev myelin correction: add script + notebook to generate corrected axon mask when user manually corrects the myelin masks

alzaia:
This is related to issue #35 

Readthedoc rendering here: http://axondeepseg.readthedocs.io/en/dev_myelin_correction/


AxonDeepSeg/visualization/generate_axons_from_myelin.py
```diff
@@ -0,0 +1,48 @@
...
+    :param start_visu: first iterations can reach extreme values, start_visu set another start than epoch 0
+    :return: merged and corrected axon+myelin image
+    """
+
+    # read output from axondeepseg and myelin mask corrected by user
+    prediction = imageio.imread(path_prediction)
+    myelin_corrected = imageio.imread(path_myelin_corrected)
+
+    # compute the axon mask from axondeepseg (axon=255, myelin=127, background=0)
+    axon_ads = prediction > 200
```
perone: I would use a constant for these 200 fixed values, or even a function like `extract_axon_ads` that will do this, just to be clear on what this will do.

PR #44, opened 2018-02-03
PR noise simulation

alzaia:
Added notebook for:
- Generation of noise in images (based on different models)
- Batch that, for a given microscopy image, generates noisy images, segments with AxonDeepSeg, saves results in new subfolder, computes and saves performance metrics.

You can check it here: https://github.com/neuropoly/axondeepseg/blob/dev_noise_sim/notebooks/noise_simulation.ipynb

Will be useful to quantitatively assess how AxonDeepSeg models perform when noise types/params are added to input image.


alzaia:
Done. Added plotting code + results.

You can check it here: https://github.com/neuropoly/axondeepseg/blob/dev_noise_sim/notebooks/noise_simulation.ipynb

PR #45, opened 2018-02-08
Implement more complete set of  segmentation validation metrics

alzaia:
- Class with methods to compute: sensitivity, specificity, precision, accuracy, F1 score, pixelwise Dice, Jaccard, FP rate, FN rate, elementwise Dice (for axons), global Hausdorff distance metric (for contour/interface between axons and myelins).
- Added some visualization tools for pixelwise metrics (barplots) for both axon and myelin, and elementwise Dice (for axons).

You can check the notebook with a use case here: https://github.com/neuropoly/axondeepseg/blob/dev_metrics/notebooks/performance_metrics.ipynb


alzaia:
Yeah, with so many useful notebooks, I thought about it. I think we should add a section in the doc when we list all main notebooks and a brief summary for each. I will add that.

alzaia:
Added notebooks info to doc.

PR #46, opened 2018-02-15
Dev morphometrics

alzaia:
New PR for morphometrics extraction from AxonDeepSeg outputs.

Includes:
- Implementation for individual axons morphometrics
- Implementation of aggregate morphometrics for both axon and myelin
- Display tools to create overlays of color-coded axon segmentation on image, plot individual axons distributions for a given metric
- Saves axon morphometrics (list of dictionaries) + aggregate morphometrics (txt)

Link to the notebook: https://github.com/neuropoly/axondeepseg/blob/dev_morphometrics/notebooks/morphometrics_extraction.ipynb

We were doing these tasks with AxonSeg code as for now, so now (almost) everything can be done directly with AxonDeepSeg.


alzaia:
Yes, I agree. Will update that.

alzaia:
Ok, updated.

Issue #63, opened 2018-02-28
Duplication of images on Jupyter notebook


jcohenadad:
The same image is displayed 3x at section 12:
https://github.com/neuropoly/axondeepseg/blob/master/notebooks/morphometrics_extraction.ipynb

alzaia:
It is similar but not the same. The first one is only the axon mask, the 2nd one is axon mask overlay on image, the 3rd is axon mask +myelin overlay on image. 

I can remove the first 2 as they are 'included' in the 3rd.

jcohenadad:
yes pls remove because it is confusing

alzaia:
This was resolved with PR #64

PR #64, opened 2018-02-28
Remove displays for morphometrics in notebook

alzaia:
Fixes issue #63 

You can check the notebook here: https://github.com/neuropoly/axondeepseg/blob/dev_issue63/notebooks/morphometrics_extraction.ipynb


alzaia:
I updated a function in the code, and then re-run the notebook (that was calling that function) to update the display.

jcohenadad:
right, but what I meant is that the title of this PR is misleading ("Remove displays for morphometrics in notebook"), because the change was done on a function called by the notebook, not on the notebook itself, hence if the user decides to use the function outside of the notebook the changes from this PR will also affect behaviour.

Issue #69, opened 2018-03-06
Change output names: axon_mask.png and myelin_mask.png


jcohenadad:
In case the user has two files in the same folder, e.g. image1.png and image2.png, and he runs ADS on both images, then the output segmentations will overwrite results from the first image.

A suggestion is to output the following files:
~~~
xxx_axon.png
xxx_myelin.png
xxx_axon-myelin.png
~~~
with xxx being the name of the input file.

alzaia:
Fixed in PR #71

PR #71, opened 2018-03-06
Axon and myelin filenames when saving as masks

alzaia:
Fixes issue #69 

I think it is more intuitive to use `segmented` instead of `axon-myelin` for the prediction output, but I can modify if you think that `axon-myelin` is better.

For input `xxx` I propose the following in this PR:
```
xxx_segmented.png
xxx_axon.png
xxx_myelin.png
```


jcohenadad:
yes, i would prefer axon-myelin, just to be consistent. If one is called `xxx_segmented.png`, then it implies that the others are not segmented, which would be wrong. 
Another though, is that you also produce maps of axon diameter, myelin thickness, etc. so maybe it would make more sense to be fully explicit and call them:
~~~
xxx_seg-axon.png
xxx_seg-myelin.png
xxx_seg-axonmyelin.png
~~~
and then call the maps:
~~~
xxx_map-axondiameter.png
xxx_map-myelinthickness.png
xxx_map-gratio.png
...
~~~
However, I see another problem: we have two types of maps: (i) one that is directly derived from the segmentation (e.g., seg-axon with colormap added), and (ii) one that is subsampled in order to derive values of fractional volume. So we probably need to discuss these two cases.
Just to be sure: can both i and ii maps be generated with axondeepseg, or is there a dependency with axonseg?

alzaia:
Yes I prefer the fully explicit version with `xxx_seg-axon`, `xxx_seg-myelin` and `xxx_seg-axonmyelin`. I will modify for that format.

Regarding the maps, (ii) is still a dependency with AxonSeg (the only dependency remaining between the 2). I will work on it as soon as I finish writing my thesis. I can add an issue about that.

jcohenadad:
Issue to [previous comment](https://github.com/neuropoly/axondeepseg/pull/71#issuecomment-370983737) is opened [here](https://github.com/neuropoly/axondeepseg/issues/72).

Issue #47, opened 2018-02-16
Clean OSX-related files (e.g., DS_Store)


jcohenadad:
They should not be uploaded to github.
Use .gitignore. See example here:
https://github.com/neuropoly/spinalcordtoolbox/blob/master/.gitignore

mathieuboudreau:
Done in #79

PR #79, opened 2018-04-06
Fix auto-generated mask filenames & minor cleanup of OSX (#47)/temp files.

mathieuboudreau:
Changes:

- ~~Adds Jupyter to the installation requirements~~ **reverted this change, refer to issue #80.
- Removes .DS_Store files that were left over from before the .gitignore rule. (Issue #47)
- Removes the duplicate *.png* extension in the middle of the generate mask filenames. 
    - e.g. AxonDeepSeg.png_seg-myelin.png -> AxonDeepSeg_seg-myelin.png
- Add .gitignore rule for the mask files generated by the axondeepseg_test command.


AxonDeepSeg/visualization/get_masks.py
```diff
@@ -18,9 +18,13 @@ def get_masks(path_prediction):
     
     # get main path
     path_folder, file_name = os.path.split(path_prediction)
```
perone: Not for this PR, but this line is dead code.

AxonDeepSeg/visualization/get_masks.py
```diff
@@ -18,9 +18,13 @@ def get_masks(path_prediction):
...
+
     tmp_path = path_prediction.split('_seg-axonmyelin')
```
perone: This can be changed to: `tmp_path = path_prediction.split('_seg-axonmyelin')[0]`

requirements.txt
```diff
@@ -9,4 +9,5 @@ tensorflow>=1.3.0
 mpld3>=0.3
 tqdm>=4.14.0
 pillow>=3.4.2
-imageio>=2.2.0
\ No newline at end of file
+imageio>=2.2.0
+jupyter>=1.0.0
```
perone: My opinion is that Jupyter shouldn't be required for the `AxonDeepSeg` package as it isn't required for segmentation but only for the notebook examples (@alzaia is this correct ?).

It would be better to add as an extra requirement (`extras_require`), so people can install by request. Another alternative is to document that they need Jupyter to open notebooks.

AxonDeepSeg/visualization/get_masks.py
```diff
@@ -18,9 +18,13 @@ def get_masks(path_prediction):
...
 
     # save the masks
+    if tmp_path[0].endswith('.png'):
```
perone: Makeing the change above, this can be changed to:

```
if '.png' in tmp_path:
    (...)
```
(there is also a function called `os.path.splitext()` that can be used to split the root and extension of the path.

requirements.txt
mathieuboudreau: That's fine – in that case, I feel like the getting started [Getting Started](https://neuropoly.github.io/axondeepseg/documentation.html#getting-started) instructions should be changed. The very first thing you suggest to do after installation is to explore your Jupyter notebooks, and as a user I expected being able to do it out-of-the-box, locally.

I had no instructions to tell me that the Jupyter notebooks are only supposed to be used online (which don't require my installation at all, that I had just verified to pass the tests).

AxonDeepSeg/visualization/get_masks.py
mathieuboudreau: Is there a reason why you feel that the `endswith` string method should be avoided?

I feel the way I implemented is "safer", as within the if statement I then remove the last four string characters. Hypothetically, using your `if` statement, if the string contains a ".png" sequence within the string (not at the end), it will still pass, and behave strangely.

perone: Just a suggestion, feel free to keep as it is, however the way that the extension is being captured isn't the recommended one (a better approach would be to use the correct splits from the `os.path`) instead of doing manual string manipulation.

mathieuboudreau: I agree, that makes sense, thanks @perone !

requirements.txt
alzaia: Yes @perone you are right. Pretty much everything on ADS can be done without the Jupyter notebooks, but these notebooks are definitely useful for both dev/users. I agree with @mathieuboudreau that the first line in the Getting Started section of the doc is confusing, especially considering that the [getting_started notebook](https://github.com/neuropoly/axondeepseg/blob/master/notebooks/getting_started.ipynb) was mostly made for the Binder link so that users can experiment with an example without having to install the package. The important Jupyter notebooks are listed in the [Jupyter Notebooks](https://neuropoly.github.io/axondeepseg/documentation.html#jupyter-notebooks). I will update the doc in a future PR.

mathieuboudreau: @alzaia sound good! For the remaining minor fixes from this branch, do you prefer that I remove the jupyter requirements.txt line and continue with this PR, or would you prefer to cherry-pick them into another future PR?

alzaia: Yes I think we can remove the Jupyter requirement here and you can continue in this PR (just update the title of the PR).

jcohenadad: @alzaia can you please open an issue and reference the issue in this thread so we don't forget. Thanks!

PR #86, opened 2018-04-11
Move the doc rendering to ReadTheDocs

### LINKED ISSUE #43: Remove duplication of docs_source and docs
See: https://github.com/neuropoly/axondeepseg/pull/40#discussion_r163621688

alzaia:
* Removes duplicate of docs/ and docs_source/ folders in the repo
* Now it will be rendered via ReadTheDocs
* The rendering of this PR is here: http://axondeepseg.readthedocs.io/en/dev_issue43/
* The doc badge of ReadTheDocs was also added to the readme

Fixes #43


perone:
I think it would be nice to change the htmls that are served by github on (https://neuropoly.github.io/axondeepseg/) to point to the readthedocs documentation. This is the link in the paper right ?

alzaia:
@perone Yes thats the link used in the paper. So as you suggested I will open a new issue about it.

README.md
```diff
@@ -2,6 +2,7 @@
 
 [![Binder](https://mybinder.org/badge.svg)](https://mybinder.org/v2/gh/neuropoly/axondeepseg/master?filepath=notebooks%2Fgetting_started.ipynb)
 [![Build Status](https://api.travis-ci.org/neuropoly/axondeepseg.svg?branch=master)](https://travis-ci.org/neuropoly/axondeepseg)
+[![Documentation Status](https://readthedocs.org/projects/axondeepseg/badge/?version=v0.4)](http://axondeepseg.readthedocs.io/en/v0.4/?badge=v0.4)
```
jcohenadad: do we need to specify a version? can't we use something like "master"? otherwise we need to remember to update that number if we change the version, no?

alzaia:
@jcohenadad Ok so after discussing about it with @perone, we now set the 'latest' branch of ReadTheDocs to the actual version (which is v0.4). RTD will point out to the v0.4 version of the doc when a user opens the doc. For any future version releases, an admin of the repo would need to add a new tag on the master branch, and then update the default branch and default version fields in the RTD settings like here (I will add the procedure in the wiki):
capture d ecran 2018-04-12 a 15 58 44

For the badge, we will use the 'latest', which will point out to the most recent version of the repo (i.e. v0.4 now).

Moreover, @perone pointed out that it would be interesting (and more useful) to put the changelog in the doc instead of keeping it directly in the repo, so users that consult the doc will get that info as well. What do you think? I will open an issue about it if you like the idea.

jcohenadad:
ok sounds good-- agreed about moving the changelog into RTD

for the info on the wiki, please create an entry: "how to create a new release", with a list of things to do, so that everything is organized at the same place. A bit like [what we did for SCT](https://github.com/neuropoly/spinalcordtoolbox/wiki/versioning#checklist-for-creating-a-new-release).

mathieuboudreau:
*Re: the doc links.

This is happens on the README page if you switch to the dev_issue43 branch.

alzaia:
@mathieuboudreau thanks a lot for the feedback!

- I added you as maintainer on the RTD page of the project.
- Yes you are right, the docs badge and the readme link should point to the same doc (which now is hosted in the RTD). I will update that in this PR.
- I agree that this RTD procedure is not ideal (i.e. having devs go to the RTD options and change default doc branch every time a new version is released). I don't have too much experience on that, maybe @perone has any ideas).
- Regarding Julien's comment about the badge, in this commit: https://github.com/neuropoly/axondeepseg/pull/86/commits/001f2bbc7966d3ce690fa415a7ff5f2d8d182500 now we use the link to the 'latest' and it does need to be changed (latest is going to point to the version we select in the RTD settings).

jcohenadad:
@alzaia:

> Regarding Julien's comment about the badge, in this commit: 001f2bb now we use the link to the 'latest' and it does need to be changed (latest is going to point to the version we select in the RTD settings).

doesn't that contradict the [wiki](https://github.com/neuropoly/axondeepseg/wiki/How-to-create-a-new-release) where it says " change the default branchand default version to the new version (for instance here v4.3)."?

perone:
I looked into the ReadTheDocs [documentation](http://docs.readthedocs.io/en/latest/versions.html#how-we-envision-versions-working), and we confused the `stable` with the `latest`. The `latest` will actually always point to the last *development* code, while the `stable` will point to the latest released version (assuming that we only tag when we release the package version, which makes sense). So we should have a badge pointing to the stable and another pointing to latest (so we have status for both dev and stable versions). And add a link always to the stable (which is the one users will use). We just need to be careful with the [versioning](https://semver.org/), otherwise ReadTheDocs won't be able to parse it correctly and automatically ajust the pointers to the correct versions. By doing that way, we don't need to update docs or readme because they'll always point to the correct versions.

alzaia:
We had an issue with the `stable` version that was pointing to an older tag and not the latest one so we had to delete and re-create the axondeepseg project in RTD. It now seems to work, i.e. the link to the `stable` version (http://axondeepseg.readthedocs.io/en/stable/) is pointing to the last version we have. So based on @perone feedback above, I will update the PR and the wiki and keep you updated when it's ready to be reviewed.

alzaia:
Ok so I made the changes. 

To summarize:

- We now use the `stable`version of the project that points to the latest released version that has a tag. So the doc link is now this one: http://axondeepseg.readthedocs.io/en/stable/.
- I updated the [wiki](https://github.com/neuropoly/axondeepseg/wiki/How-to-create-a-new-release). For every new version (i.e. tag added to the repo), the only thing to do is to change the default branch to the branch of the last tagged version.

perone:
Just to complement: tag the version and also change the version in the `AxonDeepSeg` package. After releasing a version, let's say that we release the 0.5 version, soon after this last commit, the first one should be incrementing the version to 0.6.

mathieuboudreau:
>and also change the version in the AxonDeepSeg package.

@perone I don't think I fully understand this step – could you elaborate? Where would that be done?

perone:
Sorry @mathieuboudreau I wasn't very clear. Everytime a new release is made, we need to generate the package to upload it to the official Python repositories (PyPI), the `setup.py` uses the versioning of the package that is present in the `__init__.py` file inside the `AxonDeepSeg` package. This versions needs to be changed (incremented) after each release tagging.

mathieuboudreau:
@perone Ah I see, that's clear to me now, thanks!

Instructions for this step should likely be added to the [How to create a new release](https://github.com/neuropoly/axondeepseg/wiki/How-to-create-a-new-release) wiki section then.

alzaia:
I updated the wiki.

alzaia:
@jcohenadad I updated the wiki about the following line:
> they are listed in the link above

About the other comment, at that point of the wiki guideline the new version is already tagged in the repository and built by RTD (step 1) so it is not a 'soon-to-be-released' version but it is a released version. The ```stable``` version of the RTD will always point to the latest tagged version of the project, while the ```latest``` version will point to the last dev code in master.

Issue #73, opened 2018-03-16
Generate doc for training procedure for new models 


alzaia:

alzaia:
Addressed in #96

PR #96, opened 2018-06-05
Propose clean training procedure and notebook

alzaia:
This refers to issue #73 

The main contribution here is the following notebook: https://github.com/neuropoly/axondeepseg/blob/dev_issue73/notebooks/training_guideline.ipynb

There were also some param. input/output issues that were corrected.


mathieuboudreau:
@alzaia thanks a lot for making this notebook! 

Where could I find "/Users/Dropbox/SEM_3c_256/training" to test out this notebook? Is it on Duke?

alzaia:
@mathieuboudreau Yes it is on duke: /Volumes/projects/axondeepseg/labelled_datasets/SEM_3c_256

I'm using a training set of 256x256 patches here for testing purposes.

AxonDeepSeg/data_management/input_data.py
```diff
@@ -159,8 +160,8 @@ def __init__(self, trainingset_path, config, type_ ='train', batch_size = 8, pre
         self.epoch_size = len(self.samples_list)
         self.preload_all = preload_all
         self.loaded_data = None
-        self.mean = config['dataset_mean']
-        self.variance = config['dataset_variance']
+        #self.mean = config['dataset_mean']
```
mathieuboudreau: @alzaia Did you mean to commit these commented out lines? I ask because of the commented out print statement above – looks like you were debugging.

alzaia: Yes it was intended, we do not currently use the dataset mean and variance in that way (and it is not very practical because the user would need to compute it and use it as input param. in the config.). So for now we are using the `apply_legacy_preprocess` function of `patch_management_tools.py` instead of the `apply_preprocess` function (we automatically compute the mean and variance but for each patch).

mathieuboudreau: @alzaia Ok – thanks for clarifying!

alzaia:
Thanks a lot @mathieuboudreau for the feedback. I addressed your first 3 points. 

About the 'permission denied' issue, what OS are you using? I haven't seen this issue on OS X so far, but maybe we could add it somewhere in the doc for the user if that happens.

And for the last point, when I was doing this notebook I did 5 trainings of 5 epochs on this 'toy' architecture and in 1 out of the 5 trainings I had a case with only 2 classes in the prediction, but the other ones had a 3-class prediction. So with such a small network and after only 5 epochs, these unstabilities can be expected. I also asked Chris about it and it should not be an issue as it does not always happen. Can you let it train for 10 epochs, or re-run the training for 5 epochs to see if it still gives you the same prediction?

perone:
I did a quick review and I this is what I think that can be improved (I think these things can be done on another PR after merging this as well, there is no need to to everything in this PR):

- Data augmentation parameters are tied to the data, so it is important to show how and where to change the parameters for each kind of augmentation (how much will rotate, how much will shift, etc.);
- How is tensorboard integrated ? How can someone see the data that is being feed into the model during the training, which metrics are being evaluated in the validation set ? I thint it's important to add this information because training requires monitoring metrics for overfitting, etc;

alzaia:
@perone I made the changes you suggested, is this OK?

PR #103, opened 2018-06-08
Argument bug in the global_variables_initializer function in train_network.py

oumayb:
I removed the argument in the global_variables_initializer function in train_network.py, ligne 328.


mathieuboudreau:
@alzaia the branch here `oumayma_issue101` appears to be a branch off your branch `dev_issue73`, which is currently already under a pull request (#96), so maybe we should merge @oumayb 's branch into that one instead of `master`?

alzaia:
@mathieuboudreau @oumayb Yes I think it would be better to submit this PR for the ```dev_issue73``` branch instead.

PR #106, opened 2018-06-09
Added blue/red visualization function for segmented masks

### LINKED ISSUE #102: Add new blue/red display function for the segmentation masks
@oumayb You did a small code for the RGB rendering of segmented masks in ADS (red/blue) 2 weeks ago, we should add it in the code as a new function before we forget. I suggest to put it under the ```visualization```subfolder, called something like ```rgb_rendering_of_mask.py```. It should take as input the grayscale 3-class mask (i.e. output of ADS) and return (and possibly save) the colored red/blue mask.

oumayb:
Fixes #102.
I added a function that takes as input a segmented image and returns it in a red(myelin) and blue (axon) display.


oumayb:
@mathieuboudreau I agree with you, having both the `save_mask` and `writing_path` might be confusing.
By removing the `if` statement, do you suggest the function should always save the resulting mask?

I'd suggest removing the `save_mask` argument like you suggested, and keeping the `if` statement, the new condition being `if writing_path is not None`

mathieuboudreau:
>I'd suggest removing the save_mask argument like you suggested, and keeping the if statement, the new condition being if writing_path is not None

@oumayb sorry I wasn't clear, this is what I meant. (remove save_mask from the function argument **and** remove save_mask from the if statement, not the if statement itself).

oumayb:
@mathieuboudreau Ok great then! Thanks for the suggestion

Issue #91, opened 2018-05-03
AxonDeepSeg is not compatible with Windows


jcohenadad:
What are the limitations and how can we address them?

mathieuboudreau:
Just created a fresh Windows 10 VirtualBox VM and tested it.

- Installed Python 2.7 using the [Anaconda package](https://www.anaconda.com/download/).
- Opened up an Anaconda path session through the `Anaconda Prompt` logo in the App list
    - Had to run it using Administrator privileges to be able to install packages later on.
- Installed pip in using `conda install pip` on the command line.
- Installed git in using `conda install git`on the command line.
- Cloned ADS using the command line and CD'd into it.
- Created a virtual environment using `conda create --name ads_venv`
- Activated it using `conda activate ads_venv`
- Installed the ADS dependencies using `pip install -e .`

No problems yet.

- Run the installation test `axondeepseg_test`, fails, see screenshot below:

![ads](https://user-images.githubusercontent.com/1421029/39585003-5518b5d0-4ece-11e8-99a6-31d494abd2ac.png)

This was my first time running Python in windows, so I don't have any further insights yet on why this might have been thrown considering that all the dependencies seem to have installed fine.

mathieuboudreau:
See: https://stackoverflow.com/questions/3571514/python-why-should-from-module-import-be-prohibited

Not sure why this would fail in Windows but pass in OS/UNIX, I might try importing the modules outside the function definition: https://github.com/neuropoly/axondeepseg/blob/bfb0fe3ae65c97671df2aa01bbbc147135d7fc21/AxonDeepSeg/integrity_test.py#L1-L16

mathieuboudreau:
Hmmmm actually I just double checked, even though I specifically installed Anaconda 2.7, `python -V` tells me I'm using Python 3.6.5, so that might be the source of that error. I'll debug that issue and try the test again.

perone:
I don't see any limitation that would block the use of it on Windows, I think all required packages are compatible. What can happen is some issues with paths/etc, but it requires testing as @mathieuboudreau is doing, it's the only way to fix the issues that pop out.

mathieuboudreau:
**Looks like using Windows + Python 2.7 + Tensorflow is a (the?) major limitation preventing us from Windows compatibility.**

- I explicitly installed a Python 2.7 venv (`conda create -n ads_venv27 python=2.7) and activated it.
[...]

PR #117, opened 2018-07-17
[NEW] Add Sentry bug tracking.

### LINKED ISSUE #115: install sentry
have a look at SCT project for inspiration on how to deploy sentry, and also how to announce opt in/out to the user.

mathieuboudreau:
Resolves #115 

Here's my proposed implementation for integrating Sentry into ADS. As suggested, it's based off of SCT's implementation.

* Installation is now done using `install_ads` script (run with the command: `./install_ads`), which asks the users permission to track bugs.
  * This will need to be updated in ReadTheDocs as well if merged.
* All ADS modules now import a new module, `AxonDeepSeg.ads_util`, which initiates the Sentry bug tracking.
* The Sentry DSN is saved locally in the file `axondeepseg/bin/ads_sentry`, and the new folder `bin` is ignored by git.
  * I couldn't think of a better folder/filename location for it, but open to suggestions.

The one thing I couldn't get to work like SCT is the printed message to the user (when an error is sent to Sentry) that they can opt out by deleting the DSN file, as coded in this block here (which is effectively identical to the SCT implementation):

https://github.com/neuropoly/axondeepseg/blob/e10fc9067eca6a500198bb0e1a7a3a41d02bffee/AxonDeepSeg/ads_utils.py#L48-L67

I tried several different things (e.g. current implementation, calling ads.init_ads() at the \_\_main\_\_ level of segment.py, with/without a server_log_handler function like SCT that imports  SentryHandler from raven (which is currently omitted in the first commit as ADS had no loggin implemented)), but could not get the message to print (raven is never in the listed exit handlers), even though I can see it for SCT by creating some fake bugs.

@zougloub do you have any insight on how you got this feature working for SCT, and how it could be implemented here for ADS?


.gitignore
```diff
@@ -2,6 +2,8 @@
 .DS_Store
 AxonDeepSeg\.egg-info/
 
+bin/
```
zougloub: `/bin` would be more accurate actually, `bin/` can match `bin` directories at any depth

AxonDeepSeg/ads_utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def init_ads():
+    """ Initialize ads for typical terminal usage
+    :return:
+    """
+    init_error_client()
+
+
+def get_dsn_filepath():
```
zougloub: I'm not sure of the value of this filepath indirection (there's a download in the SCT install, but I'm in the process of getting rid of it) compared to just hard-coding it in the code, with an environment variable to override.

AxonDeepSeg/ads_utils.py
```diff
@@ -0,0 +1,91 @@
...
+                    pass
+                old_exitfunc()
+                if sent_something:
+                    print ("Note: you can opt out of Sentry reporting by "
+                           "deleting the file (axondeepseg)/bin/ads_sentry")
+
+            sys.exitfunc = exitfunc
+        except raven.exceptions.InvalidDsn:
+            # This could happen if sct staff change the dsn
+            print 'Sentry DSN not valid anymore, not reporting errors'
```
zougloub: Py3k alert

install_ads
```diff
@@ -0,0 +1,85 @@
...
+    fi
+
+    if [[ ${is_curl} -eq 0 && ! -e $1 ]] ; then
+        curl -o $1 -L $2
+        e_status=$?
+        echo exit status is ${e_status}
+    fi
+
+    if [[ ${is_curl} -ne 0 && ${is_wget} -ne 0 ]] ; then
+        echo You need to install "curl" OR "wget" before you can install the SPINALCORDTOOLBOX
```
zougloub: `SPINALCORDTOOLBOX` remnent

install_ads
```diff
@@ -0,0 +1,85 @@
...
+# USAGE
+# > ./install_ads
+#
+# This is the AxonDeepSeg (ADS) installer.
+#
+# If you do not want the crash reports question to be ask,
+# set ASK_REPORT_QUESTION at installation time like this:
+# >>> ASK_REPORT_QUESTION=false ./install_ads
+
+STATS_DSN="https://www.neuro.polymtl.ca/_media/downloads/stats_dsn_axondeepseg.txt"
```
zougloub: Yeah I'm getting rid of that download in https://github.com/neuropoly/spinalcordtoolbox/commit/32bdbd3477afb8e007cce9f37bbf8c3bd8d2bf97 ...

jcohenadad:
sorry @mathieuboudreau i didn't fully read your PR description (where the Sentry notice to user is mentioned). I tried debugging, and even though no module is found on [this line of code](https://github.com/neuropoly/axondeepseg/pull/117/files#diff-bb7b85302b2b6a61c4fb63cce64c038fR60) (hence `sent_something` keeps being False), something is **actually** sent to Sentry (I checked on the website). So a quick workaround would be to simply get rid of the if statement for now (until we figure out what's the proper way of doing it) and always display what's in [there](https://github.com/neuropoly/axondeepseg/pull/117/files#diff-bb7b85302b2b6a61c4fb63cce64c038fR69). Sorry this is not super helpful...

mathieuboudreau:
@zougloub thanks you very much for the advice w.r.t. the Sentry opt-out warning message and for your thorough code-review 👍 

I'll clean up the specific code-review suggestions you made, and will explore an alternate way to display the Sentry opt-out warning as you suggested.

mathieuboudreau:
@jcohenadad No worries! I'll tinker around with the strategy to display that message a little bit more, and maybe post an issue on their github repo to see if they have a recommended way about it.

.gitignore
mathieuboudreau: 6d24bcf

AxonDeepSeg/ads_utils.py
mathieuboudreau: 7f6b901

install_ads
mathieuboudreau: 6d24bcf

install_ads
mathieuboudreau: 6d24bcf

.gitignore
mathieuboudreau: 7d4795e

AxonDeepSeg/ads_utils.py
mathieuboudreau: I guess I did it this way because I'm not comfortable with the best practice on setting up environment variables in shell installer scripts (or just in general) for multi-platform compatibility (we'd like to try and make ADS Windows compatible soon, see #91). Also, ADS only has a single command-line script, so I'd like tracking to work even for non-command line usage in our case. We also because we don't have an `sct_launcher`-type script that's called prior to the .py files being executed like you do in your SCT command-line bash scripts, which is what I think sets up your ENV variables right?

mathieuboudreau:
Following today's ADS meeting where we decided to include Sentry tracking by default (but keep the warning on how they can opt-out) so that it's included in the pip installation of ADS, it looks like I can resolve @zougloub unhandled question even better by simply going away with the `./install_ads` and `bin/ads_sentry` files altogether, and hardcode the Sentry code here ads_util.py (where users will be indicated to go to remove it for opt-outs). This way I don't need to deal with environment variables either.

perone:
Hi ! The idea is to have a configuration file on the user's home with the user configuration if he decided to opt-out/opt-in for sentry. I recently added code to save and load configuration file from user's home for the [sct_dbtool](https://sct-dbtool.readthedocs.io/en/master/), the code is in [this file](https://github.com/neuropoly/sct_testing_management/blob/master/sct_dbtool/sct_dbtool/setup.py).

mathieuboudreau:
Thanks @perone ! Is there an advantage of doing it that way instead of notifying the user that they can simply change a bool flag in the code, like I've just implemented here:


https://github.com/neuropoly/axondeepseg/blob/565bfd10695298d5912a56845535c797aa17e678/AxonDeepSeg/ads_utils.py#L6-L12


https://github.com/neuropoly/axondeepseg/blob/565bfd10695298d5912a56845535c797aa17e678/AxonDeepSeg/ads_utils.py#L49-L51
I'm just wondering if it's worth creating an entirely new config file just this one setting in our case – the way above seems simpler, but maybe yours is cleaner?

mathieuboudreau:
Or is your idea that the first time they run an ADS script, that setup will appear?

jcohenadad:
i guess the nice thing with the config file is that if people reinstall ADS, or do updates, the desired configuration will not change. Also, if people work with a git version of ADS and modify the code, there will be merge conflicts in case there are future modifications of `ads_utils.py` file.

also, a config file might be useful in the future?

mathieuboudreau:
@jcohenadad Ahh those arguments all make sense to me now - I'll explore that then.

perone:
Besides what Julien commented, there is also the issue of the `setup.py`, I don't know if its possible to ask a question on the `setup.py`, so if we decide to stick with the script installation, we'll lose the `pip install axondeepseg` installation procedure.

AxonDeepSeg/ads_utils.py
```diff
@@ -0,0 +1,72 @@
...
+                    for handler, args, kw in atexit._exithandlers:
+
+                        if handler.__module__.startswith("raven."):
+                            sent_something = True
+
+                except:
+                    pass
+                old_exitfunc()
+
+                print ("Note: you can opt out of Sentry reporting by "
```
mathieuboudreau: @jcohenadad I actually just noticed that this message prints not only for when there was a bug, but when ADS exits successfully as well. I'll have to explore this a bit more. Maybe I should just wait for our Python 3.5x or 6.x upgrade, as the SCT+raven implementation for this feature might work then.

jcohenadad: oops! maybe you can try to reach out to sentry's forum or post on their github? it would be nice to implement a solution directly on this PR.

mathieuboudreau:
@jcohenadad Ok, so I got the opt-out message to work cleanly on the user-side, but using a somewhat dirty trick on the code-side, so if this is not ok I'm completely fine with reverting 56dc600 and exploring other options.

Basically this implementation overrides the raven function that prints the "*Sentry is attempting to send 1 pending error messages*" message, by creating a dummy copy of that function, adding the opt-out message print statement after the one shown above, and then assigning it to the already imported raven module's function.

<details>

https://github.com/neuropoly/axondeepseg/blob/56dc6004d88f952e3d989cd648c3ba6f666be5e7/AxonDeepSeg/ads_utils.py#L9-L59

</details>

<p>

A caveat to this solution is that if ever there are changes in the original raven function, these would also have to be hardcoded in ads_util.py. However, it seems like it's a function with fairly low activity ([as of today, the last change was made in Nov 2017](https://github.com/getsentry/raven-python/blob/a698fc406d2e0f26616f0163b2b22e1a3a5b30f5/raven/transport/threaded.py)), so as long as we fix the raven version requirement to 6.8.0 (which I did) and don't frequently upgrade it, I think we'll be ok.

Bug test output if user opted in:

<details>


```

(ads_venv) mathieuboudreau@brassens:~/neuropoly/github/axondeepseg$ axondeepseg -i test.png -t SEM
WARNING:tensorflow:From /Users/mathieuboudreau/anaconda3/envs/ads_venv/lib/python2.7/site-packages/tensorflow/contrib/learn/python/learn/datasets/base.py:198: retry (from tensorflow.contrib.learn.python.learn.datasets.base) is deprecated and will be removed in a future version.
Instructions for updating:
Use the retry module or similar alternatives.
AxonDeepSeg v.0.5
The path test.png does not exist.
  File "/Users/mathieuboudreau/anaconda3/envs/ads_venv/bin/axondeepseg", line 11, in <module>
    load_entry_point('AxonDeepSeg', 'console_scripts', 'axondeepseg')()
  File "/Users/mathieuboudreau/neuropoly/github/axondeepseg/AxonDeepSeg/segment.py", line 293, in main
    verbosity_level=verbosity_level)
  File "/Users/mathieuboudreau/neuropoly/github/axondeepseg/AxonDeepSeg/segment.py", line 94, in segment_image
    fp.close()
UnboundLocalError: local variable 'fp' referenced before assignment
Sentry is attempting to send 1 pending error messages
Waiting up to 10 seconds
Press Ctrl-C to quit
Note: you can opt out of Sentry reporting by changing the value of bugTracking to 0 in the file AxonDeepSeg/.adsconfig
(ads_venv) mathieuboudreau@brassens:~/neuropoly/github/axondeepseg$

```

</details>

<p>

Bug test output if user opted in:

<details>

```

(ads_venv) mathieuboudreau@brassens:~/neuropoly/github/axondeepseg$ axondeepseg -i test.png -t SEM
WARNING:tensorflow:From /Users/mathieuboudreau/anaconda3/envs/ads_venv/lib/python2.7/site-packages/tensorflow/contrib/learn/python/learn/datasets/base.py:198: retry (from tensorflow.contrib.learn.python.learn.datasets.base) is deprecated and will be removed in a future version.
Instructions for updating:
Use the retry module or similar alternatives.
AxonDeepSeg v.0.5
The path test.png does not exist.
  File "/Users/mathieuboudreau/anaconda3/envs/ads_venv/bin/axondeepseg", line 11, in <module>
    load_entry_point('AxonDeepSeg', 'console_scripts', 'axondeepseg')()
  File "/Users/mathieuboudreau/neuropoly/github/axondeepseg/AxonDeepSeg/segment.py", line 293, in main
    verbosity_level=verbosity_level)
  File "/Users/mathieuboudreau/neuropoly/github/axondeepseg/AxonDeepSeg/segment.py", line 94, in segment_image
    fp.close()
UnboundLocalError: local variable 'fp' referenced before assignment
(ads_venv) mathieuboudreau@brassens:~/neuropoly/github/axondeepseg$

```

</details>

<p>

As for the config-side, that was setup in b5877cf and slightly fixed in 9183a6b. Now, if the config file (AxonDeepSeg/.adsconfig) does not exist when an ADS function is run, the user will be prompted to answer if they want to opt-in or out, and save their answer to that created file. Thanks @perone for the template code for this!

mathieuboudreau:
@jcohenadad Ok thanks! If you're ok with this latest implementation, please make a final "approved review" so that I can go ahead and merge.

PR #121, opened 2018-07-26
[SENTRY] Display full path to config file in Sentry opt-out message

mathieuboudreau:
Updated the Sentry opt-out message from:

`Note: you can opt out of Sentry reporting by changing the value of bugTracking to 0 in the file AxonDeepSeg/.adsconfig`

to display the full path of the location of the file to change to opt-out, e.g.:

`Note: you can opt out of Sentry reporting by changing the value of bugTracking to 0 in the file /Users/mathieuboudreau/neuropoly/github/axondeepseg/AxonDeepSeg/.adsconfig`

since for people who install via `pip`, the folder "AxonDeepSeg" created in their virtalenv folder may be difficult to find.

### DIFF
--- a/AxonDeepSeg/ads_utils.py
+++ b/AxonDeepSeg/ads_utils.py
@@ -43,10 +43,10 @@ def _main_thread_terminated(self):
                 print("Press Ctrl-C to quit")
 
             # -- Function override statement --#
-            configPath = get_config_path()
+            config_path = get_config_path()
             print ("Note: you can opt out of Sentry reporting by changing the "
                    "value of bugTracking to 0 in the "
-                   "file {}".format(configPath))
+                   "file {}".format(config_path))
             # -- EO Function override statement --#
 
             self._timed_queue_join(timeout - initial_timeout)
@@ -62,7 +62,7 @@ raven.transport.threaded.AsyncWorker.main_thread_terminated = _main_thread_termi
 
 def config_setup():
 
-    configPath = get_config_path()
+    config_path = get_config_path()
     
     if 'pytest' in sys.modules:
         bugTracking = bool(0)
@@ -78,13 +78,13 @@ def config_setup():
     if bugTracking:
         print ("Note: you can opt out of Sentry reporting by changing the "
                "value of bugTracking from 1 to 0 in the "
-               "file {}".format(configPath))
+               "file {}".format(config_path))
 
     config = ConfigParser.ConfigParser()
     config.add_section('Global')
     config.set('Global', 'bugTracking', bugTracking)
 
-    with open(configPath, 'w') as configFile:
+    with open(config_path, 'w') as configFile:
         config.write(configFile)
 
     print("Configuration saved successfully !")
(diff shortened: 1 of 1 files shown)


mathieuboudreau:
(FYI: The increase in coverage here is due to the fact that the data_augmentation tools have some randomness hardcoded in them, so different parts of their code are hit depending on the rng. The coverage doesn't change more than a fraction of a percentage point)

jcohenadad:
how about having the config file at `~/.adsconfig`? i think it's pretty common (e.g., .vim, .ssh, .pip, .matlab, .mricron, .ipython, etc.)

mathieuboudreau:
The only reason why I didn't place it there is because I wasn't sure that the folder (`~/`) exists or works in non-Unix systems (e.g. Windows 10, because of #91 which I'm working on). But I could do that here and handle it with an if/else if the system is a non-Unix one in a future revision.

jcohenadad:
ah! i forgot about windows. @perone any advice?

perone:
If you use the code like the one I did on the `sct_dbtool`: https://github.com/neuropoly/sct_testing_management/blob/master/sct_dbtool/sct_dbtool/setup.py#L38

Where it uses the `expanduser()` function, it will work both on windows, linux and macos, so I think it's safe to go with that. You can change the file name `.adsconfig` to `axondeepseg.config` or `axondeepseg.conf`, so when someone sees this file, it will know it's from axondeepseg, otherwise they might get confused with the shortcut `ads`. It is also good to remove the `.` in the beginning of the file, because I'm not sure if windows handles this well, it's more common in *nix systems.

mathieuboudreau:
@perone Ah ha – `os.path.expanderuser()` looks like a great tool – thanks for the tip! I'll explore that!

And will do re: filename for it!

mathieuboudreau:
I've done @perone's suggested changes, and tested it both on OSX and Windows to confirm that it works as intended.

Opt-out messages now read:

OSX: `Note: you can opt out of Sentry reporting by changing the value of bugTracking to 0 in the file /Users/mathieuboudreau/axondeepseg.cfg`

Windows: `Note: you can opt out of Sentry reporting by changing the value of bugTracking to 0 in the file C:\Users\emb61\axondeepseg.cnf`
(my username on that VM is "emb6150" which windows shortens to "emb61")

AxonDeepSeg/ads_utils.py
```diff
@@ -43,9 +43,11 @@ def _main_thread_terminated(self):
                 print("Press Ctrl-C to quit")
 
             # -- Function override statement --#
+            configPath = getConfigPath()
```
perone: You don't have to change this but take a look on PEP8 for naming conventions. The convention for function names is like `get_config_path()` and for the variable it should be like `config_path`.

mathieuboudreau:
@perone you're completely right -  thanks for catching that! It was bound to happen to me considering that I've been switching between a Python and Matlab project all day – thanks!

jcohenadad:
Also is that normal to have .cfg for unix and .cnf for windows?

mathieuboudreau:
@jcohenadad Sorry that was a typo on my part, it's .cfg in both.

PR #150, opened 2018-11-09
training_guideline.ipynb: Few improvements

jcohenadad:
Clarification of notebook usage for training DL model (`training_guideline.ipynb`) + minor refactoring and doc clarification of `apply_model.py`. Modifications include:
- Automatic download of example data (hosted on OSF)
- No more ugly hard-coded path (the whole notebook now runs automatically without having to set local paths).
- Added lighter architecture to accommodate cheaper GPU (e.g. NVIDIA TitanX)


mathieuboudreau:
@jcohenadad 

>Note about data augmentation: For each type of data augmentation, the order needs to be specified if you decide to apply more than one transformation sequentially. For instance, setting da-0-shifting-activate to True means that the shifting is the first transformation that will be applied to the sample(s) during training. The default ranges of transformations are:

This isn't clear to me. What exactly decides the order of the data augmentation? In the config file, three of them are true, so which one goes first (is it in order of the da-#-(description) #?, or the order that they are defined in the config ?) I just remembered that [Python dicts preserve the order of their keys in Python 3.6+](https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6), and I changed the order in the notebook of the data augmentations in a commit above to go 0-1-2-3 (because having numbers + not sorting them was confusing & ugly to me).

If the purpose of the numbers aren't meaningful but their order in the dict is, I can revert the change that I made, but this feature isn't clear to me.

jcohenadad:
@mathieuboudreau i agree this is not clear to me either. We would need to look inside the code, unless @alzaia has some insights?

mathieuboudreau:
@jcohenadad all points in my review are simply suggestions, we could go ahead and merge now if you think these aren't pressing issues. Everything else looks great!

mathieuboudreau:
Re: max_epochs. The training finished after 599 epochs, not the ~666 (=24000/36) that I was expecting. So maybe the equation I suggested isn't completely right. Hmm.

jcohenadad:
All excellent suggestions @mathieuboudreau, thanks!  I think it is worth implementing all of them in this PR. Having a "perfect" notebook can be used as a template reference for refactoring other notebooks.

alzaia:
About the data augmentation order, if I remember well the number in the variable of each data aug. strategy determines the order of the data aug. steps.

It is read during the training here: https://github.com/neuropoly/axondeepseg/blob/003f9fa5874623e0941078e68d0c4839ae3ee8e0/AxonDeepSeg/train_network.py#L52

Which refers to this function that creates a new data aug. dictionary:
https://github.com/neuropoly/axondeepseg/blob/003f9fa5874623e0941078e68d0c4839ae3ee8e0/AxonDeepSeg/train_network_tools.py#L67

AxonDeepSeg/ads_utils.py
```diff
@@ -166,5 +172,42 @@ def excepthook(exctype, value, traceback):
 
     sys.excepthook = excepthook
 
+def download_osf(url_data, zip_filename):
```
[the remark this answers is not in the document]
mathieuboudreau: That makes sense – done!

alzaia:
As far as I remember, there were conflicting config file definitions when I was training models and the way the config files were defining the data aug. params. were not matching the code so I probably made some changes because of that (i.e. I had to either modify the config. file def. to match the code, or modify the code to match the config. file def.). But I can look more into that later and get back to you ASAP. I also think the notation in the left is better.

mathieuboudreau:
@alzaia Ok thanks! Yeah that makes sense. I can see that the code you linked used to expect the way on the left, but then also hardcoded the values using a variable called "transformations", so they weren't being set by the actual config file I think. Probably just a bunch of overlapping fixes made by different devs - I'm sure we can reintroduce the old notation (ideally while keeping retroactive compatibility to other config files with the new notation that are now floating around).

jcohenadad:
>  I'm sure we can reintroduce the old notation (ideally while keeping retroactive compatibility to other config files with the new notation that are now floating around).

good candidate for a new issue :-)

mathieuboudreau:
@jcohenadad Agreed! In that case, I'm going to remove the config validator in the current notebook right now (because it would fail if the "order" is changed), and then you can go ahead and. take one last look at my recent changes before merging.

mathieuboudreau:
@jcohenadad do you know what resolution the dataset (or more specifically, the "Testing" image) is? The segmentation call at the end of the notebook was using the standard SEM size (0.1) as the `acquired_resolution` value (probably because the notebook used to use an SEM dataset), but the segmentation in this branch were working fine regardless, probably because [the default resampled_resolutions value assumes the SEM value](https://github.com/neuropoly/axondeepseg/blob/003f9fa5874623e0941078e68d0c4839ae3ee8e0/AxonDeepSeg/apply_model.py#L175-L179) also (this should be tackled in #152). I just explicitly set resampled_resolutions  to the TEM value in f8fda22, so basically no resampling is going on right now because the pixel size is unknown to me for this dataset.

If you have the exact value(s), we should add it in the data folder (in a pixel size file) and/or to the function call argument in section 2.2 of this notebook; if not, the status quo is sufficient for just this example since the segmentation looks fine (and since we don't give any resolution in the training, I'm assuming no resampling is going on there either?).

notebooks/training_guideline.ipynb
```diff
@@ -311,10 +311,7 @@
     "# In case you want to test the segmentation with a pre-trained model created using this notebook,\n",
     "# uncomment the lines below.\n",
     "#\n",
-    "#if not download_data(\"https://osf.io/u2j9t/?action=download\", \"TEM_3c_512_2018-11-10_21-32-36.zip\")==0:\n",
-    "#     print('ERROR: Data was not succesfully downloaded and unzipped - please check your link and filename and try again.')\n",
-    "#else:\n",
-    "#    print('Data downloaded and unzipped succesfully.')\n",
```
jcohenadad: @mathieuboudreau to simplify, i was thinking of simply adding this model inside TEM_striatum.zip-- less confusing for the users (also, in addition, if one day we decide to change the training dataset we won't forget to also update the model)

jcohenadad:
> @jcohenadad do you know what resolution the dataset (or more specifically, the "Testing" image) is? 

You're right, it is 0.01 µm. A week ago I set it to 0.1 because I was getting a "Resampled image smaller than 512" error with `acquired_resolution=0.01`. Now it is working with 0.01 (and not with 0.1). I have no idea why.

notebooks/training_guideline.ipynb
mathieuboudreau: Sure – that makes sense!

mathieuboudreau:
>You're right, it is 0.01 µm. A week ago I set it to 0.1 because I was getting a "Resampled image smaller than 512" error with acquired_resolution=0.01. Now it is working with 0.01 (and not with 0.1). I have no idea why.

@jcohenadad the reason why you were getting that error is because you were setting `acquired_resolution=0.01`, but not also setting `resampled_resolutions=0.01`, which the default is set to 0.1:

https://github.com/neuropoly/axondeepseg/blob/003f9fa5874623e0941078e68d0c4839ae3ee8e0/AxonDeepSeg/apply_model.py#L175-L179

I think you would agree, this is confusing. The `resampled_resolutions` is meant to indicate the resolution that was used during the training (e.g. resampled to this resolution), and thus the resolution that the image will be resampled to during the segmentation. By not setting it to 0.01, it was using the default value (0.1 – usually for SEM), thus when you set the acquired resolution to 0.01, it was trying to downsample the image to a resolution of 0.1 during segmentation, hence your error (since then the image had less pixels than the patch size). I think, at a bare minimum, there shouldn't be a default value for `resampled_resolutions` to avoid this issue, because why would it be the SEM pixel size over any other. This is also again related to #152.

AxonDeepSeg/ads_utils.py
```diff
@@ -166,5 +172,42 @@ def excepthook(exctype, value, traceback):
 
     sys.excepthook = excepthook
 
+def download_data(url_data, zip_filename):
```
jcohenadad: @mathieuboudreau FYI: https://github.com/neuropoly/spinalcordtoolbox/issues/2056

PR #176, opened 2019-03-06
Implemented axon-wise morphometrics based on output segmentation

### LINKED ISSUE #12: Implement tools for axon-wise morphology metrics extraction based on output segmentation
- similar structure to the AxonSegPy axonlist class structure
- how to extract myelin objects from output
- stats evaluation method : aggregated vs axon per axon

At the end, save the computed metrics as follows:

- add a flag for hdf5 output that contains already computed stats (diameter, circularity, solidity, centroid and g-ratio per axon)

jcohenadad:
This PR aims at providing additional tools to compute axon-wise metrics (as opposed to aggregated metrics), such as inner/outer equivalent diameter, inner/outer area and g-ratio. Fixes #12

Notebook **morphometrics_extraction.ipynb** will be updated after merging this PR: https://github.com/neuropoly/axondeepseg/pull/177
to avoid dealing with conflicts


jcohenadad:
The proposed strategy is to add an optional input argument (myelin mask) to the function: `get_axon_morphometrics(pred_axon, path_folder, pred_myelin=None)`. If present, the output `stats_array` will also include myelin-related metrics. This approach will ensure cross-compatibility. Some unit tests might be revisited, e.g. `test_get_axon_morphometrics_returns_expected_keys()`. @mathieuboudreau what do you think?

jcohenadad:
First developments (https://github.com/neuropoly/axondeepseg/pull/176/commits/b7c4d4b1d0335fd21d1357076534b3c2e8f1ce65) are encouraging:
anim

It is likely that in cases with few touching myelin, the separation of axon objects will be robust enough. Next step will be to compute equivalent diameter on myelin+axon masks.

mathieuboudreau:
That strategy sounds good @jcohenadad ! I'm fine with making changes that aren't retro-compatible (i.e. needing changes in the unit tests), we'll just have to make sure that we revisit and possibly change how those functions/methods are called in Jupyter Notebooks & elsewhere for consistency.

jcohenadad:
we're getting there. Axon-wise `myelin_diam` and `gratio` are now implemented (a6b067dea052210a55ae435861f2931ec4e2a0a8), and a test has been added (`test_get_axon_morphometrics_with_myelin_mask()`). 

Example of the first 20 gratio values for the test image (also shown earlier in this thread):
~~~
0.2359146622232397,
 0.7433760545726621,
 0.31622776601683794,
 0.6937079855607612,
 0.8135515122979159,
 0.4985971149497049,
 0.5283380587223284,
 0.6648318179787979,
 0.6802843475405516,
 0.7961378383833277,
 0.8189999508077318,
 0.4672082855531912,
 0.3824088899852943,
 0.49708452325114083,
 0.3850660489694794,
 0.6150937860864292,
 0.5522965204859437,
 0.6839855680567694,
 0.6675954326127636,
 0.5837689201916736
~~~
So values overall make sense (we should expect gratio of ~0.75), but some gratio values are outlier (e.g. the first one), which I suspect is caused by a wrong approximation of the equivalent diameter. Next, I will explore computing gratio based on the area rather than equivalent diameter.

mathieuboudreau:
@jcohenadad would there be any value in generating an image with simulated axons with known/defined g-ratio value(s), which could then be used not only for your dev testing/optimisation but also for unit testing?

jcohenadad:
> would there be any value in generating an image with simulated axons with known/defined g-ratio value(s), which could then be used not only for your dev testing/optimisation but also for unit testing?

HUGE value. I was thinking of implementing that.

jcohenadad:
interesting: when computing the g-ratio using area (instead of equivalent diameter), we get the exact [same results](https://github.com/neuropoly/axondeepseg/pull/176#issuecomment-470706501):
~~~
0.2359146622232397,
 0.7433760545726621,
 0.31622776601683794,
 0.6937079855607612,
 0.813551512297916,
 0.49859711494970477,
 0.5283380587223283,
 0.664831817978798,
 0.6802843475405514,
 0.7961378383833279,
 0.8189999508077318,
 0.46720828555319116,
 0.3824088899852943,
 0.49708452325114083,
 0.3850660489694794,
 0.6150937860864292,
 0.5522965204859438,
 0.6839855680567695,
 0.6675954326127636,
 0.5837689201916736
~~~

AxonDeepSeg/morphometrics/compute_morphometrics.py
```diff
@@ -1,197 +1,251 @@
...
+        # Deal with myelin
+        if im_myelin is not None:
+            # Find label of axonmyelin corresponding to axon centroid
+            label_axonmyelin = im_axonmyelin_label[int(y0), int(x0)]
+            # TODO: use logger
+            # print(label_axonmyelin)
+            # print('x, y = {}, {}'.format(x0, y0))
+            if label_axonmyelin:
+                # Get corresponding index from axonmyelin_objects list
+                ind_axonmyelin = \
```
mathieuboudreau: @jcohenadad This index calculation doesn't seem to be implemented well. I tried it for an image with three axon, and by debugging with a `print(ind_axonmyelin)` call right after, it only ever returns `0`.

I found this bug by looking at what the myelin thickness/area/gratio was for each of my axons, and found that myelin thickness area was exactly the same for each one (whereas the gratio was different, but that's because it's also using the axon diameter, which is well implemented & different).

Steps I followed:

```python
import numpy as np
import math
import os
import imageio
from sys import platform as _platform
if _platform == "darwin": # Mac OSX
     import matplotlib as mpl
     mpl.use('TkAgg')
import matplotlib.pyplot as plt

from AxonDeepSeg.morphometrics.compute_morphometrics import *

%matplotlib inline

# Change the image and segmentation paths here for your sample:
path_img = '/Users/mathieuboudreau/Downloads/sample282/image.png'
path_pred = '/Users/mathieuboudreau/Downloads/sample282/AxonDeepSeg.png'

# Read paths and compute axon/myelin masks
img = imageio.imread(path_img)
pred = imageio.imread(path_pred)
pred_axon = pred > 200
pred_myelin = np.logical_and(pred >= 50, pred <= 200)
path_folder, file_name = os.path.split(path_img)

print(stats_array[0]['myelin_diam'])
print(stats_array[1]['myelin_diam'])
print(stats_array[2]['myelin_diam'])

print(stats_array[0]['myelin_area'])
print(stats_array[1]['myelin_area'])
print(stats_array[2]['myelin_area'])

print(stats_array[0]['gratio'])
print(stats_array[1]['gratio'])
print(stats_array[2]['gratio'])

# Get axon diameter value of axon object 0
print(stats_array[0]['axon_diam'])
print(stats_array[1]['axon_diam'])
print(stats_array[2]['axon_diam'])
```

Output: 
Capture d’écran 2019-03-09 à 17 51 00

Segmented mask:
AxonDeepSeg

mathieuboudreau: This might also explain the high variance in your listed above g-values.

AxonDeepSeg/morphometrics/compute_morphometrics.py
```diff
@@ -1,197 +1,251 @@
...
+                    [axonmyelin_object.label for axonmyelin_object in axonmyelin_objects].index(label_axonmyelin)
+                myelin_diam = axonmyelin_objects[ind_axonmyelin].equivalent_diameter * pixelsize
+                myelin_area = axonmyelin_objects[ind_axonmyelin].area * (pixelsize ** 2)
+                stats['myelin_diam'] = myelin_diam
+                stats['myelin_area'] = myelin_area
+                stats['gratio'] = np.sqrt(axon_area / myelin_area)
+            else:
+                # TODO: use logger
+                print('WARNING: Myelin object not found for axon centroid [{},{}]'.format(y0, x0))
```
mathieuboudreau: I got this warning for the following image:

ads_mask

`WARNING: Myelin object not found for axon centroid [150.0,800.0]`

I'm believe that it's the image at the top left; will this (myelin) functionality not work for axons objects that are cut off by the boundary? If so, we should document this so that they can be manually removed/considered as outliers by the user.

mathieuboudreau: Side note: The myelin diameter & area were all exactly equal for the three other objects, further confirming a bug from the other comment above.

AxonDeepSeg/morphometrics/compute_morphometrics.py
mathieuboudreau: @jcohenadad some additional details found through debugging

1) I think the following line is supposed to return the index of the myelin object (e.g. 0,1, or 2), but it always returns 0:

https://github.com/neuropoly/axondeepseg/blob/b3577b26945d2edbddd31e2045bf9b9d953e5872/AxonDeepSeg/morphometrics/compute_morphometrics.py#L117-L119

2) `label_axonmyelin` is always equal to 1 (found through debugging/`print()`).

3) Reminder that `label_axonmyelin` is defined as: 

https://github.com/neuropoly/axondeepseg/blob/b3577b26945d2edbddd31e2045bf9b9d953e5872/AxonDeepSeg/morphometrics/compute_morphometrics.py#L112

4) It appears that what the above was meant to return (I think) is the label of the axonmyelin object (e.g. 0,1,2; or something equivelent ?), however, it always returns `1`, because `im_axonmyelin_label` is a binary mask of the image (0 or 1). This is defined here: 

https://github.com/neuropoly/axondeepseg/blob/b3577b26945d2edbddd31e2045bf9b9d953e5872/AxonDeepSeg/morphometrics/compute_morphometrics.py#L73-L74

mathieuboudreau: It appears that the second parameter of `watershed`, the `markers` argument (here defined as `im_centroid`), should be an array of the same size as the `image` (here `-distance`), with the different integer values (the "markers). However here, `im_centroid` is a tuple of arrays of the centroid coordinates (`([724, 1260, 1517], [1115, 400, 1054])`). Maybe that's the issue?

Source for `watershed` argument definitions: http://scikit-image.org/docs/0.7.0/api/skimage.morphology.watershed.html#skimage.morphology.watershed.watershed

mathieuboudreau: Omit my last (deleted) message; I was printing the `ind_centroid` and not `im_centroid`.

mathieuboudreau: Getting closer. I think (part of) the bug lies here:

https://github.com/neuropoly/axondeepseg/blob/b3577b26945d2edbddd31e2045bf9b9d953e5872/AxonDeepSeg/morphometrics/compute_morphometrics.py#L66-L67

`np.zeros_like` copies both the shape **and** type, so since `im_axon` is a bool, so was `im_centroid`, even though later we wanted to assign it labels that are integers greater than 1 (e.g. instead of (0,1,2,3,...), the labels were being overriden by the type to (0,1,1,1...).

mathieuboudreau: Doing ` im_centroid = np.zeros_like(im_axon, dtype=uint8)` solves the first problem I flag. For the simulated image below, I now get all different myelin areas and thicknesses for the 3 out of 4 axons that were identified. For the actual axon image above, since the three axons are touching, I think only two were correctly separated by the watershed, since two of the diameters are identical but one is different (improvement!).

Capture d’écran 2019-03-09 à 23 46 37

mathieuboudreau: Yep, only two `axonmyelin_objects` were identified using that watershed run:

im_axonmyelin_label

mathieuboudreau: I know there's likely not an easy way to ensuring that these are always separated perfectly, but maybe we should make a condition where if `axonmyelin` object centroids are too far from the `axon` object centroids, to omit these automatically (because otherwise, the metrics will get screwed up for a variety of reasons), maybe defining them (the myelin metrics) as `Null` for these cases. What do you think @jcohenadad ?

AxonDeepSeg/morphometrics/compute_morphometrics.py
mathieuboudreau: It's actually the top right one that wasn't being labelled as an `axonmyelin` object, weird...:

im_axonmyelin_label

mathieuboudreau: Ah ha, no, it actually does get labelled, but as the value 0. Found out by changing the value of the label `i` to `i+1` here:

https://github.com/neuropoly/axondeepseg/blob/b3577b26945d2edbddd31e2045bf9b9d953e5872/AxonDeepSeg/morphometrics/compute_morphometrics.py#L68-L70

See:

im_axonmyelin_label

And since that object was being labelled as `0`, this line was being skipped for that object:

https://github.com/neuropoly/axondeepseg/blob/b3577b26945d2edbddd31e2045bf9b9d953e5872/AxonDeepSeg/morphometrics/compute_morphometrics.py#L116

Looks like my solution above (`im_centroid[ind_centroid[0][i], ind_centroid[1][i]] = i + 1 `) works for me; is it clean enough for you?

Results:

Capture d’écran 2019-03-10 à 00 20 21

AxonDeepSeg/morphometrics/compute_morphometrics.py
mathieuboudreau: Actually; by simply resolving the issue in my comment below (i.e. setting `im_centroid[ind_centroid[0][i], ind_centroid[1][i]] = i + 1 `), the axons (strangely) now appear to have different values:

im_axonmyelin_label

Capture d’écran 2019-03-10 à 00 29 11

mathieuboudreau:
I pushed two fixes for the bugs I identified in the threads above; let me know if you agree with them Julien, I'll continue my review once I get confirmation.

AxonDeepSeg/morphometrics/compute_morphometrics.py
mathieuboudreau: p.s., the g-ratios are almost dead-on the simulated values here (defined as 0.7 for the first 3, then 0.9, per the settings described in #179, where the default g-ratio is 0.7). 🎉

mathieuboudreau:
(last commit) Has to increase `im_centroid` to 16 bit, as if an image had more than 255 axons, the 256th would be defined as object 0 and so on. We could even go to 32 bit if we believe there (or would want to handle) would be a full slice processed (e.g. a single image with more than 65535 axons).

jcohenadad:
@mathieuboudreau thanks a lot for fixing the bugs! the gratio from the test image looks healthier:
~~~
[0.5720213986953382,
 0.7433760545726621,
 0.31622776601683794,
 0.6937079855607612,
 0.813551512297916,
 0.49859711494970477,
 0.5283380587223283,
 0.664831817978798,
 0.6802843475405514,
 0.7961378383833279,
 0.8189999508077318,
 0.5204164998665332,
 0.3824088899852943,
 0.49708452325114083,
 0.3850660489694794,
 0.6150937860864292,
 0.5522965204859438,
 0.6839855680567695,
 0.6675954326127636,
 0.631365224924603]
~~~

mathieuboudreau:
Thanks! Although, looking at those g-ratio numbers side-to-side, only axons No. 1, 12, and 20 are different between before/after my bug fixes, strangely... All the others are identical, hmm.

jcohenadad:
printing the centroids for each loop, it looks like the algorithm correctly selects the myelin associated with the axon: 
~~~
axon centroid: (5.280777537796976, 116.99352051835854)
myelin centroid: (9.869964664310954, 117.62685512367491)
---
axon centroid: (38.27070925825663, 375.78758346868796)
myelin centroid: (42.027525680662215, 377.76054652438415)
---
axon centroid: (2.6486486486486487, 501.56756756756755)
myelin centroid: (6.021621621621621, 501.9972972972973)
---
axon centroid: (16.26086956521739, 542.5063938618925)
myelin centroid: (18.56, 541.8356923076923)
---
axon centroid: (45.26150139881877, 692.7601802921977)
myelin centroid: (52.13866886122827, 692.8052155128073)
---
axon centroid: (5.948770491803279, 885.405737704918)
myelin centroid: (12.269994905756494, 894.8782475802343)
---
axon centroid: (5.04029304029304, 1132.4249084249084)
myelin centroid: (9.294478527607362, 1131.8691206543967)
---
axon centroid: (31.46802030456853, 1195.492385786802)
myelin centroid: (36.65761723132152, 1192.469374018398)
---
axon centroid: (17.741233373639663, 1242.8355501813785)
myelin centroid: (19.461667599328482, 1247.0044767767208)
~~~

mathieuboudreau:
Ok! I have absolutely no idea why so few g-ratios changed with the bug fixes, however I just tested with some generated axons so that the g-ratios and axon diameters (actually, radius) are known and ideal, and the (rounded) results match perfectly. So I'm quite confident that this code the working as expected.

Code to simulate axons (not that the first parameter is the radius, not diameter):

Capture d’écran 2019-03-12 à 23 05 05

Simulated axons:

Capture d’écran 2019-03-12 à 23 04 38

Segmented g-ratio and axon diameter results:
Capture d’écran 2019-03-12 à 23 04 38

Issue #29, opened 2017-09-07
Better exception management for the software


maxwab:
The software lacks some exception management right now.

The idea is to manage failures to find a file, a model ... And to notify the user of the exact error.

Ideally the user should never encounter a Python error when launching the software axondeepseg.

mathieuboudreau:
Related to issue #76 (Implement more complete testing module), the exception handling should be tested for as well.

mathieuboudreau:
* [ ] Add exception handling for case in issue #29

mathieuboudreau:
Stale / been done piecewise over the years, it's much better than when this issue was opened

PR #189, opened 2019-03-13
Standardise how paths are handled 

### LINKED ISSUE #154: [Refactor] Standardise how paths are handled
Currently, some paths are handled with string manipulations,

https://github.com/neuropoly/axondeepseg/blob/003f9fa5874623e0941078e68d0c4839ae3ee8e0/AxonDeepSeg/data_management/input_data.py#L146

whereas others use the more OS-agnostic `os.path.join()` calls

https://github.com/neuropoly/axondeepseg/blob/003f9fa5874623e0941078e68d0c4839ae3ee8e0/test/test_config_tools.py#L126

but these are also a bit messy to maintain and to read from a users point-of-view. Since Python 3.4, a better way (in my opinion) to handle paths was added, which is using Path objects of the pathlib (https://medium.com/@ageitgey/python-3-quick-tip-the-easy-way-to-deal-with-file-paths-on-windows-mac-and-linux-11a072b58d5f). 

The objects usually are handled correctly by functions (e.g. you can give a Path object to print, and it will convert it to a string to print), however it doesn't work well with string manipulations (e.g. you can't do `path_object + "Train/"`, you should instead do `path_object / "Train"` ). I ran into this error in #150 when giving Path objects to `train_model()` (fixed by converting them to strings first).

maf88:
Fix #154. To standardize how we handle path, we are refactoring the code to use the  [`pathlib` library](https://docs.python.org/3/library/pathlib.html). The recommended python version for using `pathlib` is any higher than 3.5, but it can work with Python 2.7.

Documentation is here: https://github.com/neuropoly/axondeepseg/wiki/How-to-handle-paths


mathieuboudreau:
@maf88 A lot of the line changes are unrelated to the main purpose of this PR. e.g. a lot of styling differences, plus a new unrelated function (e.g. `download_zip_file_in_tmp_folder()`). As discussed to today's meeting, we should aim to only change the specific lines that handle paths and no other lines.

See here: https://github.com/neuropoly/axondeepseg/pull/189/files

mathieuboudreau:
Are you doing all these changes manually, or is this some automated tool that fixes all the styling? If the latter, you may need to turn it off.

maf88:
@mathieuboudreau I agree, unfortunately I already made the changes for that file prior to the meeting. The next commits are going to be solely about path.

For reformatting, I launch manually `black myfile.py`. Generally, that's a specific commit just before pushing.

mathieuboudreau:
@maf88 I reverted a lot of your styling changes, and made the branch up to date to master, so make sure to pull before you make more changes.

AxonDeepSeg/integrity_test.py
```diff
@@ -17,20 +17,20 @@ def integrity_test():
...
 
         # get path of directory where AxonDeepSeg was installed
-        dir_path = os.path.dirname(os.path.abspath(__file__))
+        dir_path = Path(__file__).resolve().parent
 
         # input parameters
 
-        path = os.path.join('folder_name', 'file_name')
-        path_testing = os.path.join(dir_path, 'data_test')
+        path = Path('folder_name') / 'file_name'
```
jcohenadad: this will break in windows

maf88:
A small list of the new ways to hande paths: 
- `path = "/home/img.png" is now `path = Path("/home") / "img.png"`
- Get current file absolute path: `file_path = Path(__file__).resolve().parent`
-  Before to separate all elements of a path: `a_path.split(os.sep)`, now ` p=a_path.parts`
- Convert `Path` object to string: `str(Path(“.”))`
- Iter a current directory:  `a_path.itertools()`
- Get extension : a_path.suffix, get filename without extension : a_path.stem
- Regex: `a_path.match(“.png”)`
- Test if the path is a directory and it exists:  `a_path.is_dir()`
- Delete a file: `path_to_a_file.unlink()`, delete a directory tree `shutil.rmtree(pathdir)`

AxonDeepSeg/testing/launch_performance_metrics.py
```diff
@@ -1,37 +1,30 @@
-from skimage import measure
-from skimage.measure import regionprops
+"""
+#TODO: Docstring
```
mathieuboudreau: We probably shouldn't dirty the code with to-do lists; that's what Github issues are for

test/data_management/test_dataset_building.py
```diff
@@ -1,166 +1,109 @@
...
-        self.testPath = os.path.split(self.fullPath)[0]
+        self.testPath = self.fullPath.parent
 
-        self.rawPath = os.path.join(
-            self.testPath,
-            '__test_files__',
-            '__test_patch_files__',
-            'raw'
+        _create_new_test_folder = (
+            lambda s: self.testPath
```
mathieuboudreau: Is lambda syntax really necessary to concate multiple Paths with `/` like we do with join?

mathieuboudreau: Oh I see what you did now

maf88:
Documentation is here: https://github.com/neuropoly/axondeepseg/wiki/How-to-handle-paths

mathieuboudreau:
@maf88 Please add a small description of these changes to the changelog for the upcoming version (2.2) both [here](https://github.com/neuropoly/axondeepseg/blame/f769c146cc80e61ffdd2004872ec444b4806f57a/CHANGELOG.md#L13) and [here](https://github.com/neuropoly/axondeepseg/blame/f769c146cc80e61ffdd2004872ec444b4806f57a/docs/source/documentation.rst#L20).

AxonDeepSeg/apply_model.py
```diff
@@ -1,469 +1,473 @@
...
-	n_classes = config_dict["n_classes"]
-
-	# STEP 1: Load and rescale the acquisitions, and transform them into patches.
+from .visualization.get_masks import get_masks
+from .patch_management_tools import im2patches_overlap, patches2im_overlap
+from .config_tools import update_config, default_configuration
 
-	rs_acquisitions, rs_coeffs, original_acquisitions_shapes = load_acquisitions(
-		path_acquisitions, acquisitions_resolutions, resampled_resolutions, verbose_mode=verbosity_level)
+def apply_convnet(path_acquisitions, acquisitions_resolutions, path_model_folder, config_dict, ckpt_name='model',
```
mathieuboudreau: @maf88 your script that auto-formats styles seems to have been applied to this file, changing every single line. It makes it difficult for me to see what lines you actually changed content (re: paths) vs. formatting (e.g. spaces/tabs) - is there any way you can reverse that? 

Also please turn off any such automated scripts on your computer for the future - if you want to change formatting, a branch should be made and it should be done to all the files at ones, that PR would be easy to review and apply if all that's done was a formatting script. That script should be acknowledged and shared in the wiki too, so that devs don't have competing style formatting scripts.

AxonDeepSeg/data_management/dataset_building.py
```diff
@@ -53,15 +57,15 @@ def raw_img_to_patches(path_raw_data, path_patched_data, thresh_indices = [0, 0.
...
-                os.makedirs(path_patched_folder)
-            
+            path_patched_folder = path_patched_data / img_folder
+            if not path_patched_folder.exists():
+                path_patched_folder.mkdir(parents=True)
+
             for j, patch in enumerate(patches):
-                imsave(os.path.join(path_patched_folder,'image_%s.png'%j), patch[0],'png')
-                imsave(os.path.join(path_patched_folder,'mask_%s.png'%j), patch[1],'png')
+                imsave(path_patched_folder.joinpath('image_%s.png'%j), patch[0],'png')
```
mathieuboudreau: Why use this syntax instead of something like `path_patched_folder / 'image_%s.png'%j`?

mathieuboudreau: I see, that syntax doesn't work. Then how about:

```python

tmp_image = 'image_' + str(j) + '.png'

imsave(path_patched_folder / tmp_image, patch[0],'png')
```

That way we're consistent with how paths are joined. Same goes for the line below (and maybe other files, I haven't covered them yet)

AxonDeepSeg/data_management/dataset_building.py
```diff
@@ -78,43 +82,42 @@ def patched_to_dataset(path_patched_data, path_dataset, type_, random_seed=None)
...
-
-                        mask = imread(os.path.join(path_patches_folder, data), flatten=True, mode='L')
-                        L_mask.append((mask, int(data_name[-1])))
+                    elif 'mask' in data.name:
+                        mask = imread(path_patches_folder / data, flatten=True, mode='L')
+                        L_mask.append((mask, int(index)))
 
                 # Now we sort the patches to be sure we get them in the right order
-                L_img_sorted, L_mask_sorted = sort_list_files(L_img, L_mask)
+                L_img, L_mask = sort_list_files(L_img, L_mask)
```
mathieuboudreau: What's the logic behind changing the variable names here; before they were unique and now you are overwriting `L_img` and `L_mask`, which are already in use.

Also, you changed it here, but not below, where `L_img_sorted, L_mask_sorted` are still defined in another region of the code.

AxonDeepSeg/data_management/dataset_building.py
```diff
@@ -124,66 +127,70 @@ def patched_to_dataset(path_patched_data, path_dataset, type_, random_seed=None)
...
+                        L_img.append((img, int(index)))
 
-                    elif 'mask' in data:
-                        mask = imread(os.path.join(path_patches_folder, data), flatten=True, mode='L')
-                        L_mask.append((mask, int(data_name[-1])))
+                    elif 'mask' in data.name:
+                        mask = imread(path_patches_folder / data, flatten=True, mode='L')
+                        L_mask.append((mask, int(index)))
                 # Now we sort the patches to be sure we get them in the right order
                 L_img_sorted, L_mask_sorted = sort_list_files(L_img, L_mask)
```
mathieuboudreau: `L_img_sorted, L_mask_sorted` are still used here but not above.

AxonDeepSeg/data_management/dataset_building.py
```diff
@@ -124,66 +127,70 @@ def patched_to_dataset(path_patched_data, path_dataset, type_, random_seed=None)
...
+                    if 'image' in data.name:
+                        img = imread(path_patches_folder / data, flatten=True, mode='L')
+                        L_img.append((img, int(index)))
+
+                    elif 'mask' in data.name:
+                        mask = imread(path_patches_folder / data, flatten=True, mode='L')
+                        L_mask.append((mask, int(index)))
+
                 # Now we sort the patches to be sure we get them in the right order
                 L_img_sorted, L_mask_sorted = sort_list_files(L_img, L_mask)
```
mathieuboudreau: `L_img_sorted, L_mask_sorted` are still used here but not above.

AxonDeepSeg/data_management/dataset_building.py
```diff
@@ -78,43 +82,42 @@ def patched_to_dataset(path_patched_data, path_dataset, type_, random_seed=None)
...
                 # Now we sort the patches to be sure we get them in the right order
-                L_img_sorted, L_mask_sorted = sort_list_files(L_img, L_mask)
+                L_img, L_mask = sort_list_files(L_img, L_mask)
 
                 # Saving the images in the new folder
-                for img,k in L_img_sorted:
-                    imsave(os.path.join(path_dataset,'image_%s.png'%i), img, 'png')
-                    imsave(os.path.join(path_dataset,'mask_%s.png'%i), L_mask_sorted[k][0], 'png')
+                for img,k in L_img:
+                    imsave(path_dataset.joinpath('image_%s.png'%i), img, 'png')
```
mathieuboudreau: Same comment as above, maybe we should keep the standardize way of joining paths to be `path 1 \ path2` rather than using `Path`'s `joinpath` function; my example above could be used as a template.

AxonDeepSeg/data_management/input_data.py
```diff
@@ -268,36 +275,37 @@ def next_batch_WithWeights(self, augmented_data_ = {'type':'None'},
...
     def read_image(self, type_, i):
         '''
         :param i: indice of the image or mask to read.
-        :return image: the loaded image with 8 bit pixels, range of values being [0,288] 
+        :return image: the loaded image with 8 bit pixels, range of values being [0,288]
         '''
 
         # Loading the image using 8-bit pixels (0-255)
-        return imread(os.path.join(self.path,str(type_) + '_%s.png' % i), flatten=False, mode='L')
+        filename = str(type_) + '_%s.png' % i
```
mathieuboudreau: Yep, you got around `joinpath` here similar to how I recommended in the file above.

AxonDeepSeg/segment.py
```diff
@@ -7,398 +7,403 @@
...
+default_SEM_path = MODELS_PATH / SEM_DEFAULT_MODEL_NAME
+default_TEM_path = MODELS_PATH / TEM_DEFAULT_MODEL_NAME
 default_overlap = 25
 
 # Definition of the functions
 
 def segment_image(path_testing_image, path_model,
-			      overlap_value, config, resolution_model,
-				  acquired_resolution = None, verbosity_level=0):
+                  overlap_value, config, resolution_model,
```
mathieuboudreau: Same comment as for "AxonDeepSeg/apply_model.py" re: auto-style formatting, I'll try to parse these two later line-by-line for the differences if the auto-styling can't be easily reversed.

AxonDeepSeg/visualization/get_masks.py
```diff
@@ -17,17 +21,16 @@ def get_masks(path_prediction):
...
-    path_folder, file_name = os.path.split(path_prediction)
-
-    tmp_path = path_prediction.split('_seg-axonmyelin')
-
-    # save the masks
-    if tmp_path[0].endswith('.png'):
-        tmp_path[0] = os.path.splitext(tmp_path[0])[0]
+    # We want to keep the filename path up to the '_seg-axonmyelin' part
+    path = path_prediction.parent
+    broken_filename = path_prediction.name.split('_seg-axonmyelin')[0]
```
mathieuboudreau: I think we should use another word for the prefix in `broken_filename`; I though there was something wrong with it. Maybe `filename_prefix`, `split_filename`, `common_prefix`, etc?

AxonDeepSeg/visualization/get_masks.py
```diff
@@ -17,17 +21,16 @@ def get_masks(path_prediction):
...
+    # Extra check to ensure that the extension was removed
+    if broken_filename.endswith('.png'):
+        broken_filename = broken_filename.split('.png')[0]
 
-    imageio.imwrite(tmp_path[0] + '_seg-axon.png', axon_prediction.astype(int))
-    imageio.imwrite(tmp_path[0] + '_seg-myelin.png', myelin_prediction.astype(int))
+    # Save masks
+    imageio.imwrite(path.joinpath(broken_filename + '_seg-axon.png'), axon_prediction.astype(int))
```
mathieuboudreau: I think this joinpath was missed? Shouldn't we use the `path / filename` syntax?

AxonDeepSeg/visualization/get_masks.py
```diff
@@ -17,17 +21,16 @@ def get_masks(path_prediction):
...
-    # get main path
-    path_folder, file_name = os.path.split(path_prediction)
-
-    tmp_path = path_prediction.split('_seg-axonmyelin')
-
-    # save the masks
-    if tmp_path[0].endswith('.png'):
-        tmp_path[0] = os.path.splitext(tmp_path[0])[0]
+    # We want to keep the filename path up to the '_seg-axonmyelin' part
+    path = path_prediction.parent
```
mathieuboudreau: We should use a more specific variable name than `path` (used to be `path_folder`?). Because `path` is also the name of a module (i.e. try `import path`), which may have been imported, so this would behave strange, and it reads off later on (I though line 32 below was calling the module, not a method of this object).

AxonDeepSeg/visualization/visualize.py
```diff
@@ -97,10 +97,11 @@ def _create_fig_helper(overlayed_img, fig_title):
         ax.imshow(overlayed_img, cmap="hsv", alpha=0.7)
         return fig
 
-    path_img = path + "/image.png"
+    path_img = path / "image.png"
     mask = False
+    listdir = [item.name for item in path]
```
mathieuboudreau: `listdir` is not a good variable name, as there exists `os.listdir`, which is sometimes imported as `from os import listdir`, causing a conflict.

models/config_to_csv.py
```diff
@@ -48,21 +47,20 @@ def describe(write_model):
...
-        
+
     for model_name in L_models:
-        for root in os.listdir(os.path.curdir)[:]:
-            if (os.path.isdir(root)) and (root[-len(model_name):] == model_name):
-                subpath_data = os.path.join(os.path.curdir, root)
-                for data in os.listdir(subpath_data):
-                    if 'config_network' in data:
-                        models = config_decode(models, subpath_data, 'compare')  
+        for path in Path.cwd().iterdir():
```
mathieuboudreau: Same comment as above re: `path` as a filename.

test/test_train_network.py
```diff
@@ -144,7 +142,7 @@ def test_train_model_runs_successfully_for_simplified_case(self):
             "report.txt"
             ]
 
-        existingFiles = os.listdir(self.modelPath)
+        existingFiles = [f.name for f in self.modelPath.iterdir()]
```
mathieuboudreau: Just curious, is there any benefit to rewriting this line the new way? The old way I wrote it (`os.listdir(DIR)`) is easier to read (imo); just wondering if there's a specific reason why we should avoid this funciton?

AxonDeepSeg/segment.py
```diff
@@ -7,398 +7,403 @@
...
+        # Read image
+        img = imageio.imread(str(path_testing_image))
 
-		# Generate tmp file
-		fp = open(os.path.join(path_acquisition, '__tmp_segment__.png'), 'wb+')
+        # Generate tmp file
+        fp = open(path_acquisition / '__tmp_segment__.png', 'wb+')
 
-		img_name_original, file_extension = os.path.splitext(acquisition_name) 
+        img_name_original = acquisition_name.stem
```
mathieuboudreau: Cool!

AxonDeepSeg/segment.py
```diff
@@ -7,398 +7,403 @@
...
+    # Processing the arguments
+    args = vars(ap.parse_args(argv))
+    type_ = str(args["type"])
+    verbosity_level = int(args["verbose"])
+    overlap_value = int(args["overlap"])
+    if args["sizepixel"] is not None:
+        psm = float(args["sizepixel"])
+    else:
+        psm = None
+    path_target_list = [Path(path) for path in args["imgpath"]]
```
mathieuboudreau: `path` here should also be renamed, reasons posted before.

AxonDeepSeg/segment.py
```diff
@@ -7,398 +7,403 @@
...
+    new_path = args["model"]
```
mathieuboudreau: Why did `path_target_list = args["imgpath"]` have to be converted to `path_target_list = [Path(path) for path in args["imgpath"]]`, but `new_path = args["model"]` doesn't need a similar conversion?

maf88:
I did the changes, except for one or two `joinpath()` that I preferred to keep that way for convenience

AxonDeepSeg/segment.py
mathieuboudreau: @maf88 You marked this as resolved without answering my question. If you want to keep it this way that's fine, I was just trying to understand this a bit deeper.

test/test_train_network.py
mathieuboudreau: @maf88 You marked this as resolved without answering my question. If you want to keep it this way that's fine, I was just trying to understand this a bit deeper.

AxonDeepSeg/segment.py
```diff
@@ -298,8 +298,8 @@ def main(argv=None):
         psm = float(args["sizepixel"])
     else:
         psm = None
-    path_target_list = [Path(path) for path in args["imgpath"]]
-    new_path = args["model"]
+    path_target_list = [Path(p) for p in args["imgpath"]]
+    new_path = Path(args["model"])
```
mathieuboudreau: Actually, looks like this line change caused failing tests. Feel free to revert.

mathieuboudreau: Either revert, or maybe it needs the same format as above (e.g. `new_path = [Path(p) for p in args["model"]]`)?

mathieuboudreau:
Ah ha, hmm, we've encountered a hiccup.

I tried running the `guide_dataset_building.ipynb` as I would usually would on a real dataset, and encountered an error:

Capture d’écran 2019-04-04 à 20 26 06

Here is how I defined my paths in the notebook (which is what most users will do, and what we did before):

Capture d’écran 2019-04-04 à 20 35 26


Because you've converted all my tests inputs strings to `Path` objects, e.g.:

https://github.com/neuropoly/axondeepseg/blob/3d7e2ae10c7a8871e77e43f03e01ad586d7f177a/test/data_management/test_dataset_building.py#L16-L27

this essentially assumes that all users will do the same, which is unreasonable to ask. All inputs (in terms of tests filenames, directories) to functions should be plain strings, as that's what most users will do. We can require devs to write backend code using Paths, but I think it's unreasonable to ask the same from users for frontend code; @jcohenadad do you agree?

So, more or less, I believe the proper way to tackle this would be to revert all changes in the test files (test files should only be changed in very **rare** occasions, exactly for this reason), and then refactor the code until these tests pass. That means that inputs of functions that are strings should likely be handled in such a way that if they are not `Path` objects, convert them to that.

maf88:
@mathieuboudreau Actually what I did for the `segment.py:main()` function is to convert the `str` input (received potentially from the command line) to `pathlib`object. We can do something similar and convert any arguments that represent a path to a `pathlib.Path` object at the first line of a function. But I thought about it and I decided instead to change the usage behavior and ask for a `Path` instead of `str` when calling the function. I believe this will be less error-prone than systematically converting an argument to a `Path` object. The only exception will be when parsing arguments received from the command line, since they are all strings :).

maf88:
This is the kind of conversion that we can add, but I think it will quickly become redundant: 
https://github.com/neuropoly/bids_neuropoly/blob/acbf0b91466c51f980623201e872e7dc417677ce/bids_neuropoly/bids.py#L12

@jcohenadad @mathieuboudreau  What do you think?

mathieuboudreau:
>I believe this will be less error-prone than systematically converting an argument to a Path object. 

Unfortunately no, I think most user will input a string to those functions instead of `Path('this/is/my/path')` as  an input, since most other languages/softwares accept pure string for filenames or directory paths into function. The current way will be more "error prone" in the fact that users will encounter errors trying to input strings. 

I think converting them to path (if they aren't already) when first loaded by the function is likely to lead to the more natural user experience that everyone (new or experienced) are used to, and it also have the bonus of keeping all our Jupyter Notebooks & saved scripts working (which they don't in this branch at the moment)

maf88:
> The current way will be more "error prone" in the fact that users will encounter errors trying to input strings.

The error would happen only once, and the solution is to simply do the conversion before the call. This way there are no ripple effects. Whereas, automatically converting to `pathlib.Path` object can eventually create some unwanted behaviors, i.e., the conversion points to a bad folder because of a typo, or the user pass `""` thinking it's equivalent to `None` and the program deletes some file in the current folder (because `Path("")` return the local folder).

>  since most other languages/softwares accept pure string for filenames or directory paths into function.
I totally agree with that, and that's a major argument in favor of doing an automatic conversion.

mathieuboudreau:
I think both cases you mention (passing a path as a typo or `""`) would also cause errors at the moment anyways. Both of those aren't related directly to this PR, but should be solved instead using exception handling, which there's already an issue open for (#29 "Better exception management for the software")

maf88:
I think having DbC approach will be easier rather than a defensive one.

mathieuboudreau:
I think we have a disagreement on the priority for this issue (code quality vs user experience). I'll add this to the agenda for the next meeting (upcoming Wednesday) and we can discuss this further & make a decision then, as the discussion here is getting a bit long.

maf88:
I think to talk about user experience, we need a well defined interface: a CLI, a GUI, or even an API. Code quality is never something to be question, it's the sole purpose of a coder. Here the problem is that knowing that the code is still spare and heterogeneous in term of quality, purpose, and functionalities, what should we do? 1) Try to keep everything as-it-is and slowly trying to improve it, 2) set goals and have hard constraints to reach them. 1) is more time consuming (more reviews, more code to keep up-to-date, etc.) and, eventually prone to confusion. 2) can be more risky (did we break something somewhere that wasn't tested in the UT?), and if the goals are not clear and shared by the team 2) can lead to failure.

Knowing that the numbers of line in the code is quite big, I think we need to consider 2) and define an API and set goals for the progression of ADS. Wild refactoring or random refactoring are not good solution, especially for the long term.

mathieuboudreau:
Those are valid discussion points, but are out of scope of the current PR. A separate issue should be opened for a discussion about such an overall of the API, which should only be introduced in the next major release (3.x). Stability of the current API in `master` should be kept in this PR, which was supposed to only be a minor refactoring (no change in API), which is why I suggested that strings as inputs to all functions for dirs/paths should still be supported here (a 2.x release).

PR #226, opened 2019-06-27
Refactored data augmentation, changed loss function, cleaned notebooks and other improvements

### LINKED ISSUE #19: Try Dice instead of cross entropy in optimization

### LINKED ISSUE #148: Should we keep all Jupyter notebooks?
Currently, there are 18 notebooks:
- some with very cryptic names
- most of them without documentation (and some doc is in French)
- with some redundancy between notebooks
- most of them contain only few lines of code that don't deserve to be in a notebook for someone who knows a minimum of Python programming

Those notebooks are becoming a dump of "nice functionalities to have" without proper harmonization/guideline for creating them. We should control it otherwise we will end up with 300 notebooks in 10 years.

I see two approaches (for our users):

1.  We abandon notebooks and we implement useful functionalities as a wrapper that would be called from the Terminal, without the need to invoke (and install) Jupyter. Those functionalities could include those done by: generate_axons_from_myelin.ipynb, get_masks, etc.  With that approach, we could build pipelines in bash. 

2. We keep Jupyter notebooks, not only as a "tutorials", but rather as "official tools part of ADS that are necessary for doing XYZ tasks". If we go that route, I suggest we (i) include Jupyter as a requirement and (ii) we do some cleaning on the existing notebooks, with a clear vision about what notebooks are needed, for what purpose, etc.

### LINKED ISSUE #240: Repository is too large (350MB)
We should remove some fat and also host images/binaries/models somewhere else (OSF, mirrors, etc.)

### LINKED ISSUE #241: Add README under specific model
Move the information listed in the [wiki](https://github.com/neuropoly/axondeepseg/wiki/Datasets) under model-specific README files

### LINKED ISSUE #278: Remove the "_v1" suffix in models on OSF and on this repos
The versioning is now tracked by OSF, more info [here](https://github.com/neuropoly/axondeepseg/wiki/Template-for-Readme.txt)

vs74:
This major PR aims to handle the improvement in the performance of model as well as an improvised version of data augmentation. 

**DONE**

- Implemented data augmentation ([Albumentation library](https://albumentations.readthedocs.io/en/latest/api/augmentations.html)) similar to previous version of ADS

- Changed from `Cross Entropy` as a loss function to `Dice coefficient` to improve model performance as indicated in issue #19 .

- Interpolation changed from` linear` to `nearest neighbour`

- Notebooks are cleaned and removed irrelevant notebooks as indicated in #148

- Migrated models to OSF storage to prevent bloating of the repository

Fixes #148, Fixes #19, Fixes #241, Fixes #278, Fixes #240


vs74:
For now I have implemented data augmentation similar to one in TensorFlow. Since I am using [albumentation](https://albumentations.readthedocs.io/en/latest/api/augmentations.html), there is no need of script `data_augmentation.py`(data augmentation part is implemented in `train_network.py` and `input_data.py` as it required few lines of code). I want to know how should I approach at implementing` test cases` for` data augmentation` as the `earlier test cases` are useless(since `data_augmentation.py` has to deleted). Any help @jcohenadad / @mathieuboudreau  ?

jcohenadad:
@vs74 since albumentation seems to be [well tested](https://travis-ci.org/albu/albumentations), i would suggest to only add tests that are specific to the use of this library by ADS.

mathieuboudreau:
Agreed with Julien, if you're deleting [AxonDeepSeg/data_management/data_augmentation.py](https://github.com/neuropoly/axondeepseg/blob/50d8af2aefb56a3b3a78900abacdaa70a75ab920/AxonDeepSeg/data_management/data_augmentation.py) in this branch, you can remove the accompanying test file [test/data_management/test_data_augmentation.py](https://github.com/neuropoly/axondeepseg/blob/50d8af2aefb56a3b3a78900abacdaa70a75ab920/test/data_management/test_data_augmentation.py) - no need to rewrite tests unless other test fails after these steps are done.

AxonDeepSeg/train_network.py
```diff
@@ -75,59 +73,88 @@ def train_model(path_trainingset, path_model, config, path_model_init=None,
...
+    shifting = config["da-0-shifting-activate"],
+    rescaling = config["da-1-rescaling-activate"]
+    rotation = config["da-2-random_rotation-activate"]
+    elastic = config["da-3-elastic-activate"]
+    flipping = config["da-4-flipping-activate"]
+    gaussian_blur = config["da-5-gaussian_blur-activate"]
+
+    p_shift = p_rescale = p_rotate = p_elastic = p_flip = p_blur = 0
+    if shifting:
+        p_shift = 0.5
```
mathieuboudreau: Could you write in a comment what this/these values represent, and why you chose them to be 0.5 if the value is on? It begins with p, so I assume it has something to do with probability, but it's not clear to me why 0.5 values are chosen for all of these settings.

vs74: Done in commit 0c6a981.

vs74:
While removing the notebooks as mentioned in PR #148, I narrowed it down from 18 Jupiter notebooks to 7. As mentioned earlier to keep only 4 notebooks that is getting_started.ipynb, guide_dataset_building.ipynb, morphometrics_extraction.ipynb, training_guideline.ipynb. In addition to this I believe there are 3 more notebooks which we should keep ( guide_bireli.ipynb, performance_metrics.ipynb, segmentation_image.ipynb). What do you suggest @jcohenadad / @mathieuboudreau ?

jcohenadad:
> While removing the notebooks as mentioned in PR #148, I narrowed it down from 18 Jupiter notebooks to 7. As mentioned earlier to keep only 4 notebooks that is getting_started.ipynb, guide_dataset_building.ipynb, morphometrics_extraction.ipynb, training_guideline.ipynb. In addition to this I believe there are 3 more notebooks which we should keep ( guide_bireli.ipynb, performance_metrics.ipynb, segmentation_image.ipynb). What do you suggest @jcohenadad / @mathieuboudreau ?

i'm ok with those suggestions

mathieuboudreau:
I've never used either of those three extra ones. They are a bit less stale than the other ones you deleted, but they are still not quite as polished as the 4 we recommended to keep. What benefit do you see in keeping those three?

mathieuboudreau:
I think if we keep those three, it might be warranted to polish them up a bit more (e.g. more explanations of what the notebooks are useful for, deleted commented out code, running them at least once so figures are displayed, etc.)

vs74:
The following reasons I thought for keeping the notebooks
1. `guide_bireli.ipynb` - For this notebooks, I had only just small reason that is this notebook name occurrence is being used in `test_config_tools.py` (`comments` to be specific). However I am against keeping this file. The comments can be deleted from these test cases.  
```
# --------------grid_config tests-------------- #
    @pytest.mark.unit
    def test_grid_config_feature_augmentation_x(self):
        # Sample L_struct and dict_params values below taken from
        # guide_bireli.ipynb
        L_struct = [{
            'structure': [[5, 5, 5], [3, 3, 3], [3, 3, 3], [3, 3, 3]],
            'features_augmentation': 'x2',
            'first_num_features': 16
            }]
        dict_params = {
            'trainingset': ['SEM_3c_512'],
            'trainingset_patchsize': 512,
            'learning_rate_decay_period': 24000
            }

        config_list = grid_config(
            L_struct,
            dict_params,
            base_config=self.config
            )

        for key in list(config_list.keys()):
            assert validate_config(config_list[key])

    @pytest.mark.unit
    def test_grid_config_feature_augmentation_p(self):
        # Sample L_struct and dict_params values below taken from
        # guide_bireli.ipynb
        L_struct = [{
            'structure': [[5, 5, 5], [3, 3, 3], [3, 3, 3], [3, 3, 3]],
            'features_augmentation': 'p2',
            'first_num_features': 16
            }]
        dict_params = {
            'trainingset': ['SEM_3c_512'],
            'trainingset_patchsize': 512,
            'learning_rate_decay_period': 24000
            }

        config_list = grid_config(
            L_struct,
            dict_params,
            base_config=self.config
            )

        for key in list(config_list.keys()):
            assert validate_config(config_list[key])

    @pytest.mark.unit
    def test_grid_config_feature_augmentation_invalid(self):
        # Sample L_struct and dict_params values below taken from
        # guide_bireli.ipynb

        # Note: 'features_augmentation':'d2'-> d is not a valid augmentation
        # flag.
        L_struct = [{
            'structure': [[5, 5, 5], [3, 3, 3], [3, 3, 3], [3, 3, 3]],
            'features_augmentation': 'd2',
            'first_num_features': 16
            }]
        dict_params = {
            'trainingset': ['SEM_3c_512'],
            'trainingset_patchsize': 512,
            'learning_rate_decay_period': 24000
            }

        with pytest.raises(ValueError):
            config_list = grid_config(
                L_struct,
                dict_params,
                base_config=self.config
                )

```

2. `performance_metrics.ipynb` and `segmentation_image.ipynb` -  I believe that keeping these files will help us in future to debug and play easily with the model. Let's say if we want to fine tune model or extend our model it would be relatively easy if we had Jupyter notebook ( so that we can assess  performance of the model and segmentation on the test data / unknown data)  rather than creating one / looking up in the previous commits. Nonetheless as pointed by @mathieuboudreau,  I will perform a validation of the two notebooks.

vs74:
What we delete the information about notebook (`noise_simualtion.ipynb`) mentioned in RTD or should I revert this notebook back and clean it ?
Screen Shot 2019-07-30 at 1 06 52 PM

mathieuboudreau:
I would just edit RTD accordingly for the current state of our notebook collection.

vs74:
@mathieuboudreau could you review this PR ?

vs74:
Done and ready for review.

mathieuboudreau:
Sorry I lost track of this one @vs74 , I will review it ASAP.

AxonDeepSeg/models/default_SEM_model_v1/readme.txt
```diff
@@ -0,0 +1,7 @@
+Current training/validation/testing dataset:
+
+ Patches of 512x512 pixels
+ 3 classes (axon/myelin/background)
+ Rat microscopy
+ white matter spinal cord
+Path to dataset: /Volumes/projects/axondeepseg/labelled_datasets/SEM_3c_512
```
mathieuboudreau: That path is where you mounted it on your computer; you should instead specify that it's on `duke:` instead of your own `/Volumes` folder

AxonDeepSeg/models/default_TEM_model_v1/readme.txt
```diff
@@ -0,0 +1,7 @@
+Current training/validation/testing dataset:
+
+ Patches of 512x512 pixels
+ 3 classes (axon/myelin/background)
+ Mouse microscopy
+ corpus callosum
+Path to dataset: /duke/projects/axondeepseg/labelled_datasets/TEM_3c_512
```
mathieuboudreau: See, like here.

AxonDeepSeg/train_network.py
```diff
@@ -75,63 +73,104 @@ def train_model(path_trainingset, path_model, config, path_model_init=None,
...
 
-    ###########################Validation###########
-    data_gen_args = dict()
-    image_datagen = ImageDataGenerator(**data_gen_args)
-    mask_datagen = ImageDataGenerator(**data_gen_args)
+    #Randomly rotates an image between low limit and high limit.
+    ShiftScaleRotate(shift_limit=(low_limit, high_limit) , scale_limit=(0,0),rotate_limit=(0,0), border_mode=border_mode, p = p_shift, interpolation = cv2.INTER_NEAREST),
 
+    #Randomly applies elastic transformation on the image.
+    ElasticTransform(alpha= alpha, sigma= sigma, p = p_elastic, alpha_affine = alpha),
```
mathieuboudreau: Can the ElasticTransform not have nearest neighbour interpolation for the masks with Augmentations? If not, we should maybe explore omitting this augmentation, since the masks will have more than 3 classes. Otherwise, we might need to rebin the mask to the three classes after this augmentation is performed if possible.

AxonDeepSeg/train_network.py
```diff
@@ -75,63 +73,104 @@ def train_model(path_trainingset, path_model, config, path_model_init=None,
...
 
-    image_generator_valid = image_datagen.flow(x=image_valid_gen, y=None,  batch_size=batch_size, seed = seed, shuffle=False)
-    mask_generator_valid = mask_datagen.flow(x=mask_valid_gen, y=None,  batch_size=batch_size, seed = seed, shuffle=False)
+    #Blurs an image using gaussian kernal.
+    GaussianBlur(p = p_blur),
```
mathieuboudreau: Same question as for the ElasticTransform

mathieuboudreau:
@vs74 this branch is missing information in the CHANGELOG.md file about the changes that were made (e.g. refactored the data augmentation to use Augmentations, cleaned notebooks, updated models, etc).

AxonDeepSeg/train_network.py
```diff
@@ -75,63 +73,104 @@ def train_model(path_trainingset, path_model, config, path_model_init=None,
...
-                         # height_shift_range = shifting(patch_size, n_classes) [0],
-                         # fill_mode = "constant",
-                         # cval = 0
-                         )
+    shifting = config["da-0-shifting-activate"],
+    rescaling = config["da-1-rescaling-activate"]
+    rotation = config["da-2-random_rotation-activate"]
+    elastic = config["da-3-elastic-activate"]
+    flipping = config["da-4-flipping-activate"]
+    gaussian_blur = config["da-5-gaussian_blur-activate"]
```
mathieuboudreau: Here, you've implicitly changed the config key from 

`da-5-noise_addition-activate`

to

`da-5-gaussian_blur-activate`

Meaning that all our (and my) old configs will break. A better way to handle this refactor would be to retain the old compatibility as a backup, e.g. something like

```python

if "da-5-gaussian_blur-activate" in config:
    gaussian_blur = config["da-5-gaussian_blur-activate"]
elif "da-5-noise_addition-activate" in config:
    # Preserved for retrocompatibility with old configs
    gaussian_blur = config["da-5-noise_addition-activate"]
end
```

I discovered this because the training notebook failed (so that's another change that's needed, both da-5-gaussian_blur-activate is needed as a key, but also da-6-reflection_border-activate is missing.

AxonDeepSeg/train_network.py
vs74: Elastic transform and GaussianBlur don't have nearest neighbour interpolation for masks. Could you elaborate on `omitting the augmentation` as it's unclear to me ?

AxonDeepSeg/models/default_SEM_model_v1/readme.txt
vs74: Resolved in commit d0e33c5.

AxonDeepSeg/train_network.py
vs74: Fixed in commit(#1041d32)

vs74:
> @vs74 were you required to generate new updated model files in this branch? AxonDeepSeg/models/default_SEM_model_v1/model.ckpt.data-00000-of-00001 is a hefty 22 MB, and we already have one copy of it in the repo.
> 
> And oddly, the TEM version of that file is 0 bytes - was it trained correctly? AxonDeepSeg/models/default_TEM_model_v1/model.ckpt.data-00000-of-00001

Since in the last meeting we diced to shift to` dice loss `as loss function while training our network ,models have to retrained. As TEM model is concerned, I will cross verify and will update in this issue.

vs74:
> @vs74 this branch is missing information in the CHANGELOG.md file about the changes that were made (e.g. refactored the data augmentation to use Augmentations, cleaned notebooks, updated models, etc).

Resolved in commit (aa49404)

mathieuboudreau:
To resolve merging with master (which had conflicting trained model files from changes there and changes here), and in anticipation of #261, I resolved the conflicts during the merge by deleting the model files in commit 8c3a251. The tests will mostly fail now, until #261 is complete and merged into this branch.

AxonDeepSeg/train_network.py
mathieuboudreau: This needs to be fixed in the training guidelines notebook too, as it calls `da-6-reflection_border-activate`

mathieuboudreau: Sorry, it's because it's missing there

mathieuboudreau:
@vs74 I'm trying to train a dataset using this branch, however I'm getting an error from the mask shape, see here:

<details>

```
WARNING:tensorflow:From /home/GRAMES.POLYMTL.CA/u097381/miniconda/envs/model_train/lib/python3.6/site-packages/tensorflow/python/framework/op_def_library.py:263: colocate_with (from tensorflow.python.framework.ops) is deprecated and will be removed in a future version.
Instructions for updating:
Colocations handled automatically by placer.
WARNING:tensorflow:From /home/GRAMES.POLYMTL.CA/u097381/miniconda/envs/model_train/lib/python3.6/site-packages/keras/backend/tensorflow_backend.py:3445: calling dropout (from tensorflow.python.ops.nn_ops) with keep_prob is deprecated and will be removed in a future version.
Instructions for updating:
Please use `rate` instead of `keep_prob`. Rate should be set to `rate = 1 - keep_prob`.
WARNING:tensorflow:From /home/GRAMES.POLYMTL.CA/u097381/miniconda/envs/model_train/lib/python3.6/site-packages/tensorflow/python/ops/math_ops.py:3066: to_int32 (from tensorflow.python.ops.math_ops) is deprecated and will be removed in a future version.
Instructions for updating:
Use tf.cast instead.
WARNING:tensorflow:From /home/GRAMES.POLYMTL.CA/u097381/miniconda/envs/model_train/lib/python3.6/site-packages/tensorflow/python/ops/math_grad.py:102: div (from tensorflow.python.ops.math_ops) is deprecated and will be removed in a future version.
Instructions for updating:
Deprecated in favor of operator or tf.math.divide.
Epoch 1/2000
 9/63 [===>..........................] - ETA: 19:19 - loss: 0.5188 - acc: 0.5108 - dice_axon: 0.3940 - dice_myelin: 0.4783
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-8-d7876a16e1e7> in <module>
      2 tf.reset_default_graph()
      3 
----> 4 train_model(str(path_training), str(path_model), config)
      5 # Note: For multi-OS compatibility of this notebook, paths were defined as Path objects from the pathlib module.
      6 # They need to be converted into strings prior to be given as arguments to train_model(), as some old-style string

~/model_training/axondeepseg/AxonDeepSeg/train_network.py in train_model(path_trainingset, path_model, config, path_model_init, save_trainable, gpu, debug_mode, gpu_per)
    217     model.fit_generator(train_generator, validation_data=(valid_generator), steps_per_epoch=train_steps,
    218                         validation_steps=valid_steps,
--> 219                         epochs=epochs, callbacks=[tensorboard, checkpoint_loss, checkpoint_acc])
    220 
    221     ########################## Save the model after Training ###########

~/miniconda/envs/model_train/lib/python3.6/site-packages/keras/legacy/interfaces.py in wrapper(*args, **kwargs)
     89                 warnings.warn('Update your `' + object_name + '` call to the ' +
     90                               'Keras 2 API: ' + signature, stacklevel=2)
---> 91             return func(*args, **kwargs)
     92         wrapper._original_function = func
     93         return wrapper

~/miniconda/envs/model_train/lib/python3.6/site-packages/keras/engine/training.py in fit_generator(self, generator, steps_per_epoch, epochs, verbose, callbacks, validation_data, validation_steps, class_weight, max_queue_size, workers, use_multiprocessing, shuffle, initial_epoch)
   1416             use_multiprocessing=use_multiprocessing,
   1417             shuffle=shuffle,
-> 1418             initial_epoch=initial_epoch)
   1419 
   1420     @interfaces.legacy_generator_methods_support

~/miniconda/envs/model_train/lib/python3.6/site-packages/keras/engine/training_generator.py in fit_generator(model, generator, steps_per_epoch, epochs, verbose, callbacks, validation_data, validation_steps, class_weight, max_queue_size, workers, use_multiprocessing, shuffle, initial_epoch)
    179             batch_index = 0
    180             while steps_done < steps_per_epoch:
--> 181                 generator_output = next(output_generator)
    182 
    183                 if not hasattr(generator_output, '__len__'):

~/miniconda/envs/model_train/lib/python3.6/site-packages/keras/utils/data_utils.py in get(self)
    599         except Exception as e:
    600             self.stop()
--> 601             six.reraise(*sys.exc_info())
    602 
    603 

~/miniconda/envs/model_train/lib/python3.6/site-packages/six.py in reraise(tp, value, tb)
    691             if value.__traceback__ is not tb:
    692                 raise value.with_traceback(tb)
--> 693             raise value
    694         finally:
    695             value = None

~/miniconda/envs/model_train/lib/python3.6/site-packages/keras/utils/data_utils.py in get(self)
    593         try:
    594             while self.is_running():
--> 595                 inputs = self.queue.get(block=True).get()
    596                 self.queue.task_done()
    597                 if inputs is not None:

~/miniconda/envs/model_train/lib/python3.6/multiprocessing/pool.py in get(self, timeout)
    642             return self._value
    643         else:
--> 644             raise self._value
    645 
    646     def _set(self, i, obj):

~/miniconda/envs/model_train/lib/python3.6/multiprocessing/pool.py in worker(inqueue, outqueue, initializer, initargs, maxtasks, wrap_exception)
    117         job, i, func, args, kwds = task
    118         try:
--> 119             result = (True, func(*args, **kwds))
    120         except Exception as e:
    121             if wrap_exception and func is not _helper_reraises_exception:

~/miniconda/envs/model_train/lib/python3.6/site-packages/keras/utils/data_utils.py in get_index(uid, i)
    399         The value at index `i`.
    400     """
--> 401     return _SHARED_SEQUENCES[uid][i]
    402 
    403 

~/model_training/axondeepseg/AxonDeepSeg/data_management/input_data.py in __getitem__(self, index)
     67 
     68         images = np.array(image)
---> 69         masks  = np.array(mask)
     70 
     71         image_aug = []

ValueError: could not broadcast input array from shape (512,512,3) into shape (512)
```

</details>

<p>

Any insight on a quick fix?

mathieuboudreau:
I've uploaded the two notebooks I've used here : https://drive.google.com/drive/folders/1YTk9SpwrDySyErlQ1bm9ZWr3jJFj_qVT?usp=sharing

Run the `guide_dataset_building_TEM` one first, then the training one. The data is hosted on OSF, and should be downloaded and split automatically (that functionality is still a WIP, so it's writing within the script. Just run all the cells).

vs74:
> @vs74 I'm trying to train a dataset using this branch, however I'm getting an error from the mask shape, see here:
> 
> Any insight on a quick fix?

@mathieuboudreau, perhaps the error might be caused by the introduction of PR #246 into this branch. Can you try removing this PR from this branch and then retrain the model again? Though, to be fully certain of the exact cause of the error, I need some time and access to Rosenberg as mine has already expired ( until then can you try that option and let me know.

mathieuboudreau:
@vs74 That PR was an important one, as it overhauled all of the image reading/writing, so I think removing it wouldn't be quite simple in this branch and might break other things. As for debugging; my error happened happened within the first epoch, so this might be something you can test out on your local computer (it shouldn't take long to crash, no longer than an hour or so on your slower machine if my estimate is correct).

vs74:
@mathieuboudreau, I tried reproduced the error encountered by you during training, however, during running the first script ( that is `guide_dataset_building_TEM`),  I faced an error. Here in the error.
Screenshot 2019-10-22 at 5 56 33 PM

**NOTE : - The error was seen while running the second last cell of the notebook, all the other cells worked smoothly.**

mathieuboudreau:
@vs74 Did you download the TEM data from OSF and run the `guide_dataset_building.ipynb` notebook to generate the patches?

vs74:
> @vs74 Did you download the TEM data from OSF and run the `guide_dataset_building.ipynb` notebook to generate the patches?

I just simply run all the cells in `guide_dataset_building.ipynb` and in the process TEM data was downloaded from OSF.

mathieuboudreau:
Ok, that should have worked bzcause it's whqt I did too. Let me try again. Have you looked to see which folder the patched data is in after running that script, and why ot may not be where the training was specified?

Télécharger Outlook pour Android<https://aka.ms/ghei36>

mathieuboudreau:
@vs74 commit 410ddda should have resolved the issue you had in running my scripts.

mathieuboudreau:
Re: original issue https://github.com/neuropoly/axondeepseg/pull/226#issuecomment-539763835, I'm starting to suspect that it may be specific to a certain mask file. Now sure why, but I'm noticing that the error is not happening at the same batch # (which are randomly generated) every time I run it. To be continued..

mathieuboudreau:
@vs74 I think I'm getting close to the source of the issue. My gut feeling is telling me that the problem is that in some instances, the batch of masks (e.g. 8 patches) contains one that does not have the same number of channels/mask values than all the others; for example, 7 of them might be images 512x512x3 (3 being the number of classes) and then one of them would be 512x512x4 (or 2, or whatever). So in that instance, when this line is called,

https://github.com/neuropoly/axondeepseg/blob/b670f5f252f27ca07aad19e893f9ee7fbffe7d75/AxonDeepSeg/data_management/input_data.py#L68

an error is thrown, because np.array can't convert a list of arrays if their dimensions don't all match.

I suspect that (part) of the cause of the culprit is this function:

https://github.com/neuropoly/axondeepseg/blob/b670f5f252f27ca07aad19e893f9ee7fbffe7d75/AxonDeepSeg/data_management/input_data.py#L68

Which 1- **[doesn't have unit tests written for it](https://github.com/neuropoly/axondeepseg/tree/master/test/data_management)** so I can't be sure this function works well for all use cases (coverage still passes because other functions use it, but that may not be sufficient), and 2- uses odd threshold indices; I expected [0, 128, 255], but you're using [0, 0.2, 0.8]:

https://github.com/neuropoly/axondeepseg/blob/b670f5f252f27ca07aad19e893f9ee7fbffe7d75/AxonDeepSeg/data_management/input_data.py#L13

So overall, I think this may not be (directly) related to the single channel PR change you did in #246, but may be moreso a bug that was discovered using this large dataset, which may contain a usecase that's not handled well.

mathieuboudreau:
Actually by looking at the history, the [0, 0.2, 0.8] thresholds appear to be the way we've always done it (some lines are two years old), so I think we can eliminate that from a potential source.

vs74:
> @vs74 I think I'm getting close to the source of the issue. My gut feeling is telling me that the problem is that in some instances, the batch of masks (e.g. 8 patches) contains one that does not have the same number of channels/mask values than all the others; for example, 7 of them might be images 512x512x3 (3 being the number of classes) and then one of them would be 512x512x4 (or 2, or whatever). So in that instance, when this line is called,
> 
> https://github.com/neuropoly/axondeepseg/blob/b670f5f252f27ca07aad19e893f9ee7fbffe7d75/AxonDeepSeg/data_management/input_data.py#L68
> 
> an error is thrown, because np.array can't convert a list of arrays if their dimensions don't all match.
> 
> I suspect that (part) of the cause of the culprit is this function:
> 
> https://github.com/neuropoly/axondeepseg/blob/b670f5f252f27ca07aad19e893f9ee7fbffe7d75/AxonDeepSeg/data_management/input_data.py#L68
> 
> Which 1- **[doesn't have unit tests written for it](https://github.com/neuropoly/axondeepseg/tree/master/test/data_management)** so I can't be sure this function works well for all use cases (coverage still passes because other functions use it, but that may not be sufficient), and 2- uses odd threshold indices; I expected [0, 128, 255], but you're using [0, 0.2, 0.8]:
> 
> https://github.com/neuropoly/axondeepseg/blob/b670f5f252f27ca07aad19e893f9ee7fbffe7d75/AxonDeepSeg/data_management/input_data.py#L13
> 
> So overall, I think this may not be (directly) related to the single channel PR change you did in #246, but may be moreso a bug that was discovered using this large dataset, which may contain a usecase that's not handled well.

@mathieuboudreau, The only role of` mask = np.array(mask)` is to just convert array object into NumPy array object, so the main cause might be attributed to thresholding. One question, did you check manually if any one of mask say for instance had a shape (512,512,4) or any other dimension except the last one being 3 ( The only reason I am asking is  that I don't think it would any channel except from 3, actually the thresholding [0, 0.2, 0.8] is ought to divide the single channel mask into 3 channel mask)?

vs74:
> @vs74 I think I'm getting close to the source of the issue. My gut feeling is telling me that the problem is that in some instances, the batch of masks (e.g. 8 patches) contains one that does not have the same number of channels/mask values than all the others; for example, 7 of them might be images 512x512x3 (3 being the number of classes) and then one of them would be 512x512x4 (or 2, or whatever). So in that instance, when this line is called,
> 
> https://github.com/neuropoly/axondeepseg/blob/b670f5f252f27ca07aad19e893f9ee7fbffe7d75/AxonDeepSeg/data_management/input_data.py#L68
> 
> an error is thrown, because np.array can't convert a list of arrays if their dimensions don't all match.
> 
> I suspect that (part) of the cause of the culprit is this function:
> 
> https://github.com/neuropoly/axondeepseg/blob/b670f5f252f27ca07aad19e893f9ee7fbffe7d75/AxonDeepSeg/data_management/input_data.py#L68
> 
> Which 1- **[doesn't have unit tests written for it](https://github.com/neuropoly/axondeepseg/tree/master/test/data_management)** so I can't be sure this function works well for all use cases (coverage still passes because other functions use it, but that may not be sufficient), and 2- uses odd threshold indices; I expected [0, 128, 255], but you're using [0, 0.2, 0.8]:
> 
> https://github.com/neuropoly/axondeepseg/blob/b670f5f252f27ca07aad19e893f9ee7fbffe7d75/AxonDeepSeg/data_management/input_data.py#L13
> 
> So overall, I think this may not be (directly) related to the single channel PR change you did in #246, but may be moreso a bug that was discovered using this large dataset, which may contain a usecase that's not handled well.

For your question related to thresholding using indices [0, 0.2, 0.8] not [0, 128, 255] - actually there is no need of that because in the code of thresholding we are multiplying the threshold  indice values by 255. For more information you can see the source code [here](https://github.com/neuropoly/axondeepseg/blob/b670f5f252f27ca07aad19e893f9ee7fbffe7d75/AxonDeepSeg/data_management/input_data.py#L87)

mathieuboudreau:
@vs74 So looks like you're right, that's not directly the issue. All images in the patch were the same size:

Capture d’écran 2019-10-29 à 11 38 43
Capture d’écran 2019-10-29 à 11 38 54

I'm really perplexed as to why this error would be thrown then, as the variable `masks` is undeclared beforehand in that function, 

https://github.com/neuropoly/axondeepseg/blob/410ddda295d7b32b051dd411f0caf9f6fe10082d/AxonDeepSeg/data_management/input_data.py#L53-L69

...unless it's surviving out of scope somehow from another function. Maybe initialising it to be empty `masks=[]` at the beginning will resolve the issue.

vs74:
> @vs74 So looks like you're right, that's not directly the issue. All images in the patch were the same size:
> 
> Capture d’écran 2019-10-29 à 11 38 43
> 
> Capture d’écran 2019-10-29 à 11 38 54
> 
> I'm really perplexed as to why this error would be thrown then, as the variable `masks` is undeclared beforehand in that function,
> 
> https://github.com/neuropoly/axondeepseg/blob/410ddda295d7b32b051dd411f0caf9f6fe10082d/AxonDeepSeg/data_management/input_data.py#L53-L69
> 
> ...unless it's surviving out of scope somehow from another function. Maybe initialising it to be empty `masks=[]` at the beginning will resolve the issue.

@mathieuboudreau, I don't feel initialising masks = [] will resolve the error. I will personally try looking into this error by following your latest commit in this branch.

mathieuboudreau:
Ha, cool, the branch does pass now. The training issue is still here, but that's progress nonetheless.

vs74:
Ahh... I think figured out the problem. Actually, I also mentioned this in one of the meetings, that is while doing patching some the images were having different dimension than expected. For example one of mask in Grouped_patches/Train is having dimension 511 X 512 but the dimension should have been 512 X 512. Here is the mask image (mask_273 ) having this dimension: - 
mask_237

The problem lies in the in patching script that is why training script was not able to handle this use case.

vs74:
There are 2 solutions to this problem 
1. First to fix the source of the problem that is patching script
2. To handle this case in training script wherein the missing pixel value can be fixed
What do you think @jcohenadad, @mathieuboudreau  ?

jcohenadad:
> There are 2 solutions to this problem
> 
> 1. First to fix the source of the problem that is patching script
> 2. To handle this use case in training script wherein the missing pixel value can be fixed
>    What do you think @jcohenadad, @mathieuboudreau  ?

generally speaking i am always more enthusiastic when it comes to fixing the source of problems ;-)

mathieuboudreau:
Huh! Interesting. We should definitely fix it at the source, and figure out why this was happening. What we could do is write an assertion check when patching to ensure that all images have the expected dimension, and see when it crashes & debug that one. I'm wondering if this is a data corruption issue or a code issue.

mathieuboudreau:
Ok so I looked into it, and found that it was only one image that was causing the issue.

`20160718_nyu_mouse_26_0004`

For this image, two patches were 512x511, and only from masks (the matching image patches did not have that mismatch).

The root cause of this is because the raw image and raw mask have a dimension mismatch (image: 3762 × 2286, mask: 3760 × 2284), which should never be the case and is not something we should handle (i.e. images and masks must always have the same dimensions!!). The issue was originating from the fact that only the size of the image is used in planning the patch position for both the image and the mask (in [this function](https://github.com/neuropoly/axondeepseg/blob/master/AxonDeepSeg/data_management/patch_extraction.py)).

The solution to this is to either remove this image from the dataset, or create a new mask such that it matches the image size. The fix for issue #273 is to assert the size of images match the size of masks, and if they don't, throw an exception and inform the user that they should either throw this data away or fix it.

Thanks for finding the 512x511 issue @vs74, that put us on the right track!

mathieuboudreau:
I removed the dataset `20160718_nyu_mouse_26_0004` from the OSF zipped file, and training now works without any issues.

@vs74 is there anything else that needs to be done in this branch before we can start reviewing again?

vs74:
@mathieuboudreau, one last thing needs to be done that is in OSF, till now I haven't updated model's file with the commit number, which establishes compatibility with the repo as discussed in one of our meetings. Apart from it, everything seems ok to me. Maybe you can start reviewing and in the meanwhile I will update this above mentioned change in OSF storage

mathieuboudreau:
I am currently retraining the model for TEM, and if that works well, will do SEM after. So maybe hold off for the OSF commit change until the new models are up there.

vs74:
Since the introduction of latest commit in the master branch, this PR has failed. @mathieuboudreau any idea what might be the probable cause ?

vs74:
Earlier images in` input_data.py` script was being read by `cv2`, but since now images are now being read by `ads_utils.imread()`. For training the network the albumentations library was using the `cv2`. Introduction of recent commit removed the import of `cv2` from the train_network.py script. Hence, the necessary import was required. I hope this will solve travis failure.

mathieuboudreau:
Good catch @vs74 ! Yes I had removed the cv2 import because of the recent image i/o changes, but didn't know that you still used CV2 in some places for other functionalities.

vs74:
@mathieuboudreau, is the SEM model trained ? Just want to know, as the OSF update is pending for this PR to be complete.

mathieuboudreau:
Not yet - sorry about the delay. I'll get that done and the full TEM one ASAP.

vs74:
The latest commit had caused the PR build to fail. Any clue ?

mathieuboudreau:
@vs74 looks like it's a timeout on the Travis side, it just needs a reset. For some reason though, I can't access that option at the moment, I'm looking into it.

mathieuboudreau:
@vs74 I've uploaded the trained models. They were trained on commit b05a40aa03979a83313fe8704ee389672ed26ed7 . Please change it wherever needed, and we'll review the branch.

vs74:
Thanks @mathieuboudreau, I will make the necessary changes.

vs74:
As discussed, to keep our OSF model with the ADS, README.txt should be included under the respective model ( SEM and TEM ).

Here, I have included a rough draft of README.txt which details following things : - 
1. `ADS Git Hashtag` section to specify the **Git commit** on which new models files were trained.
2. `Saving (OSF) ` which includes the **OSF zip version number**  and t**he directory structure** under which the models are saved in OSF storage. 
3. "Steps to train SEM/TEM model" which describes the reproducibility of the model files (SEM or TEM ).
4. A `NOTE` section explaining the necessary changes which are reuquired to make in OSF storage, Readme.txt, ADS software for contributors / collaborators, just in case they want to upload a improved / new model. 

Here is the screenshot of `README.txt` describing above details : - 

Screenshot 2019-12-15 at 7 13 02 AM

What do you think @jcohenadad  and @mathieuboudreau ?

jcohenadad:
@vs74 i have a few modifications to suggest, which is difficult to do on this screenshot. How about adding to this GH repository a txt template that people would use whenever they create a new model, and we would instruct how to use this template in [this wiki](https://github.com/neuropoly/axondeepseg/wiki/How-to-test-(validate)-new-model).

vs74:
Sorry for the delay. That seems perfect to me. So just one question, is the txt template you are mentioning about would be in the WIKI pages or in the OSF storage ? Also, I have created a Wiki page instructing on how to update model files on the repo and OSF storage, feel free to edit by referring [here](https://github.com/neuropoly/axondeepseg/wiki/Instructions-on-How-to-Update-Models-in-Repo-and-OSF-Storage)

vs74:
I have updated the instruction on WIKI for the template [here](https://github.com/neuropoly/axondeepseg/wiki/Template-for-Readme.txt) and how to update model files, visit [here](https://github.com/neuropoly/axondeepseg/wiki/Instructions-on-How-to-Update-Models-in-Repo-and-OSF-Storage). Kindly feel free to suggest any changes. Once the changes for the template are finalised, I will upload the template along with model files in the OSF storage, thereafter this PR will be ready for review.

mathieuboudreau:
> I have updated the instruction on WIKI for the template [here](https://github.com/neuropoly/axondeepseg/wiki/Template-for-Readme.txt) and how to update model files, visit [here](https://github.com/neuropoly/axondeepseg/wiki/Instructions-on-How-to-Update-Models-in-Repo-and-OSF-Storage). Kindly feel free to suggest any changes. Once the changes for the template are finalised, I will upload the template along with model files in the OSF storage, thereafter this PR will be ready for review.

@vs74 is this PR ready for review?

vs74:
@mathieuboudreau, One last thing is left that is **template** for the readme.txt needs to be decided upon as I am having some reservations. Once that done, this branch is ready for review.

vs74:
Also, can you make the necessary changes in your [PR](https://github.com/neuropoly/axondeepseg/pull/276) so that we can merge that one before reviewing and merging this branch.

setup.py
```diff
@@ -47,7 +61,11 @@
     include_package_data=True,
     entry_points={
         'console_scripts': [
-            'axondeepseg = AxonDeepSeg.segment:main','axondeepseg_test = AxonDeepSeg.integrity_test:integrity_test'
+           'axondeepseg_models = AxonDeepSeg.models.download_model:main', 'axondeepseg = AxonDeepSeg.segment:main','axondeepseg_test = AxonDeepSeg.integrity_test:integrity_test'
```
jcohenadad: maybe change `axondeepseg_models` for `download_models`

vs74: done in commit (b4a3276)

vs74:
Finally this PR is ready for review. But before merging we should first merge https://github.com/neuropoly/axondeepseg/pull/275 as in oder to solve #278, the necessary changes from @Stoyan-I-A PR needs to be merged first.

AxonDeepSeg/models/download_model.py
```diff
@@ -0,0 +1,17 @@
+from AxonDeepSeg.ads_utils import download_data
+import shutil
+
+def main():
+    url_TEM_model = "https://osf.io/2hcfv/?action=download&version=4"  # URL of TEM model hosted on OSF storage with the appropriate versioning on OSF
+    url_SEM_model = "https://osf.io/sv7u2/?action=download&version=4"  # URL of SEM model hosted on OSF storage with the appropriate versioning on OSF
+
+    if (not download_data(url_TEM_model) and not download_data(url_SEM_model)) ==1:
```
jcohenadad: add space after "==" according to PEP8

AxonDeepSeg/train_network.py
```diff
@@ -75,63 +74,108 @@ def train_model(path_trainingset, path_model, config, path_model_init=None,
...
-    image_generator_valid = image_datagen.flow(x=image_valid_gen, y=None,  batch_size=batch_size, seed = seed, shuffle=False)
-    mask_generator_valid = mask_datagen.flow(x=mask_valid_gen, y=None,  batch_size=batch_size, seed = seed, shuffle=False)
+    #Blurs an image using gaussian kernal.
+    GaussianBlur(p = p_blur),
+
+    #Randomly rotates the image between low bound and high bound.
+    Rotate(limit=(low_bound, high_bound), border_mode=border_mode, p = p_rotate, interpolation = cv2.INTER_NEAREST)
+
+    ])
+
```
jcohenadad: any reason to keep all those empty lines?

docs/source/documentation.rst
```diff
@@ -449,4 +446,4 @@ SOFTWARE.
 Contributors
 ===============================================================================
 
-Pierre-Louis Antonsanti, Mathieu Boudreau, Oumayma Bounou, Julien Cohen-Adad, Victor Herman, Melanie Lubrano, Christian Perone, Maxime Wabartha, Aldo Zaimi.
+Pierre-Louis Antonsanti, Mathieu Boudreau, Oumayma Bounou, Julien Cohen-Adad, Victor Herman, Melanie Lubrano, Christian Perone, Maxime Wabartha, Aldo Zaimi, Vasudev Sharma.
```
jcohenadad: add Stoyan and Marie-Hélène

vs74: Done in commit(174f026)

AxonDeepSeg/train_network.py
vs74: Just because  there are many data augmentation operations done on the image, and to make it easy to understand, I have made some comments on the top of each operation while keeping some empty lines.

AxonDeepSeg/models/download_model.py
vs74: Resolved in commit (096440d)

AxonDeepSeg/train_network.py
mathieuboudreau: This is a point that needs to be revisited, as it wasn't resolved. By not having nearest neighbour intperpolation, elastic and gaussianblur transformations will cause data to be trained on masks with more than three classes if used .

mathieuboudreau:
> Finally this PR is ready for review. But before merging we should first merge #275 as in oder to solve #278, the necessary changes from @Stoyan-I-A PR needs to be merged first.

@vs74 #275 was merged, so now we should be able to solve #278 for this branch.

vs74:
> > Finally this PR is ready for review. But before merging we should first merge #275 as in oder to solve #278, the necessary changes from @Stoyan-I-A PR needs to be merged first.
> 
> @vs74 #275 was merged, so now we should be able to solve #278 for this branch.

Thanks @mathieuboudreau for reviewing an merging that branch, now my PR is ready to be merged.

notebooks/performance_metrics.ipynb
```diff
@@ -49,21 +49,21 @@
...
    "outputs": [],
    "source": [
-    "path_img = '~/tmp/axondeepseg/AxonDeepSeg/data_test/image.png'"
+    "path_img = '../AxonDeepSeg/data_test/image.png'"
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 33,
```
mariehbourget: _v1 is still in model_name on line 81 below

CHANGELOG.md
```diff
@@ -5,7 +5,10 @@ Version [2.2dev] - XXXX-XX-XX
 -------------------------------------------------------------------------------
 
 **Changed:**
-
+- Refractored data augmentation to use `Albumentation` library.
+- Cleaned jupyter notebooks and reduced from 13 notebooks to 4.
```
mariehbourget: should be from 13 notebooks to 5 (instead of 4)

mariehbourget:
I’ve reviewed the Jupyter notebooks of the PR.

In getting_started.ipynb, section 1.2: _v1 is still in the model_name. My understanding is that with this PR, the model will be downloaded from OSF instead of being in the models folder by default. If that is the case, shouldn’t we add the download step in the getting started notebook?

In morphometrics_extraction.ipynb, section 7: I would add a comment explaining what this cell does.

CHANGELOG.md
vs74: Yes you are right, earlier we decided to keep 4 notebooks that's why the number is 4. However, we decided to keep 5, but thanks for pointing out, I have updated in commit (
992804f)

mathieuboudreau:
> Thanks @mathieuboudreau for reviewing and merging that branch, now my PR is ready to be merged.

One unresolved comment is this one: https://github.com/neuropoly/axondeepseg/pull/226#discussion_r386137758

We need to decide on how to tackle this issue, because I'm concerned that using the Gaussian/Elastic data augmentation steps might negatively impact the training if there is no option for nearest neighbour interpolation. That means that after using these data augmentations, the data will be trained on masks with smooth edges (all numbers between 0 and 255), instead of three classes (myelin=127, axon=255, background=0).

jcohenadad:
> We need to decide on how to tackle this issue, because I'm concerned that using the Gaussian/Elastic data augmentation steps might negatively impact the training if there is no option for nearest neighbour interpolation. That means that after using these data augmentations, the data will be trained on masks with smooth edges (all numbers between 0 and 255), instead of three classes (myelin=127, axon=255, background=0).

how about we raise a `NotImplementedError` if users want to use it, and we open a specific issue for it?

vs74:
> I’ve reviewed the Jupyter notebooks of the PR.
> 
> In getting_started.ipynb, section 1.2: _v1 is still in the model_name. My understanding is that with this PR, the model will be downloaded from OSF instead of being in the models folder by default. If that is the case, shouldn’t we add the download step in the getting started notebook?
> 
> In morphometrics_extraction.ipynb, section 7: I would add a comment explaining what this cell does.

Thanks @mariehbourget  for pointing out the '_v1' under models name in the Jupyter notebooks. I have updated them in the latest commit. To your first question, the models will be downloaded from OSF and will be saved under the models directory. But for correct installation of models  a user must run `pip install -e .` after cloning the repository, so there is no need to add download step in `getting_started.ipynb`.

However, I think that some instructions on `pip install -e .` in `getting_started.ipynb` needs to added, just in case a user doesn't run the command,  I will do that.

mathieuboudreau:
Sure - we'll also have to turn those options off in the training guidelines notebooks, and possibly some unit tests. We'll also need to retrain the model, since the previous model was trained with elastic data augmentation on, so likely contains this issue (which although likely minor, will still impact performance).

mathieuboudreau:
>However, I think that some instructions on pip install -e . in getting_started.ipynb needs to added, just in case a user doesn't run the command, I will do that.

No need to tell the user to do `pip install -e .` in that notebook, they should have already done that. All I would add is a link to our ReadTheDocs documentation for full instructions (because if they did pip install, they likely also didn't create a virtual environment in conda, etc.)

AxonDeepSeg/train_network.py
vs74: But in the case of augmentation applied with Elastic and Gaussian Blur, i didn't see the behaviour of more than 3 classes. I am not sure that it impact our results.

notebooks/performance_metrics.ipynb
vs74: resolved in commit(5afb9bb)

CHANGELOG.md
vs74: Done in commit(992804f)

AxonDeepSeg/train_network.py
mathieuboudreau: The final output masks won't have more than 3 classes, because we likely thresholded them. We need to verify if the intermediate data (i.e. augmented dataset generated by these functions) have more than three classes. If you're telling me that they don't have a "nearest neighbour" function, then it's very likely that they'll be interpolating between the values (i.e. between 255 and 127 for the axon/myelin border), meaning more than 3 classes being used to train the data.

vs74:
> > However, I think that some instructions on pip install -e . in getting_started.ipynb needs to added, just in case a user doesn't run the command, I will do that.
> 
> No need to tell the user to do `pip install -e .` in that notebook, they should have already done that. All I would add is a link to our ReadTheDocs documentation for full instructions (because if they didn't do pip install, they likely also didn't create a virtual environment in conda, etc.)

Done in commit (4fab2ce)

AxonDeepSeg/train_network.py
vs74: Agreed and  I was wrong for the` ElasticTranform`, they do have interpolation mode of nearest neighbour( earlier, I must have seen the parameter of other augmentation operation on the documentation)

mathieuboudreau: That's great news! Do they have it for Gaussian as well?

vs74: I don't think there is a need to Gaussian as it just add noise to the data. And they don't have for Gaussian.

jcohenadad:
@mathieuboudreau we can force interpolation type with `ElasticTransform`, more info:
https://albumentations.readthedocs.io/en/latest/api/augmentations.html#albumentations.augmentations.transforms.ElasticTransform

AxonDeepSeg/train_network.py
vs74: Resolved Elastic Transform interpolation to Nearest Neighbour in commit(dae4a1d).

vs74:
> @mathieuboudreau we can force interpolation type with `ElasticTransform`, more info:
> https://albumentations.readthedocs.io/en/latest/api/augmentations.html#albumentations.augmentations.transforms.ElasticTransform

Implemented in commit(dae4a1d)

vs74:
@mathieuboudreau, is this PR reviewed ?

vs74:
@mathieuboudreau , shall we merge this PR, if it is reviewed ?

mathieuboudreau:
> @mathieuboudreau , shall we merge this PR, if it is reviewed ?

@vs74 I'll do one final pass tomorrow (Thursday), and retest the functionality hands on. If there's only minor issues, I'll open them as new separate issues so we can work on them individually.

ads_plugin.py
```diff
@@ -319,7 +319,10 @@ def on_apply_model_button(self, event):
         # Get the path of the selected model
         if any(selected_model in models for models in ads_utils.get_existing_models_list()):
             model_path = os.path.dirname(AxonDeepSeg.__file__)
```
mathieuboudreau: Instead of using mode_path as a dummy variable here that gets rewritten, we should fix it and use a different name because it was confusing for me to understand what was going on at first site. I'll make this change.

README.md
```diff
@@ -62,4 +62,4 @@ THE SOFTWARE IS PROVIDED “AS IS”, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR I
 
 ## Contributors
 
-Pierre-Louis Antonsanti, Stoyan Asenov, Mathieu Boudreau, Oumayma Bounou, Julien Cohen-Adad, Victor Herman, Melanie Lubrano, Antoine Moevus, Christian Perone, Vasudev Sharma, Thibault Tabarin, Maxime Wabartha, Aldo Zaimi.
+Pierre-Louis Antonsanti, Stoyan Asenov, Mathieu Boudreau, Oumayma Bounou, Julien Cohen-Adad, Victor Herman, Melanie Lubrano, Antoine Moevus, Christian Perone, Vasudev Sharma, Thibault Tabarin, Maxime Wabartha, Aldo Zaimi, Marie-Hélène Bourget.
```
mathieuboudreau: The list is in alphabetical order (last names), so I'll move Marie-Helene's name to the proper location.

CHANGELOG.md
```diff
@@ -5,7 +5,10 @@ Version [2.2dev] - XXXX-XX-XX
 -------------------------------------------------------------------------------
```
mathieuboudreau: As discussed at a previous meeting, this will be version 3.0, so we should address this in the changelog (and put the merge date in it)

AxonDeepSeg/train_network.py
mathieuboudreau: >and to make it easy to understand, I have made some comments on the top of each operation while keeping some empty lines.

The best way to make code easy to read is by following the PEP8 standard.

I'll do a pass of the new files in this branch using [black](https://github.com/psf/black) to apply these standards automatically.

AxonDeepSeg/models/download_model.py
```diff
@@ -0,0 +1,17 @@
...
+
+def main():
+    url_TEM_model = "https://osf.io/2hcfv/?action=download&version=4"  # URL of TEM model hosted on OSF storage with the appropriate versioning on OSF
+    url_SEM_model = "https://osf.io/sv7u2/?action=download&version=4"  # URL of SEM model hosted on OSF storage with the appropriate versioning on OSF
+
+    if (not download_data(url_TEM_model) and not download_data(url_SEM_model)) == 1:
+        print('Data downloaded and unzipped succesfully.')
+    else:
+        print('ERROR: Data was not succesfully downloaded and unzipped- please check your link and filename and try again.')
+    shutil.move("default_SEM_model", "AxonDeepSeg/models/default_SEM_model")
```
mathieuboudreau: Not sure if this command will work for Windows; maybe we should use a Path() object instead of hardcoding the path. Also, if called from a different folder than the root project, it might not place the data in the correct location. I'll open an issue, this is a minor issue but in need of a quick fix nonetheless.

AxonDeepSeg/ads_utils.py
```diff
@@ -282,4 +282,4 @@ def get_existing_models_list():
     return models_list
 
 # Call init_ads() automatically when module is imported
-init_ads()
+# init_ads()
```
mathieuboudreau: This line needs to be uncommented I think, as the error tracking depends on it. If doing so causes a problem, then we should open an issue about it.

.travis.yml
```diff
@@ -6,8 +6,9 @@ python:
 
 # command to install dependencies
 install:
+  - pip install yes
```
mathieuboudreau: Is `yes` actually required anywhere in the code? It's not used here in Travis. If it's needed in the code, then we should add it in requirement.txt

CHANGELOG.md
mathieuboudreau: Resolved in 3fc1032

README.md
mathieuboudreau: Resolved in cc23fa2

.travis.yml
mathieuboudreau: Resolved in f53197e

AxonDeepSeg/ads_utils.py
mathieuboudreau: Unable to resolve quickly - opened issue #285 about it.

ads_plugin.py
mathieuboudreau: Resolved in 971a46c

AxonDeepSeg/models/download_model.py
mathieuboudreau: Opened #286

AxonDeepSeg/train_network.py
mathieuboudreau: Resolved in 9ebed10

PR #234, opened 2019-07-16
Enforce Python 3.6.x version during setup

mathieuboudreau:
Changes
* Only allow ADS installation for Python versions 3.6.x 
* Revert Path() syntax in setup.y, which errors for versions before <=3.5 without a descriptive message (related to Python version).

If someone tried to install ADS with an incompatible Python version, they will now get something like following message.

>`AxonDeepSeg requires Python '>=3.6, <3.7' but the running Python is 3.5.2`

I looked into if I could customise this message or append one, but after a lot of searching it appears it's not possible (for unsuccessful installs).

### DIFF
--- a/docs/source/documentation.rst
+++ b/docs/source/documentation.rst
@@ -13,13 +13,14 @@ Version [2.2dev] - XXXX-XX-XX
 
 **Changed:**
 
-- Shifted AxonDeepSeg from TensorFlow to Keras framework.
-- Upgraded CUDA to 10.0 and tensorflow to 1.13.1.
 - Resolve image rescale warnings
 - Handle exception for images smaller than minimum patch size after resizing
 - Revert tensorflow requirekment to 1.3.0 and remove tifffile requirement
 - Remove `matplotlib.pyplot` from source code and refactor to full OO plotting
 - Standardize path management to `pathlib` library
+- Shifted AxonDeepSeg from TensorFlow to Keras framework.
+- Upgraded CUDA to 10.0 and tensorflow to 1.13.1.
+- Require Python 3.6.x version at installation
 
 
 Version [2.1] - 2018-09-25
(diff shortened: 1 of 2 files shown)


vs74:
The changes you made in changelog aren't reflecting in [RTD](https://axondeepseg.readthedocs.io/en/latest/documentation.html). There are two files of changelog that is  [docs/source/documentation.rst ](https://github.com/neuropoly/axondeepseg/blob/pypi_pyver/docs/source/documentation.rst)and [CHANGELOG.md](https://github.com/neuropoly/axondeepseg/blob/pypi_pyver/CHANGELOG.md). I think `CHANGELOG.md` is the main file which reflects changes made and appropriate changes must be made there also. But still I don't understand why there are 2 files of changelog instead of one.

mathieuboudreau:
>But still I don't understand why there are 2 files of changelog instead of one.

This was discussed at the last meeting, it's been like this for a while. Up until now, any changes always had to be added in both CHANGELOGS (I've corrected that for your PRs a few times).

I am going to look into how to remove the hardcoded changelog in the ReadTheDocs files and simply import the CHANGELOG.md file.

mathieuboudreau:
>The changes you made in changelog aren't reflecting in RTD. 

That is because the RTD link you linked here points to "latest" which is the version in the master branch. Only after this branch is merged will the changes be reflected there.

mathieuboudreau:
>Shouldn't we change date as well in the latest version of ADS (that is 2.2dev) ?

This PR isn't releasing a new version of ADS yet (which would warrant upgrading to 2.3). Per discussed at the last meeting, this will be done once your data augmentation branch is merged (since it will be a more "complete" keras upgrade than the current "partial" one).

The date on the changelogs will be updated accordingly once that release is done.

mathieuboudreau:
To test this PR, what should be done is try installing (`pip install -e .`) in three different virtual environments: one with python <3.6, one with >=3.7, and one with a 3.6.x (x can be any number between 0 and 8).

You can create a conda env with a specific python version using the following command:

`conda create -n ads python=3.6.8`

Only the 3.6.x one should install correctly; the other two should give you the warning that I described in the description of this PR

vs74:
> To test this PR, what should be done is try installing (`pip install -e .`) in three different virtual environments: one with python <3.6, one with >=3.7, and one with a 3.6.x (x can be any number between 0 and 8).
> 
> You can create a conda env with a specific python version using the following command:
> 
> `conda create -n ads python=3.6.8`
> 
> Only the 3.6.x one should install correctly; the other two should give you the warning that I described in the description of this PR

I followed the instructions and for `python version < 3.6` it's working fine but not for `python version >= 3.7`. I am on  `Python 3.7.3 ( Conda virtual environment)` and it throws no warning. Can you recheck on `python >=3.7 `

PR #238, opened 2019-07-30
Remove session close exception

vs74:
While performing segmentation though ADS performs segmentation it throws the following session closed exception. This quick PR aims to resolve this issue.
Screen Shot 2019-07-30 at 2 39 19 PM

### DIFF
--- a/AxonDeepSeg/apply_model.py
+++ b/AxonDeepSeg/apply_model.py
@@ -160,8 +160,7 @@ def apply_convnet(path_acquisitions, acquisitions_resolutions, path_model_folder
             # Update of the predictions lists.
             predictions_list.extend(current_batch_prediction)
 
-    # End of the inference step. Closing Tensorflow.
-    sess.close()
+    # End of the inference step.
     tf.reset_default_graph()
 
     # Now we have to transform the list of predictions in list of lists,


mathieuboudreau:
That's odd; this line (`sess.close()`) has been in the software for a very long time (a couple of years at least), and it hasn't caused an issue like yours before. Why do you think this is - is it possible it has to do to with the Keras upgrade?

@perone @alzaia I'm not as well versed in tensorflow as you both; is there any risk (or ressource-wise) to not closing the tensorflow session here?

mathieuboudreau:
Unless there is somewhere that we recently deleted a statement that opens a session - maybe that was in the bulk-delete that was done during the Keras transition?

mathieuboudreau:
Something about this solution just doesn't feel clean to me; maybe we could. take a look at a couple of the recent merges to see when this issue began.

vs74:
It's true unit tests didn't catch (I don't know why) this though at the end of each travis build you could see the following  exception.

vs74:
> That's odd; this line (`sess.close()`) has been in the software for a very long time (a couple of years at least), and it hasn't caused an issue like yours before. Why do you think this is - is it possible it has to do to with the Keras upgrade?
> 
> @perone @alzaia I'm not as well versed in tensorflow as you both; is there any risk (or ressource-wise) to not closing the tensorflow session here?

When we were on `TensorFlow 1.3.0` we were not encountering this exception but since when  we shifted to `TensorFlow 1.13.1` this error came into picture.

mathieuboudreau:
Yup I just saw that while comparing Travis logs. It means that something about how tensorflow sessions are started/stopped changed; we should look into that to make sure we're doing it the proper way.

Issue #139, opened 2018-10-29
Potential issue with gold standards (or fifty shades of myelin)


perone:
I was experimenting with a model and I used data from the folder `SEM_3c_512` (Duke) and I found that the gold standards contained a range of pixel intensities that weren't one of the three classes [background, myelin, axon]. For instance, if I just want the myelin content, I threshold for the myelin pixel intensity and it creates a ground truth like this one below:

![image](https://user-images.githubusercontent.com/412328/47664983-8348a480-db76-11e8-9724-a90413e60a19.png)

![image](https://user-images.githubusercontent.com/412328/47665092-ba1eba80-db76-11e8-9ab7-c6241c1fb360.png)

Where we can see the noisy pixels on many places. Please note that GitHub _is resampling these images_, so they look less bad then they're in reality.

When we look on the original gold standards, we can see that there are indeed many different shades representing the classes:

![image](https://user-images.githubusercontent.com/412328/47665213-fce09280-db76-11e8-88d0-c370ea72fa52.png)

This might be an effect of resampling of the gold standards, that might be creating interpolation on the gold standard and therefore messing up with the class definitions. This may be a problem for AxonDeepSeg if the training is being done without taking into consideration these issues, but I'm not sure how these gold standars are being handled today on AxonDeepSeg. I'm not sure if the data used for training is also from this same folder on Duke.

mathieuboudreau:
I suspect this line could be the source of this issue:

https://github.com/neuropoly/axondeepseg/blob/9e59f536da6c709d0341999ae2b11945de588cb2/AxonDeepSeg/data_management/dataset_building.py#L48

I'm not 100% sure, but by quickly browsing Duke, it appears to me that this issue only happens to patched datasets, not the raw masks. I can't confirm though, as I am not sure yet where the matching raw datasets from the "gold standard" are located. @perone would a nearest-neighbour interpolation be the best approach? Currently we're using the default (`order=1` -> bi-linear), whereas setting `order=0` would use a nearest neighbour interpolation.

Bonus: these lines could also be creating to a similar problem that's hidden from the end-user during training/data-augmentation:

https://github.com/neuropoly/axondeepseg/blob/9e59f536da6c709d0341999ae2b11945de588cb2/AxonDeepSeg/data_management/data_augmentation.py#L79

[...]

PR #242, opened 2019-08-01
[BUG] Switch to nearest neighbour interpolation when creating patches of mask images.

mathieuboudreau:
Related to #139 and our discussion at the last group meeting.

Changes:

* Created a new function to check the number of unique values in an image, and what those values are, and add test.
* Switch to nearest neighbour interpolation when creating patches of mask images (`rescale()` from scikit-image needs to be passed `order=0` to do this).
* Add test that masks created by the data management tool output patches with only three unique values (without the above fix, this test failed **!**).

There is likely several other spots in the code that might be introducing shades of grey (e.g. during data augmentation), but I think that this PR will be a large leap forward in the right direction enough that it warrants to be merged ASAP.


mathieuboudreau:
* [ ] Note to myself: send raw image to @vs74  so that he may test it in this branch

mathieuboudreau:
@vs74 there are already files ready and formatted (named) the correct way in the project that you can use to test this branch with. They are located [here]()

All you need to do is run the following script in your `master` branch:

```python
from pathlib import Path
from AxonDeepSeg.data_management.dataset_building import *

test_data = Path("/Users/mathieuboudreau/neuropoly/github/ads_grey/test/__test_files__/__test_patch_files__/")
raw_images = test_data / 'raw'
patched_output = test_data / 'patch'
raw_img_to_patches(str(raw_images), str(patched_output))
```

Substituting `/Users/mathieuboudreau/neuropoly/github/ads_grey` for your full path to your AxonDeepSeg clone.

Look at the mask files in the patch folders using an image editor: at the boundaries, they have values between the three classes (0, 128, 255).

Then, checkout the `mask_interpolation` branch (this PR), and rerun the script. Check out the mask files in the patch folders, and all pixels should only contain one of those three classes

vs74:
Thank you, @mathieuboudreau for the detailed instructions. I will review this PR.

vs74:
This PR is working as expected.

PR #243, opened 2019-08-02
Fixed compatibility with 16 bit files, introduced ads_utils.imread() for data loading

### LINKED ISSUE #166: [Bug] Patches truncated for 16 bit images
Likely due to this line: https://github.com/neuropoly/axondeepseg/blob/57047da51781c12d06c1b5d59dcb705bf4adc376/AxonDeepSeg/data_management/dataset_building.py#L44

Combining `preserve_range` plus `asint` may be what's truncating the patches (all white).

capture d ecran 2019-02-22 a 14 17 54

The solution may be as simple as removing the `asint` call.

@maxwab you wrote that line; do you see any issues with removing the `asint` call? I'll double check the rest of the code to make sure it's not assumed to be an 8bit int anywhere else. I guess it means that if I remove this, the data size will be much larger... Otherwise, removing the perserve_range could also possibly be a solution.

### LINKED ISSUE #170: Extend functionality to handle images with any bit depth
Extend functionality of AxonDeepSeg to handle images with any bit depth (would resolve #166), currently can only handle 8 bit.

**Things that may need to be done:**
  * [ ] Identify what combination of `.astype(int)`, ` preserve_range=True`, and `“L” mode` are causing the truncation for 16 bit.
  * [ ] Once identified, fix it in a way that retains the current functionality (i.e. either do better conversion to 8-bit that doesn’t truncate, or don’t do a conversion at all).
  * [ ] Write tests to verify that the new functionality doesn’t truncate 16b images anymore.
  * [ ] Demonstrate that the new functionality works for a non-8 bit image (or image dataset) in both the data patch splitting set and the training step.

**Potential roadblocks:**
* Lots of function that may be doing this conversion.
* Some functions may look at first glance like they might be doing this conversion, but may not and should remain untouched.
* Not converting high-bit images might result in large dataset sizes, so possible local space issues or tensorflow space issues.
* Ensure that unit tests pass, and if not identify if it’s because the changes made changed the API and:
  * could be avoided by proper refactoring
  * can’t be avoided with the new functionality, and the tests need to be changed.

mathieuboudreau:
Resolves #166  and Resolves #170

Changes:

* Create image i/o functions  to centralise image reading and writing in functions of the `ads_utils` module. ([here](https://github.com/neuropoly/axondeepseg/pull/243/files#diff-bb7b85302b2b6a61c4fb63cce64c038fR253-R272))
* Wrote those functions to correctly convert non-8bit images (e.g. 16 bit images) to 8bit (this is the default option, but can chose any bit depth to be the range in the loadedvariable).
* Refactored all AxonDeepSeg files to use this new i/o functionality (e.g. calling `ads.imread()` instead of `imagio.imread()`).


mathieuboudreau:
@vs74 or @jcohenadad : This branch is ready to review. AxonDeepSeg should now correctly handle 16 bit images (or any other bit value).

To test: 

* Download this 16 bit image and rename it to image_16b.png:

image_16b

* Checkout master and make sure it's `pip install -e .` was run.
* Run the following command in the folder with the above image: `axondeepseg -t SEM -i image_16b.png -s 0.07` 
* You should notice that the command outputs an error, and didn't properly segment the image.
* Now checkout this branch (`mb_issue170`); if you didn't to a fresh clone, make sure you `git pull` after.
* `pip install -e .` again
* Run the following command in the folder with the above image: `axondeepseg -t SEM -i image_16b.png -s 0.07` 
* The image segmentation should complete succesfully, and output well-behaved masks.

mathieuboudreau:
Mathieu to do:

* [ ] Verify that notebooks use the new image i/o functions

mathieuboudreau:
@vs74 or @jcohenadad This PR is now ready for review. Instructions on how to do so are located in this comment: https://github.com/neuropoly/axondeepseg/pull/243#issuecomment-531874814

PR #246, opened 2019-08-12
Memory issue - Loads image as single channel instead of 3 channels 

### LINKED ISSUE #244: Does Keras require images to be loaded with 3 channels to train data?
Question for you @vs74,

I just fell on a funny bug in one of my branches (see PR #243 ) while refactoring one of your recently modifications to use Keras.

I noticed that you're loading training data with CV2, e.g.

https://github.com/neuropoly/axondeepseg/blob/a7da6a1e7bf56a84b917045b9cdee1b8d0d48246/AxonDeepSeg/data_management/input_data.py#L39

and that this loads data with three channels (even if the images are in B&W). For masks you don't load with three channels:

https://github.com/neuropoly/axondeepseg/blob/a7da6a1e7bf56a84b917045b9cdee1b8d0d48246/AxonDeepSeg/data_management/input_data.py#L42

But you create three channels right after anyways to hold your three tissue times:

https://github.com/neuropoly/axondeepseg/blob/a7da6a1e7bf56a84b917045b9cdee1b8d0d48246/AxonDeepSeg/data_management/input_data.py#L43

https://github.com/neuropoly/axondeepseg/blob/a7da6a1e7bf56a84b917045b9cdee1b8d0d48246/AxonDeepSeg/data_management/input_data.py#L94-L99

**This feels like it might hog the memory, as it stores 3x the amount of data necessary as before**, and might impact us when training large datasets.

If I try loading these with imageio's imread as 8bit (as we do everywhere else in AxonDeepSeg), Keras throws me an error, as it appears it expects to have more dimensions (because imageio is only giving me 2 dimensions, and CV2 is giving us 3).

```
E           ValueError: ('Input data in `NumpyArrayIterator` should have rank 4. You passed an array with shape', (2, 256, 256))
```

See Travis log: https://travis-ci.org/neuropoly/axondeepseg/builds/566773230?utm_source=github_status&utm_medium=notification

Is there are way to get around this that we can load 8bit B&W images instead of your RGB to get this working with Keras?

vs74:
Currently, ADS was using cv2 to read image as RGB since it might generate memory hog while training with large dataset. This PR aims to load image as GRAYSCALE.
**Changes**

- Retrain `SEM` and `TEM` model 
- Modified `cv2` type in `AxonDeepSeg/data_management/input_data.py` to read `GRAYSCALE `images
- Made appropriate changes in `AxonDeepSeg/network_construction.py` to predict batch of single dimmension
Fixes #244


mathieuboudreau:
@vs74 after I merge this branch to master, because you needed to retrain the models here and currently have another branch where the models were retrained (#226),  you will need to merge master into that branch and retrain the models with the changes that were also introduced here.

vs74:
@mathieuboudreau, we can first merge this branch and I will upload the retrained models from this branch into OSF storage which will be then merged into PR #226, once it's completed.

Issue #252, opened 2019-09-03
Add a button to the GUI that computes and returns metrics


Stoyan-I-A:

Stoyan-I-A:
Done in #256

PR #256, opened 2019-09-06
[GUI] Add buton to compute and save morphometrics

mathieuboudreau:
This branch adds a button in the FSLeyes GUI to compute and save morphometrics from a segmented image.

**Workflow**

After you've loaded and segmented an image using the ADS prediction model, you click the new "Compute morphometrics" button here:

Capture d’écran 2019-09-06 à 14 34 05

which will compute the morphometrics for each axon. After that's done (a few seconds), a popup window will appear for you to select where to save the Excel file containing these data, and to name it:

Capture d’écran 2019-09-06 à 14 36 17

You can then double click the file after it's saved:

Capture d’écran 2019-09-06 à 14 36 51

And view the table of morphometrics data for this image:

Capture d’écran 2019-09-06 à 14 37 17

### DIFF
--- a/AxonDeepSeg/morphometrics/compute_morphometrics.py
+++ b/AxonDeepSeg/morphometrics/compute_morphometrics.py
@@ -45,7 +45,7 @@ def get_pixelsize(path_pixelsize_file):
         return pixelsize
 
 
-def get_axon_morphometrics(im_axon, path_folder, im_myelin=None):
+def get_axon_morphometrics(im_axon, path_folder=None, im_myelin=None, pixel_size=None):
     """
     Find each axon and compute axon-wise morphometric data, e.g., equivalent diameter, eccentricity, etc.
     If a mask of myelin is provided, also compute myelin-related metrics (myelin thickness, g-ratio, etc.).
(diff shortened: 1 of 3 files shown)


mathieuboudreau:
@Stoyan-I-A my PR here is simple enough that I thought it might be mergeable into your own FSLeyes branch, which is what will currently happen if it is approved.

If you'd prefer not to risk it and wait until yours is merged into master to approve this one, I'm 100% with it. I'll leave it up to you to decide.

mathieuboudreau:
One thing to note, which is a minor annoyance, is that according to our old convention, the pixel location (0,0) is top left (which is what image editors like GIMP use), whereas the FSLeyes convention in the GUI sets it at the bottom left of the image. So users looking at the morphometrics file will have (x0, y0) axon locations that are flipped from the numbers FSLeyes show. I don't think this is more so an issue with this GUI button, but moreso with the actual implementation of `compute_morphometrics.py`, so I didn't tackle it in this PR. It's likely that we might want to flip the morphometric y values (but then, if the user uses another editor like GIMP later on, then they will see those locations as flipped...). Anywhom, just a note.

Stoyan-I-A:
I think it is okay to merge this PR to my branch. I'll just need to make sure we use the corrected axon mask if there is one. ([see last meeting's notes](https://docs.google.com/document/d/12MSxgeMiOxm43AMhkXW9IhP-Cmpxn2ft00CANA3xjYA/edit))

This PR fixes #252

PR #261, opened 2019-09-20
Shift Models to OSF

### LINKED ISSUE #240: Repository is too large (350MB)

vs74:
This PR aims to download models from` OSF` storage. Right now, the implemented script is able to download models from OSF storage via the procedure given below:-

**STEPS TO DOWNLOAD MODEL FROM OSF STORAGE**

`pip install -e .`- **Run this command in CLI**

**TODO**  

- [x] Fix the directory in which models are downloaded - Right now it's being downloaded in Parent directory, instead of `AxonDeepSeg/models` directory

- [x] To figure out how to link GIT and OSF; naming conventions of the models, compatibility between these platforms for easy version control

Fixes #240


vs74:
@mathieuboudreau, is there a way to enforce `axondeepseg_models` command when ADS is cloned? I meant like when ADS is downloaded, we could run this command in ADS script ( being first in priority, though I don't know how to do this), and by this way we could download models from OSF storage.

mathieuboudreau:
Yeah I was wondering the same thing. I don't know about running a command-line command, but I feel like there must be a way to run a python script/function during the setup stage (when `pip install -e .` is called), so somewhere in setup.py. I just took a quick look and this forum post suggests one method, but I haven't tried it myself : https://stackoverflow.com/questions/20288711/post-install-script-with-python-setuptools

jcohenadad:
> Yeah I was wondering the same thing. I don't know about running a command-line command, but I feel like there must be a way to run a python script/function during the setup stage (when `pip install -e .` is called), so somewhere in setup.py. I just took a quick look and this forum post suggests one method, but I haven't tried it myself : https://stackoverflow.com/questions/20288711/post-install-script-with-python-setuptools

[this solution](https://stackoverflow.com/posts/36902139/revisions) from @mathieuboudreau's link seems very elegant

vs74:
> > Yeah I was wondering the same thing. I don't know about running a command-line command, but I feel like there must be a way to run a python script/function during the setup stage (when `pip install -e .` is called), so somewhere in setup.py. I just took a quick look and this forum post suggests one method, but I haven't tried it myself : https://stackoverflow.com/questions/20288711/post-install-script-with-python-setuptools
> 
> [this solution](https://stackoverflow.com/posts/36902139/revisions) from @mathieuboudreau's link seems very elegant

I will try implementing the same.

vs74:
I tried calling script, unfortunately it didn't work out. Till the time being I figure it out, I have a temporary fix to it which is given below:-
```
1. pip install  -e .
2. python3 setup.py bdist_wheel --universal 
```
**NOTE:** The same thing can be deployed in Travis CI in place of `pip install .` and it will download models from OSF but right now the models are not being saved in the required directory which needs to be fixed.

vs74:
> I tried calling script, unfortunately it didn't work out. Till the time being I figure it out, I have a temporary fix to it which is given below:-
> 
> ```
> 1. pip install  -e .
> 2. python3 setup.py bdist_wheel --universal 
> ```
> 
> **NOTE:** The same thing can be deployed in Travis CI in place of `pip install .` and it will download models from OSF but right now the models are not being saved in the required directory which needs to be fixed.

@mathieuboudreau , Do let me know if this solution seems feasible or should be go with calling the script directly?

mathieuboudreau:
Why call the setup.py file twice (that'swhat pip install -e . does) instead of just omiting the second one and calling your commandline tool you created directly?

Regardless, I don't find this to be a good enough solution; I'll look into it more.

mathieuboudreau:
Done in 0ea5aeffc4 - now works with simply `pip install -e .`

jcohenadad:
might be relevant for debugging setup.py: https://stackoverflow.com/questions/55769962/how-to-debug-pip-install-package-name

vs74:
Sorry Deleted this PR by mistake. @mathieuboudreau I have now replaced the download functionality using a function in` Setup.py`. However, it's not downloading the models in the required directory. Can you take a look at it (ideally the downloaded models should be present in the `AxonDeepSeg/models` directory instead of main directory) ? By the time being, I will focus on your issue related to training.

vs74:
@mathieuboudreau, this branch is ready for review. However, the travis will fail as of now, to get it working `pip install .` needs to replaced with` pip install -e .`. The reason for the change is that if download model script from OSF storage which was earlier implemented in different file but if it is now implemented in `setup.py`, then it will require dependency ( like requests and other dependencies present in `ads_utils.py` script), but since the AxonDeepSeg package is not developed prior to importing `AxonDeepSeg.ads_utils.download_model`, ` pip install . ` does not work.

vs74:
You could verify the downloading of models from OSF storage on your local machine by following the below steps:-

1. Create a virtual environment
2. Run command `pip install -e .`

mathieuboudreau:
Ok I will check it out. 

We always instructed users to install using `pip install -e .` (see documentation [here](https://axondeepseg.readthedocs.io/en/latest/documentation.html#development-version)), so I don't quite understand why you were trying with `pip install .`. Nonetheless, if it works the old way, that should be great!

mathieuboudreau:
Ahh I understand now; for some reason, the Travis config file was running `pip install .` instead of `pip install -e .`. I just fixed that in this branch, lets see if Travis passes.

vs74:
> @vs74 @jcohenadad I rolledback the changes Vasudev recently made (removing the script and using only the setup.py file to download files), because to get the Travis tests to run correctly & user installs it would have gotten much messier than we'd want.
> 
> * With that implementation, we had to do `pip install -r requirements.txt` to first download all the required packages, only to then run `pip install -e .` to install ADS, which would try and install the packages a second time.
> * There were a bunch of other hacks needed here and there too, so I just gave up and decided to roll it back.
> 
> @vs74 one additional note for you:
> 
> * You had used a `system()` call inside a Python script to move the model folders after they were downloaded.
> 
> https://github.com/neuropoly/axondeepseg/blob/227ad80254d9bab9a509952b730cd6035215e7ba/setup.py#L35
> 
> I would recommend never doing this, because in almost all cases I can think of (besides running unit/integrity tests on CLI commands), there is an available Python tool to do the job you need inside that environment. For this case, I simply needed to called `shutil.move()`.
> 
> Finally, to get the installation running correctly using the old way (with a CLI script that's run in the post-build), I needed to turn off Sentry bug tracking, since the way I had it setup the prompt for that would come up anytime ads_utils was imported, and `download_data()` was being imported from that module to download your models. For users this wouldn't have been an issue, but Travis was crashing (even if I tried `yes | command`). I can try to fix this in the future, but we weren't getting that many good reports from Sentry in the past anyways (nothing really usable; mostly bad argument/values errors).
> 
> Note that Travis is now running as expected, but failing. This is because your models appear to be out of date with the base of this branch (vs_data_aug_keras); you'll have to resolve these tests by doing the appropriate changes there.
> 
> I'm going to go ahead and merge this now into that branch.

That's odd, when I tried downloading models from new implementation, I had just to write `pip install -e .` and it internally downloads all the requirements in addition to the models.

mathieuboudreau:
Yeah, that worked for me on my computer, but didn't for Travis. Likely because I may already of had some of the required packages installed before calling that (e.g. `requests`).

jcohenadad:
thanks a lot for these investigations @mathieuboudreau and @vs74!

PR #267, opened 2019-10-15
Changes to the FSLeyes plugin

Stoyan-I-A:
Changes :
 - Button labels are now darker
- Inverted Y axis (see https://github.com/pauldmccarthy/fsleyes/issues/30)


mathieuboudreau:
@Stoyan-I-A it's best if new branches are created (from master) once a pull request as been merged, as it's cleaner. Working on old branches results in a lot of commits showing up, even though they've already been merged. We'll find a way around this, just wanted to give you a heads up moving forward.

mathieuboudreau:
As you can see, there is not a merge conflict that will need to be addressed (which typically shouldn't be an issue if it was branched from master)

mathieuboudreau:
Both changes appear to work for me! Buttons text is darker (although even darker, or bold, would be nice)

Capture d’écran 2019-10-15 à 11 44 30

And the (0,0) coordinate is at the top left for me. 🎉

Issue #210, opened 2019-04-24
FSLeyes integration - Fetch models based on physical presence in folder /models


jcohenadad:
Instead of hard-coding the names in the code. It will be more flexible

Stoyan-I-A:
Implemented in #275

PR #275, opened 2019-11-08
Centralize mask cutoff values & get_existing_models_list

### LINKED ISSUE #274: Centralize mask cutoff values
See https://github.com/neuropoly/axondeepseg/pull/271#issuecomment-547520293

Stoyan-I-A:
Changes:
- Added params.py file
- Used params in the plugin file where appropriate
- The plugin now displays the available models based on their presence in the models folder

Edit: There are places where I didn't use params yet because I'm afraid that some of our output functions still use hardcoded values not in params.py and the FSLeyes plugin won't be able to read them correctly.

Fixes #274 , #210


mathieuboudreau:
What's the difference between `singe_mask_file_intensity` and `axon_intensity` (they both have the value 255).

Stoyan-I-A:
> What's the difference between `singe_mask_file_intensity` and `axon_intensity` (they both have the value 255).

AxonDeepSeg outputs 3 images : 
- One axon+myelin image
- One axon-only image
- One myelin-only image

Since images with a single mask are saved with a value of 255, it is less confusing to use the `single_mask_file_intensity` parameter if we operate on an image that happens to be a myelin-only image.

AxonDeepSeg/params.py
```diff
@@ -0,0 +1,4 @@
+
+singe_mask_file_intensity = 255
+axon_intensity = 255
```
jcohenadad: could be a dict, e.g.:
intensity  = {
  'axon': 255.
  'myelin': 127'
}

jcohenadad: instead of hard-coded 255, we can replace with `np.iinfo(np.uint8).max`
same for the 127 (make sure to round it properly)

ads_plugin.py
```diff
@@ -895,6 +875,17 @@ def get_visible_myelin_overlay(self):
 
         return myelin_overlay
 
+    def get_existing_models_list(self):
```
jcohenadad: maybe move this function under `ads_utils`

vs74:
@Stoyan-I-A, is this branch ready to be merged so I can make the changes in the model name that is to remove the suffix _v1 https://github.com/neuropoly/axondeepseg/issues/278. If not,  may I know how long will it take, in case some work  needs to be done for your branch,  I will defer my changes.

Stoyan-I-A:
@vs74 ,this branch is not ready yet, but I just need to make a small modification. So, this branch should be ready in about 2 days.

Stoyan-I-A:
This branch is now ready for review/merge

PR #276, opened 2019-11-26
Add split_data() function and tests

mathieuboudreau:
Splits a dataset into Train and Validation folders.

See [this test](https://github.com/neuropoly/axondeepseg/blob/21b3a59bc9db6d7a0dc2f2b8b755090a7e426141/test/data_management/test_dataset_building.py#L184-L194) for an example usage. You could try it out with the TEM dataset instead (https://osf.io/uewd9/), which has many more images to split.


AxonDeepSeg/data_management/dataset_building.py
```diff
@@ -247,3 +248,50 @@ def find_minority_type(SEM_patches_folder, TEM_patches_folder):
         len_majority = SEM_len
 
     return minority_patches_folder, len_minority, majority_patches_folder, len_majority
+
+def split_data(raw_dir, out_dir, seed=2019, split = [0.8, 0.2], override=False):
```
vs74: I think one better way to split the dataset is during training itself using Keas API. For more details look at keras `model.fit `method , refer here https://keras.io/models/model/. Here using` validation split =0.2 `, we can split the data into training 80 % of the data and validation 20 % of the data. For additional details on how the shuffling and spilt is performed, https://stackoverflow.com/questions/51006505/how-training-and-test-data-is-split-keras-on-tensorflow#comment89010893_51006505 is good resource. Personally, instead of writing our own method we could use splitting API of keras during training.  @mathieuboudreau, let me know what  you think about it ?

vs74: Also note, I don't know whether this will create directories for train and validation separately. And, also, if we think of implementing it, some small changes into how  training the models had to be  carried out.

mathieuboudreau:
Thats fine, as long as its perfectly compatible with 1) our input data (osf datasets), and 2) the required qplitted dataset folder tree (i.e. what the folder tree needs to be for the data to train as-is), and 3) it accepts a fixed seed for reproducible splits (in case we want to reproduce a data split from download from OSF). Also, if done durring training, does it copy or move data? Copying will use more space for large datasets.

If you could demonstrate that for all three points with Kera, that works for me. Otherwise, my function could wrap around it.

Télécharger Outlook pour Android<https://aka.ms/ghei36>

AxonDeepSeg/data_management/dataset_building.py
mathieuboudreau: In that case, maybe we should move ahead with my implementation for the time being, as your other branch is being held up by the OSF i/o (needs this feature to work smoothly in notebooks). We could then open an issue and look into your proposed method later on. What do you think @jcohenadad

vs74:
@mathieuboudreau , shouldn't this functionality be illustrated in `guide_dataset_building.ipynb` notebook as well ?

mathieuboudreau:
@vs74 great point - I missed that! I'll do it for that one and subsequently the training notebook too, so that they follow eachother.

vs74:
@mathieuboudreau, I was thinking maybe could you include instructions on `guide_dataset_building.ipynb` and` training_guideline.ipynb` so that this PR is complete and we can merge it before merging data augmentation PR.

mathieuboudreau:
Will do ASAP!

Télécharger Outlook pour Android<https://aka.ms/ghei36>

mathieuboudreau:
@vs74 this branch is ready to be reviewed and/or merged. I updated the data management and training notebooks so that they now use this function, as well as downloads the OSF data for a demo. All you need to do now us run all cells in the guide_dataset_building.ipynb notebook and then run all cells in the training_guideline.ipynb notebook to train an SEM dataset; no modifications required. @jcohenadad might be a good thing for @mariehbourget to test as well, this notebook workflow would ideally be very user friendly even for completely new users (after they've installed everything).

vs74:
@mathieuboudreau , sorry for the late reply. I tested this branch and it's working fine for me and thanks for that. Should I merge this branch or wait for @mariehbourget to  test it ?

mathieuboudreau:
@vs74 Great! Let's merge it now, @mariehbourget can test it whenever it's merged.

mariehbourget:
Hi @mathieuboudreau and @vs74, I have tested the 2 new notebooks on that branch and everything works fine and is clear to me.
However, I was wondering if it is normal that this PR was closed without being merged?

mathieuboudreau:
>However, I was wondering if it is normal that this PR was closed without being merged?
 Delete branch

Good catch, @mariehbourget ! I thought @vs74 merged it when he closed the issue.

vs74:
> > However, I was wondering if it is normal that this PR was closed without being merged?
> > Delete branch
> 
> Good catch, @mariehbourget ! I thought @vs74 merged it when he closed the issue.

Sorry for that I closed it without merging

PR #290, opened 2020-03-23
Add missing import to apply_model

### LINKED ISSUE #289: Error after training finishes (Test image display issue?)
**Describe the bug**
After running the training file, the training runs fine but after it is done I get an error message which I am fully unsure what it is. The test image doesn't display either. I used a master version of branch from around early-mid February.
Console log: [error.txt](https://github.com/neuropoly/axondeepseg/files/4363126/error.txt)

**To Reproduce**
Run the training file.
Training file (Included both as notebook and python file): [training.zip](https://github.com/neuropoly/axondeepseg/files/4363143/training.zip)

**Desktop (please complete the following information):**
 - OS: [openSUSE]
 - Version [Leap 15.1]

mathieuboudreau:
Apply model calls `np` on [line 183](https://github.com/neuropoly/axondeepseg/blob/274ab3748e214f53c073f75cd6b98e1db84eab4f/AxonDeepSeg/apply_model.py#L103), but [doesn't import it first](https://github.com/neuropoly/axondeepseg/blob/274ab3748e214f53c073f75cd6b98e1db84eab4f/AxonDeepSeg/apply_model.py#L1-L14)

Resolves #289


vs74:
As mentioned in issue #289, there was missing key value for `data augmentation reflection mode`. `config_tools.py` script had missing key and value pair for this data augmentation mode, but I don't know why the build tests were passing for the build. Might resolve the problem faced.

jcohenadad (dismissed):
disclaimer: I have not tested the new code

mathieuboudreau:
@mariehbourget could you confirm that we do indeed want `'da-6-reflection_border-activate':True,` to be added to the config tool? @vs74 I don't remember why/not sure why you added it to the config tool in this PR, since my original PR was just an unrelated quick fix.

mariehbourget:
> @mariehbourget could you confirm that we do indeed want `'da-6-reflection_border-activate':True,` to be added to the config tool?

I'm not sure what this option is and why it is set to True. In our models on OSF, this key is false.
The default_configuration in outdated and other keys are missing to. The tests are passing anyway because all keys of the tests config are present. I've done some research on this for the [PR301](https://github.com/neuropoly/axondeepseg/pull/301) and I'll will be talking about it in the meeting today.

mathieuboudreau:
@mariehbourget can we revisit the issue of whether or not we want `'da-6-reflection_border-activate':` to be true or false by default in the config, in this PR? What's your opinion?

mariehbourget:
I would set it to false only because it is how it is set in our default models config json files.

I think it’s a good idea to add the key in the default configuration for consistency. However, it doesn’t really matter because the default calibration is used only in the [validate_config](https://github.com/neuropoly/axondeepseg/blob/220f6ba46391aa97edd7a51d3fcbbde0ccd36781/AxonDeepSeg/config_tools.py#L21) function. Although there is a [test](https://github.com/neuropoly/axondeepseg/blob/220f6ba46391aa97edd7a51d3fcbbde0ccd36781/test/test_config_tools.py#L80) for that function, it is not called anywhere for training or inference.

The default calibration is also used for inference [here](https://github.com/neuropoly/axondeepseg/blob/85686c983399a07f26b1dc10aea2b006ce867cec/AxonDeepSeg/apply_model.py#L247), but no data augmentation occurs in inference.

So setting `da-6-reflection_border-activate` to true or false by default won’t change the behavior.

mathieuboudreau:
Great, thanks for the feedback @mariehbourget !

mariehbourget (approved):

Issue #293, opened 2020-04-14
TEM data not segmenting well in version 3.0


mathieuboudreau:
Just discovered this while looking into #249 (but I think this problem is unrelated to that one).

* Downloaded the most recent version of AxonDeepSeg
* Downloaded one of the TEM images used in for the training on OSF as a sample dataset  to test with: https://osf.io/uewd9/
* Ran the command line segmentation call: axondeepseg -t TEM -i image.png -s 0.00236 
* Output was very poorly segmented (see below. Images in order of the gif are: Raw image, manually segmented mask, ADS mask).

![out](https://user-images.githubusercontent.com/1421029/79281832-2e920a00-7e8e-11ea-9333-24d656513f9b.gif)

I tried the same steps with SEM, but that segmentation resulted ok:

![out](https://user-images.githubusercontent.com/1421029/79281846-3782db80-7e8e-11ea-9e4c-fff7d4320495.gif)

mathieuboudreau:
I'm currently re-processing the TEM model, I'll inform you folks of if this resolved the issue when it's done.

panpeterlin:
Hi,  I also downloaded the latest version and tried it on test TEM image and it did not work well. Have you solved this issue or not?

mathieuboudreau:
I tried training a new model and it did not work, which was suspicious since training worked well for SEM data. The only main difference between the two in the code that I could think of was in fact the block of code talked about in #249, so I removed the TEM condition and it ended up segmenting well with the old trained model. So all that should be needed is merging PR #294 - no need to re-upload/download a trained TEM model.

@panpeterlin we'll let you know when it's resolved in the `master` branch, currently the fix is only in the branch `tem_flip`. You'll have to pull master again after it's merged.

mathieuboudreau:
See here for new gif of segmented TEM images (raw image, manual mask, segmented image):


![out](https://user-images.githubusercontent.com/1421029/79345386-9dfa0f00-7f0b-11ea-90c6-05bb0f029506.gif)

mathieuboudreau:
Resolved in #294. @panpeterlin if pull master or redownload the software, the issue should be resolved for you too.

Issue #249, opened 2019-08-14
Why is TEM image contrast being flipped, when model should be train on native TEM contrast


mathieuboudreau:
See this code block:

https://github.com/neuropoly/axondeepseg/blame/aa285598371b089242db832ef2a4e8f804c7892c/AxonDeepSeg/segment.py#L64-L67

and issue #245

mathieuboudreau:
Answer: because the contrast of the TEM patches that we trained were manually inverted. I'm going to try training an updated model with the uninverted contrast, and remove this hack.

jcohenadad:
@mathieuboudreau is this issue fixed with the recently-trained model (without the manual flipping hack)?

mathieuboudreau:
Resolved in #294.

PR #294, opened 2020-04-15
Resolve bug that makes TEM data not segment well (#293 and #249)

mathieuboudreau:
This resolves the bug in #293 by also resolving issue #249, since the trained TEM masks are no longer flipped prior to training.

Please test the following when reviewing:

* Pull this branch
* Download this image:

foo

* Rename it to image.png
* Segment using the command line: `axondeepseg -t TEM -i image.png -s 0.00236`
* Verify that the output image_seg-axonmyelin.png is well segmented.


vs74:
It's true that I didn't flip the TEM images during training that would the probable source of the issue. I will review this PR. Thanks for resolving this issue.

vs74:
So, I tested this PR and in the master, there was segmentation problem because of TEM contrast image flipped, whereas when I followed the instruction as described by @mathieuboudreau  in this PR, it successfully resolved the issue of TEM segmentation.

Here is the output
3wr27u


Since it is very urgent to resolve this issue, I am merging this PR to master. If any problem in future occurs related to this PR, the issue #293 can be reopened for discussion.

mathieuboudreau:
Thank you @vs74 ! Yes, you/we were never supposed to flip the TEM masks prior to training. That piece of code was in there because the masks were originally flipped prior to training for the paper, since I think the authors were exploring ways to use the same model to train different contrasts (since SEM and TEM have opposite contrasts). We didn't need that piece of code anymore, so I'm glad it's gone now and the issue is resolved.

PR #301, opened 2020-06-03
Resume training from checkpoint

### LINKED ISSUE #292: How to use a model I've trained on another system.
I have been training a model at home and I would like to use the model to segment images on a computer at my lab? How would I proceed doing that? Also, if I want to train 300 epochs but don't want to leave my computer on for 50 hours straight can I just train 50 epochs every night? Would it be the same or does it make it different?

mariehbourget:
This PR aims to clarify the procedure to resume training from a model checkpoint and integrate it to the 02-training-guideline notebook by adding a checkpoint parameter in the config file.

Fixes #292


mariehbourget:
@vs74  I've added 2 parameters "checkpoint" and "checkpoint_period" in the config file and modified the train_network.py accordingly.
Let me know if this is what your had in mind.

vs74:
@mariehbourget, yes I had this in mind. However, from Travis Build, I can see the tests are failing. I probable reason is that the test files also contain `config` file which you might not have altered. Fixing it would resolve the Travis build issue.

mariehbourget:
Thanks @vs74 , I've corrected test_train_network and the tests are ok now.

I have a question though.
After resuming training, the following files are overwritten:
`model.ckpt.data-00000-of-00001`
`model.ckpt.index`
`model.ckpt.meta`
`model.hdf5`
is it a problem? Would it be pertinent to keep those files between trainings?

Another potential problem for reproductibility is the `config_network.json file`.
The parameters `epoch` and `checkpoint` (and possibly `checkpoint_period`) change when resuming the training, but the parameters in config_network files won't be reflecting those changes.
Do you have an idea on how we could manage that?

vs74:
As per the training files is concerned, it's fine. Because after every regular interval of **checkpoint_period**, the best model is saved in comparison to the previous best model based on the training `accuracy` and `loss` to minimize during the training process.

vs74:
If I understand your second question correctly, is it related to manually coding the number of epochs each time after `checkpoints_period` during the training loop?

mariehbourget:
Yes. In the notebook, the user must change `epoch` and `checkpoint` manually in the config before resuming training. So the inital config_network file saved with the model won't track those changes. It is only an issue of reproductibility, if someone wants to redo the training, the informations about the whole process won't be available.

vs74:
Indeed the user needs to change the number of `epochs`  if the user is performing training using checkpoint, that's why I don't recommend training through this method. Actually there is a problem with Keras framework, it doesn't resume training with the previous epoch of the checkpoint model ( it usually starts training next checkpoint with `epoch 0 ` and not from the last epoch it was left at, hence in issue #292, I provided an example where after every checkpoint period the user needs to change the epochs). 

To counter this issue, I have one solution as of now. 

 https://github.com/dorukkarinca/keras-buoy - a library which we can use to tackle this problem, however, I have not tested out if it works.

mariehbourget:
I discussed with @jcohenadad and we will review the PR as is for now.

DONE
- Add new checkpoint and checkpoint_period parameters in config to resume training with best_as described [here](https://github.com/neuropoly/axondeepseg/issues/292#issuecomment-628911240)
- Include instructions for using checkpoints in [notebook](https://github.com/neuropoly/axondeepseg/blob/dev_issue292/notebooks/02-training_guideline.ipynb)

NOT IMPLEMENTED
- Keras ModelCheckpoint callbacks resume training at epoch #0 instead of the previous epoch of the model checkpoint (the user must change the number of epoch manually between trainings)

jcohenadad:
i will let @vs74 review this one

vs74:
Will review ASAP.

AxonDeepSeg/train_network.py
```diff
@@ -66,8 +66,16 @@ def train_model(
     epochs = config["epochs"]
     image_size = config["trainingset_patchsize"]
     thresh_indices = config["thresholds"]
-    checkpoint = config["checkpoint"]
-    checkpoint_period = config["checkpoint_period"]
+    if "checkpoint" in config:
+        checkpoint = config["checkpoint"]
+    else:
+        # For retrocompatibility with old configs
+        checkpoint = None
```
vs74, line 73: Do we require hardcoding of the checkpoint and checkpoint period as they have been defined in the config file already?

mariehbourget: I harcoded `checkpoint = None` and `checkpoint_period = 5` in case someone want to retrain with an old config where they are not defined. For example, in the config files saved with the default models on OSF.

vs74: Since you hardcoded already in the config file, that wouldn't change the results. Also, for the code consistency and easy maintainability in the future, I would have removed from the script and kept the parameter values ( as you have already done).

mariehbourget: Sorry for the delay. I'm not sure I understand your comment. I agree with you that for new config files created with the notebook, the `checkpoint` and `checkpoint_period `will be present and with values `None` and `5`. However, on OSF, the config files saved with our default models don't have those parameters. If we don't hardcode them here, the training will failed if the user load the old config files without adding the new parameters.

vs74: So, shouldn't we update the model's readme file on the OSF storage to comply with the changes?

mariehbourget: Yes, that would work to. I honestly don't know what's best. @mathieuboudreau can you take a look? (we can talk about that PR in the meeting tomorrow for some context)

mathieuboudreau: @mariehbourget I think we might have covered this at the last meeting, but in case we didn't, here's my thought. Instead of doing patchwork if/elses every time we add a new parameter in the config to preserve retrocompatibility, I think the better strategy would be to write a function that checks the config file as soon as we load it, and stop with an exception if we load a config file that's missing a parameter. Or, possibly even better, we could se the value there and throw a warning to the user that the value was missing, and here's the default value that we set. It would also help centralise the config changes in the future. I think we already have somewhat of a function that checks configs written, but I think it doesn't work perfectly (or as we really need it) so we could maybe build off of that?

I wouldn't edit the README of the models on OSF every time we change the software (e.g. new configs), that'll just become messy and we'll likely forget to do it. Just warn the user when needed from the software side is my view.

vs74: Thanks @mathieuboudreau  for your valuable suggestion. @mariehbourget , let me know what do you think about it and if in agreement whether you plan to include the function in this PR or would you like to open another PR to resolve this issue, in that case, I will approve this PR.

mariehbourget: Thanks @vs74 and @mathieuboudreau for your suggestions. As discussed in group meeting, we'll keep the hardcoding for retrocompatibility, as we don't want to do major refactoring at the moment. If we go ahead with the function, it will be in a future PR.

PR #305, opened 2020-06-10
Use a specific version of FSLeyes for the plugin

### LINKED ISSUE #303: Set a specific version of FSLeyes to use for the plugin.
There are a few problems with the most recent versions of FSLeyes
- FSLeyes 0.33.0 only supports python 3.7 and above.
- Mathieu reported that the plugin was not showing up on the menu-bar on FSLeyes 0.32.3 (although this wasn't the case for me)

Also, the installation of FSLeyes on Linux doesn't work with pip anymore. It must be installed with conda-forge. The documentation must then be updated.

I'd suggest using FSLeyes 0.30.1

Stoyan-I-A:
Resolves #303


Stoyan-I-A:
I have set the FSLeyes version to 0.30.1. However, if we now use Python 3.7, we could test if the latest version of FSLeyes works

Stoyan-I-A:
I'll test the Linux installation with FSLeyes 0.33.1 right now

Stoyan-I-A:
Tested the Linux installation. It works 👍

PR #307, opened 2020-06-12
Upgrade to Python 3.7

### LINKED ISSUE #306: Bump to Python 3.7
Motivation: https://github.com/neuropoly/axondeepseg/issues/303

mariehbourget:
Fixes #306


setup.py
```diff
@@ -31,7 +31,7 @@ def run(self):
 
 setup(
     name='AxonDeepSeg',
-    python_requires='>=3.6, <3.7',
```
jcohenadad, line 34: Why not simply =3.7?

Stoyan-I-A: I believe it's to allow Python versions from 3.7.0 to 3.7.7

jcohenadad: ah? i thought 3.7 do include all these? (ie when you install python, it fetches the latest patch)

mariehbourget: good point, thanks, i'll check as soon as i can (my workstation is down at the moment)

mariehbourget: The interval is required to allow versions from 3.7.0 to 3.7.7.
'3.7' and '=3.7' are invalid version specifier and '==3.7' returns `ERROR: Package 'AxonDeepSeg' requires a different Python: 3.7.7 not in '==3.7'`

jcohenadad: so i guess the solution is `>=3.7` as discussed here https://github.com/neuropoly/axondeepseg/issues/306

mariehbourget:
The travis tests won't run on my last commit. In Travis, I can see that the tests passed [here](https://travis-ci.org/github/neuropoly/axondeepseg/builds/699719374) and [here](https://travis-ci.org/github/neuropoly/axondeepseg/builds/699719367). but it doesn't appear in the PR. It seems to be a problem with the settings or the branch (as explain [here](https://github.community/t/expected-waiting-for-status-to-be-reported/18001)), but i don't have permissions to access the settings tab and fix it.

jcohenadad:
> The travis tests won't run on my last commit. In Travis, I can see that the tests passed [here](https://travis-ci.org/github/neuropoly/axondeepseg/builds/699719374) and [here](https://travis-ci.org/github/neuropoly/axondeepseg/builds/699719367). but it doesn't appear in the PR. It seems to be a problem with the settings or the branch (as explain [here](https://github.community/t/expected-waiting-for-status-to-be-reported/18001)), but i don't have permissions to access the settings tab and fix it.

it seems to be a bug, also see here: https://github.com/neuropoly/spinalcordtoolbox/pull/2739
try to re-commit something?

mariehbourget:
Re-commit did the trick, thanks!

setup.py
mariehbourget: Some dependencies are not compatible with 3.8, so we keep the interval `>=3.7, <3.8`.

mariehbourget:
This PR is ready for re-review with the addition of ADS version change to 3.2.

PR #320, opened 2020-06-26
[Bug] Resolve segment folder bug when using relative paths

### LINKED ISSUE #318: Error with segment_folder function in segment.py
I found two problems while trying to segment multiple images in the same folder.

1) In the documentation, there is a space in the filename that shouldn't be there (`image 1_sem/`) [here](fileName). 

2) Even though the test for segment_folder is passing, the command in the documentation doesn't work. There is a problem with the path folder being duplicated, so the images are not found.

I'm going to open a PR to fix it and see what happen with the test after.

mathieuboudreau:
Resolves #318

### DIFF
--- a/docs/source/documentation.rst
+++ b/docs/source/documentation.rst
@@ -298,8 +298,8 @@ The script to launch is called **axondeepseg**. It takes several arguments:
 **Optional arguments:**
 
 -m MODEL            Folder where the model is located. 
-                    The default SEM model path is **default_SEM_model_v1**. 
-                    The default TEM model path is **default_TEM_model_v1**.
+                    The default SEM model path is **default_SEM_model**. 
+                    The default TEM model path is **default_TEM_model**.
 
 -s SIZEPIXEL        Pixel size of the image(s) to segment, in micrometers. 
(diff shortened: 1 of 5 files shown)


docs/source/documentation.rst
```diff
@@ -324,9 +324,9 @@ To segment the same image by using the **'pixel_size_in_micrometer.txt'** file i
 Segment multiple images of the same resolution
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To segment multiple microscopy images of the same resolution that are located in the same folder, specify the path to the folder in the **-i** argument. For instance, to segment the images in folder **'test_sem_image/image 1_sem/'** of the test dataset that have a pixel size of 0.07 micrometers, use the following command::
```
mariehbourget, line 339: There are some`_v1` left in the models in doc on lines 289-290. We could use this PR to correct the names.

mathieuboudreau: @mariehbourget my laptop battery just died 😅, if you can please pull my branch, make the change, and push it. Thanks!

mathieuboudreau: (or make the change through the github interface)

mathieuboudreau:
@mariehbourget can you please re-approve? Your old review was dismissed since we pushed more changes.

Issue #286, opened 2020-03-13
Remove hardcoded paths in download_data()


mathieuboudreau:
Introduced when merging #226 

https://github.com/neuropoly/axondeepseg/blob/dae4a1dcd90539cfcc7347d38cb8264395de8a66/AxonDeepSeg/models/download_model.py#L12-L13

Should probably used the Path() function as well as relative paths to the main repo location.

mariehbourget:
I'm closing this issue because it is resolved with PR #322

PR #322, opened 2020-06-27
Fix redownloading models bug

### LINKED ISSUE #316: Error with downloading default models
This error occurs when trying to install ADS multiple times in a virtual environment.
It was previously discussed in issue [303](https://github.com/neuropoly/axondeepseg/issues/303#issuecomment-648300546), but I wanted to add some precision about the behaviour.

When installing ADS with `pip install -e .` :

- If a default model is already present in `AxonDeepSeg/models/default_SEM_model`, the installation works but the new downloaded model appears in `AxonDeepSeg/models/default_SEM_model/default_SEM_model`
- If `AxonDeepSeg/models/default_SEM_model/default_SEM_model` is already present, the installation returns the error `Error: Destination path 'AxonDeepSeg/models/default_SEM_model/default_SEM_model' already exists`.

The workaround for now is to delete the default models before re-installing.

mathieuboudreau:
Resolves #316 and #286

* If model folders already exists, deletes them before moving downloaded ones there
* Adds optional destination input to download_models
* Removed unused __main__ and renamed download_models function.
* Wrote tests for download_models.py (both basic and redownloading behaviour)
* Cleaned up a file that had an unused `pass` call.

### DIFF
--- a/AxonDeepSeg/models/download_model.py
+++ b/AxonDeepSeg/models/download_model.py
@@ -1,8 +1,17 @@
-from AxonDeepSeg.ads_utils import download_data
+from AxonDeepSeg.ads_utils import *
+from pathlib import Path
 import shutil
 
 
-def main():
+def download_model(destination = None):
+    if destination is None:
+        sem_destination = Path("AxonDeepSeg/models/default_SEM_model")
+        tem_destination = Path("AxonDeepSeg/models/default_TEM_model")
+    else:
+        destination = convert_path(destination)
+        sem_destination = destination / "default_SEM_model"
+        tem_destination = destination / "default_TEM_model"
+
     url_TEM_model = "https://osf.io/2hcfv/?action=download&version=4"  # URL of TEM model hosted on OSF storage with the appropriate versioning on OSF
     url_SEM_model = "https://osf.io/sv7u2/?action=download&version=4"  # URL of SEM model hosted on OSF storage with the appropriate versioning on OSF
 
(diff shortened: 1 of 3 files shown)


AxonDeepSeg/models/download_model.py
```diff
@@ -14,10 +23,17 @@ def main():
         print(
             "ERROR: Data was not succesfully downloaded and unzipped- please check your link and filename and try again."
         )
-    shutil.move("default_SEM_model", "AxonDeepSeg/models/default_SEM_model")
-    shutil.move("default_TEM_model", "AxonDeepSeg/models/default_TEM_model")
+
+    if sem_destination.exists():
+        print('SEM model folder already existed - deleting old one.')
```
mariehbourget, line 28: The print appears when using `download_models` from `setup.py` directly in the terminal but not while running `pip install -e .`

AxonDeepSeg/models/download_model.py
```diff
@@ -1,8 +1,17 @@
...
 
 
-def main():
+def download_model(destination = None):
+    if destination is None:
+        sem_destination = Path("AxonDeepSeg/models/default_SEM_model")
+        tem_destination = Path("AxonDeepSeg/models/default_TEM_model")
+    else:
+        destination = convert_path(destination)
+        sem_destination = destination / "default_SEM_model"
```
mariehbourget, line 12: I was unable to test the optional destination input. I don't think the path can be passed in arguments to `download_models` in command line from the entry point in `setup.py`, only if the function download_model() is called directly (like in the tests).

mariehbourget:
Also, I think that this PR will resolve issue [286](https://github.com/neuropoly/axondeepseg/issues/286).

AxonDeepSeg/models/download_model.py
mathieuboudreau: That's normal - we can't print anything when running setup.py (which is what `pip install -e .` does). There's no way around this: https://github.com/pypa/pip/issues/2933#issuecomment-290407667

AxonDeepSeg/models/download_model.py
mathieuboudreau: @mariehbourget yes you're right, I haven't implemented that feature for the CLI tool - only the within python. I needed to do this to be able to test it separately from the mode installed during download. I could implement it for the CLI now, or to merge this issue quicker we could open a new issue for that new feature.

mariehbourget: Ok, I suggest we merge now and open a new issue. I'll approve the PR.

PR #325, opened 2020-07-01
Add resampled_resolutions parameter in getting_started notebook

### LINKED ISSUE #324: Error with TEM segmentation in getting_started notebook
TEM segmentation fails in getting_started notebook because of the size of the patches.

The reason is that the parameter `resampled_resolutions` argument is not passed to the `axon_segmentation` function in section 2.2. `axon-segmentation` then used the default `resampled_resolutions=0.1` (corresponding to SEM).
With our TEM images, the resulting resampled image is smaller than a patch of 512x512.

We should add the argument in the function as well as a description for the user.

mariehbourget:
Fixes #324 

- Added `resampled_resolutions` parameter in `axon_segmentation` function in section 2.2 of `00-getting_started.ipynb`.
- Added description of the parameter for the user.

### DIFF
--- a/notebooks/00-getting_started.ipynb
+++ b/notebooks/00-getting_started.ipynb
@@ -189,7 +189,7 @@
    "source": [
     "**2.2. Launch the segmentation of the image.**\n",
     "\n",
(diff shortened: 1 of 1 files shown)


mathieuboudreau:
@mariehbourget I've been trying to review this PR for the last hour, however I've encountered a bad Jupyter Notebook bug (https://github.com/jupyter/notebook/issues/2563). I'm certain it has nothing to do with this PR, however I can't test the notebook as Jupyter isn't seeing my venv python installation. I'll keep you up to date, hopefully I can resolve this issue soon and continue reviewing.

PR #332, opened 2020-07-20
Remove gaussian blur option for data augmentation

### LINKED ISSUE #288: GaussianBlur in data augmentation
GaussianBlur doesn’t have the option for nearest neighbour interpolation in the release 3.0.
As discussed:
- this may cause problems with mask having multiple classes after augmentation,
- we could do our own nearest neighbour by thresholding the results of the augmentation,
- do we really want to have the GausianBlur option for data augmentation? ( see #21 )

mariehbourget:
Fixes #288

### DIFF
--- a/AxonDeepSeg/train_network.py
+++ b/AxonDeepSeg/train_network.py
@@ -89,6 +89,7 @@ def train_model(
     rotation = config["da-2-random_rotation-activate"]
     elastic = config["da-3-elastic-activate"]
     flipping = config["da-4-flipping-activate"]
+    gaussian_blur = False
     if "da-5-noise_addition-activate" in config:
         # Preserved for retrocompatibility with old configs
         gaussian_blur = config["da-5-noise_addition-activate"]
--- a/AxonDeepSeg/train_network.py
+++ b/AxonDeepSeg/train_network.py
@@ -89,9 +89,7 @@ def train_model(
     rotation = config["da-2-random_rotation-activate"]
     elastic = config["da-3-elastic-activate"]
     flipping = config["da-4-flipping-activate"]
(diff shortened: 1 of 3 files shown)


AxonDeepSeg/train_network.py
```diff
@@ -89,9 +89,8 @@ def train_model(
     rotation = config["da-2-random_rotation-activate"]
     elastic = config["da-3-elastic-activate"]
     flipping = config["da-4-flipping-activate"]
-    if "da-5-gaussian_blur-activate" in config:
-        gaussian_blur = config["da-5-gaussian_blur-activate"]
-    elif "da-5-noise_addition-activate" in config:
+    gaussian_blur = False
+    if "da-5-noise_addition-activate" in config:
         # Preserved for retrocompatibility with old configs
         gaussian_blur = config["da-5-noise_addition-activate"]
```
mathieuboudreau, line 95: If we're removing this functionality for a good reason (which you listed in issue #288), maybe it would be best to   remove the retrocompatibility here, and simply throw a warning that the old config had this setting on, and we don't recommended it so it's set to False. What do you think @mariehbourget ?

mariehbourget: Yes, I agree with you. I'm not sure if it is even necessary to throw a warning. I haven't found any example of config files with that key True (in the current models of previous releases models). Maybe it was used for development only? Let me know what would be the best practice here and I'll make the changes.

mariehbourget:
As discussed in group meeting, we'll keep the retrocompatibility. I put back a key that I had previously removed ("da-5-gaussian_blur-activate") in train_network.py as it make more sense to keep both options ("da-5-gaussian_blur-activate" and da-5-noise_addition-activate).

PR #342, opened 2020-08-19
Add an "Axon numbers" overlay in FSLeyes

### LINKED ISSUE #341: FSLeyes: Overlay axon numbers on the image
As suggested by one of our user: after computing morphometrics, it would be nice to be able to show the axons numbers (in the first column of the xls file) over the image in FSLeyes. It would be useful to identify rapidly the legitimate axon+myelin pairs when the model identifies small fragments outside of myelin as axons by mistake.

Stoyan-I-A:
TODO:
- fix #341
- Determine if the PIL.ImageFont.truetype() function works correctly on Mac.

DOC:
- Removed the indexed_gzip downgrade, as the issue where it would prevent NIfTI files from opening has been patched.


Stoyan-I-A:
@mathieuboudreau @mariehbourget If any of you two have some time, could you please try to run these lines on a Python console (or a Python script) in your ADS environment installed on a MacOS machine :
```
from PIL import Image, ImageDraw, ImageOps, ImageFont
font = ImageFont.truetype("arial.ttf", size=20)
```
If the second line returns an error, try to run this line insead:
`font = ImageFont.truetype("/usr/share/fonts/truetype/freefont/FreeMono.ttf", size=20, encoding="unic")
`

mathieuboudreau:
@Stoyan-I-A see below.

Capture d’écran 2020-08-28 à 16 41 06

mariehbourget:
Same for me. There is no `fonts` folder in `/usr/share/`.

Stoyan-I-A:
Thanks for the test.
A font is required to write the numbers on the image.

Do you know any directory where there are fonts on Mac? 
According to this [stackoverflow question](https://stackoverflow.com/questions/58968752/loading-fonts-in-python-pillow-on-a-mac), there are fonts in `/System/Library/Fonts`. Can anyone confirm this?

That being said, I'm not sure it's okay to hardcode a font path since that wont work with different operating systems.
Maybe we could add an existing package to `requirements.txt` that contains fonts?

jcohenadad:
> Maybe we could add an existing package to requirements.txt that contains fonts?

sounds safer than relying on user OS

Stoyan-I-A:
I found out that matplotlib comes with font files. I will use those!

mathieuboudreau:
@Stoyan-I-A is this branch ready for review?

Stoyan-I-A:
@mathieuboudreau Currently, the axon numbers overlay appears right after the user presses the `compute morphometrics` button and selects where to save the output file. 
If this is the way we want to keep this feature, then yes it is ready for review.

ads_plugin.py
```diff
@@ -156,7 +156,12 @@ def __init__(self, ortho, *args, **kwargs):
         compute_morphometrics_button = wx.Button(self, label="Compute morphometrics")
         compute_morphometrics_button.SetForegroundColour(button_label_color)
         compute_morphometrics_button.Bind(wx.EVT_BUTTON, self.on_compute_morphometrics_button)
-        compute_morphometrics_button.SetToolTip(wx.ToolTip("Calculates and saves the morphometrics to an excel and csv file.")) 
+        compute_morphometrics_button.SetToolTip(
+            wx.ToolTip(
+                "Calculates and saves the morphometrics to an excel and csv file. "
```
mariehbourget, line 161: Not directly related to this PR but I haven't find how to save the file in csv. The only option I have is xlsx.

mariehbourget (commented):
Tested and everything works well. My only concern is with very small axons on large images where the "numbers" layer gets pretty crowded. Example from SEM data14 (1901x1801)

Capture

Maybe we could use the mean diameter of the axons (in pixels) as an indicator to determine the font size (instead of image dimensions)? with the same minimum of 10 and a maximum to be determined.

jcohenadad:
> Maybe we could use the mean diameter of the axons (in pixels) as an indicator to determine the font size (instead of image dimensions)? with the same minimum of 10 and a maximum to be determined.

i would leave that for another PR

Stoyan-I-A:
> Maybe we could use the mean diameter of the axons (in pixels) as an indicator to determine the font size (instead of image dimensions)?

That's a good idea. There might be situations where this method is not optimal, but I can implement it and check if it's generally better.

docs/source/documentation.rst
```diff
@@ -192,10 +192,6 @@ Install FSLeyes using conda-forge ::
 
            yes | conda install -c conda-forge fsleyes=0.33.1
 
-Downgrade from latest version of indexed_gzip to the most recent working version ::
-
```
mathieuboudreau, line 196: Thanks for updating this!

mathieuboudreau (approved):
Reviewed the code - all looks good to me.

Tested the feature, it works as expected!

Capture d’écran 2020-09-09 à 21 23 23

Unexpected bonus of this feature: I was able to easily visually identify an area that would need too be manually corrected by where one of the numbers showed up (for the TEM dataset, it was axon number 41)!

Thanks a lot @Stoyan-I-A - excellent work!

PR #352, opened 2020-09-18
Release version 3.2.0

jcohenadad:
- Downgraded __ini__ from 3.2 to 3.1 because there was never a 3.1 version

TODO

- Add tags for each PR of 3.1
- Update CHANGELOG in preparation for 3.1
- Update python version requirement on RTD
- Change file name to stick to convention: CHANGELOG --> CHANGES

PR #364, opened 2020-10-14
Add actions for GitHub releases, pushing to pypi, and pypi_test tags

### LINKED ISSUE #353: Automatize PyPI package creation with github action
Automatize pypi package: example here: https://github.com/ivadomed/ivadomed/actions/runs/267724393/workflow

mathieuboudreau:
Resolves #353 
I've created GitHub actions and added relevant tokens for the repository. As I understand it, 

* release.yaml will create GitHub release for pushes/prs matching the patter "v*" (e.g. v3.2.0. This means that #352 would need to be renamed to `v3.2.0`)
* pypi_release.yaml will package and push to PyPi whenever a GitHub release is created (should be initiated after the release.yaml is completed).
* test_pypi_release.yaml was an optional one that I included/created, and allows you to test upload a package on test_pypi 
by created a push even matching the pattern "vt*". 

This is my first time creating GitHub actions, and in particular this workflow, so @jcohenadad maybe it would be a good idea if someone with experience in this matter could also review this PR, maybe someone working on another of your projects like SCT or IVADOMed.


mathieuboudreau:
Testing the release.yaml in a fork, it seems that it doesn't create a release the way I thought it should. Neither pushing a PR with `v.3.1.9` or pushing a tag with that same name results in a GitHub action being initiated. Looking into it, but might need some clarification on how this works; the documentation was lacking for that github action page: https://github.com/marketplace/actions/create-a-release

.github/workflows/pipy_release_test.yaml
```diff
@@ -0,0 +1,33 @@
+# This workflows will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Publish Package
+
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - 'vt*' # Push events to matching vt*, i.e. vt1.0, vt20.15.10
```
jcohenadad, line 10: removed "t", right?
```suggestion
      - 'v*' # Push events to matching v*, e.g., v2.0.5
```

jcohenadad:
> Testing the release.yaml in a fork, it seems that it doesn't create a release the way I thought it should. Neither pushing a PR with v.3.1.9 or pushing a tag with that same name results in a GitHub action being initiated. Looking into it, but might need some clarification on how this works; the documentation was lacking for that github action page: https://github.com/marketplace/actions/create-a-release

given that the GITHUB_TOKEN is specific to the repos (not the fork), maybe that's why it didn't create it? Or maybe was it caused by the "vt*" tag?

jcohenadad:
p.s.: we need to make sure we don't hire someone with first name starting with a "v", otherwise we will end up with a lot of releases 🤣

mathieuboudreau:
> > Testing the release.yaml in a fork, it seems that it doesn't create a release the way I thought it should. Neither pushing a PR with v.3.1.9 or pushing a tag with that same name results in a GitHub action being initiated. Looking into it, but might need some clarification on how this works; the documentation was lacking for that github action page: https://github.com/marketplace/actions/create-a-release
> 
> given that the GITHUB_TOKEN is specific to the repos (not the fork), maybe that's why it didn't create it? Or maybe was it caused by the "vt*" tag?

Hmm good point. I'll create a new repo from scratch to test this action out on.

mathieuboudreau:
> p.s.: we need to make sure we don't hire someone with first name starting with a "v", otherwise we will end up with a lot of releases 🤣

I would hope it doesn't try to match authors names haha. Likely just the commit/pr? But yeah, full functionality is still not clear to me; GitHub action documentations are definitely lacking from my experience in exploring them...

mathieuboudreau:
Ok so I changed strategy and now am using a different package to do both the semantic versioning and pushing to pypi.

These actions are triggered as followed:

* Bug version number bump: Commit message starting with `fix:`
* Minor version number bump: Commit message starting with `feat:`
* Major version number bump: Commit message starting with `feat:` and then in the description paragraph, `BREAKING CHANGE:`.

All of these trigger builds and pushes to pypi.

Tested on this test repo: https://github.com/mathieuboudreau/semantic_test & https://pypi.org/project/semantic-test/#history

jcohenadad:
what's cool is that it automatically generates the assets (zip + whl).
how come v1.0.0 comes before v2.0.0? https://github.com/mathieuboudreau/semantic_test/releases

jcohenadad:
IIUC, this proposed workflow is not compatible with the workflow used in the lab to list changes pertaining to specific versions, which uses [label categories](https://www.neuro.polymtl.ca/open_source/contributing#pr_labels). So, let's say, if we want to list several bug fixes for one release, this will not be possible with the proposed workflow (or we would need to fix the release body manually afterwards--), right?

mathieuboudreau:
> what's cool is that it automatically generates the assets (zip + whl).
> how come v1.0.0 comes before v2.0.0? https://github.com/mathieuboudreau/semantic_test/releases

I was wondering that too. I created v.1.0.0 manually on the GitHub website, whereas all the other ones were automatically generated. The other difference I see is that the tag is 1.0.0 and not v1.0.0, whereas the automatically generated ones are tagged v2.0.0.

mathieuboudreau:
> IIUC, this proposed workflow is not compatible with the workflow used in the lab to list changes pertaining to specific versions, which uses [label categories](https://www.neuro.polymtl.ca/open_source/contributing#pr_labels). So, let's say, if we want to list several bug fixes for one release, this will not be possible with the proposed workflow (or we would need to fix the release body manually afterwards--), right?

If I understand correctly, yes. I'm not aware of a tool that would allow that with labels - if you know one we could look into it.

mathieuboudreau:
What could be done is continue to use label categories, but when you want to make a release, then use the `feat:` syntax in a PR. All of the commits between that release and the most recent one would be then included in that release, regardless of tags. But the version would only be incremented once per `feat:` occurence, not for each `LABEL:`.

mathieuboudreau:
The other option I think is that you can do `feat(LABEL):` for syntax.

jcohenadad:
why don't we simply create releases (and the tag) manually on the GH interface (under create release), and the release creation would push on PyPI via a GH actions. So we wouldn't need a "GH release" action. We use [that scheme for ivadomed](https://github.com/ivadomed/ivadomed/actions/runs/267724393/workflow). It is fully compatible with category labels, and offers more flexibility. 

also, i anticipate we will screw up commit messages at some point (l'erreur est humaine), which will require some annoying manual git rebase procedures.

mathieuboudreau:
We can totally do that, but the more manual we need to do these steps, the more discipline we have to have (something we've had a challenge in this project to keep up long term, i.e. across different generations of students coming in and out of the project). Automated versioning also removes the chance to manually type a something like `33.2.0` instead of `3.2.0`, which is an easier typo than accidentally writing `feat:` =)

mathieuboudreau:
Forgetting to type `feat:` is easily fixable by simply writing a new commit to do so.

mathieuboudreau:
Just give me the green light and I'll make the change similar to IVADOMED, we'll just have to be super strict in PRs about remembering to always remind folks to 1- manually change __init__.py, 2- update the CHANGELOG.md, and 3- create a new release manually on the GitHub interface after it's merged.

jcohenadad:
you made valid points (wrt. manual tasks), and we haven't explored that avenue yet, so why don't we try your suggestion, and see how it goes. If we are not happy about it we can also go back to plan B (ivadomed approach).

mathieuboudreau:
Sounds like a good plan to me! @mariehbourget ready for a review and/or test with version 3.2.0 when you are available.

mariehbourget:
Thanks @mathieuboudreau, I will be available to review it tomorrow.

mariehbourget:
I have a few questions to make sure that I understand correctly:
- A new release would be trigger by the `fix(LABEL):` or `feat(LABEL):` string at the beginning of the commit message?
- This new release would include all the previous commit between that release and the most recent one?
- Does it mean that the changelog will include a line for each commit or for each PR?
- Will the lines of the changelog be sorted in categories corresponding to LABEL?

mathieuboudreau:
> I have a few questions to make sure that I understand correctly:
> 
> * A new release would be trigger by the `fix(LABEL):` or `feat(LABEL):` string at the beginning of the commit message?
> * This new release would include all the previous commit between that release and the most recent one?
> * Does it mean that the changelog will include a line for each commit or for each PR?
> * Will the lines of the changelog be sorted in categories corresponding to LABEL?

I think yes for the first two, and I'm not sure for the last two (they might be possible). I suggest you play around with some of these options in my playground repository that's already set up for it (https://github.com/mathieuboudreau/semantic_test), I'll give you admin access now.

mariehbourget:
I tested it with our ADS workflow (branch --> PR --> squash and merge) with the `feat:` trigger in the commit message or the merge. It works well for the release on ADS and PyPI but the changelog only contains 1 line referring to the merge commit.

Drulex:
Hello @jcohenadad mentioned this PR during a meeting on another project (sct). I found it interesting and wanted to share a few thoughts:
- Depending on your workflow and how often you commit to `master` you may not want to publish to `pypi` on every push to `master` (this could be very frequent). 
   - This is appropriate if you have a `release` branch, merge your features in `release` and then merge `release` to `master` when you want to release. 
   - An alternative is to run the action on a push to `release` branch. In that case you would simply merge `master` into `release` any time you would want to release.
- Maybe it would be useful to use the "neuropoly" changelog generator (https://github.com/neuropoly/changelog). In that case we could fork this action and tweak it to our needs (and then re-use it across other python projects inside `neuropoly`). The alternative is to write the action step-by-step. I am guessing that the complicated part would be to compute the correct version based on the history, but if we want to control the version manually this becomes pretty simple.

Good initiative!

jcohenadad:
Thank you for your thoughts @Drulex. The use of a "buffer" `release` branch sounds like a good idea, but we would need to figure out how to plug our changelog script. A few thoughts:
- Currently, changelog works by looking at the GH milestone, which is assigned manually, so it would be great to get rid of that manual step. Ideally, if the PR commit message includes the magic word (ie: presence of "feat"/"fix" as trigger), the changelog script would look at the latest tag, and include *all* PRs after that tag. 
- If we go with a `release` branch, after triggering the GH action, my understanding is that the `__init__` and CHANGELOG will be update automatically, so we need to make sure to merge back those updated files to `master` afterwards (which is not a big deal, but just so we are aware of all the manual interventions to do during the release workflow).

mathieuboudreau:
Closed for being stale, decided on suspending this idea for the moment.

jcohenadad:
A suggestion would be to use the same procedure as in https://ivadomed.org

PR #370, opened 2020-10-23
Resolve bug that makes TEM data not segment well with multiple files

### LINKED ISSUE #369: TEM data not segmenting well with CLI for multiple images in a folder
In the CLI, when pointing on a folder with multiple images, the function `segment_folder` of `AxonDeepSeg/segment.py` is called. 
ex: `axondeepseg -t TEM -i path_image/`
The function contains an old code line to invert contrast for TEM images [here](https://github.com/neuropoly/axondeepseg/blob/d13c0b0a541f139690afdb93ccb38cf92db4682e/AxonDeepSeg/segment.py#L162).

We had this bug earlier and the same code line was removed in PR #294 for the function `segment_image` used with the CLI pointing on a single file.
ex: `axondeepseg -t TEM -i path_image/image.png`

mariehbourget:
Fixes #369 

TEM image contrasts are no longer inverted when using CLI pointing on a folder with multiple files.

### DIFF
--- a/AxonDeepSeg/segment.py
+++ b/AxonDeepSeg/segment.py
@@ -159,10 +159,7 @@ def segment_folders(path_testing_images_folder, path_model,
 
         img_name_original = file_.stem
 
-        if selected_model == "default_TEM_model":
-            ads.imwrite(fp,255-img, format='png')
-        else:
-            ads.imwrite(fp,img, format='png')
+        ads.imwrite(fp,img, format='png')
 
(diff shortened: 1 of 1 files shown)


mathieuboudreau (approved):
Same fix as we did in #294 for individual images, so fine with me! We should maybe look into reducing duplication in code to avoid these types of bugs after we've fixed it for a related feature (e.g. single image vs folders). I'll open an issue.

Issue #368, opened 2020-10-20
Issues with 3.2.0 PyPI release


mariehbourget:
I tested the new 3.2.0 release on PyPI. There are a few issues and we might want to revisit the idea of maintaining PyPI.

1. The `integrity_test` won’t pass because the data_test are now in each default model folder, see [here](https://github.com/neuropoly/axondeepseg/blob/d13c0b0a541f139690afdb93ccb38cf92db4682e/AxonDeepSeg/integrity_test.py#L27).
The default models were included in the release but not the folder data_test inside of them. From my understanding, we should add `models/default_SEM_model/data_test/*` and `models/default_TEM_model/data_test/*` [here](https://github.com/neuropoly/axondeepseg/blob/d13c0b0a541f139690afdb93ccb38cf92db4682e/setup.py#L53) in setup.py, and remove `data_test/*` which no longer exists.

2. Since the default models folders are not in the master branch but are downloaded upon installation, I’m not sure if the release via github actions in PR #364 will include them. From my understanding, it is only a checkout of the master branch.

3. FSLeyes cannot be used because the top level of the package is `AxonDeepSeg` and the `ads_plugin.py` file is in the parent folder` axondeepseg`. Also, the file `logo_ads-alpha_small.png` used in the ADS control panel in FSLeyes is in `AxonDeepSeg` but not included in the package.

4. Jupyter notebook: same as above, the notebooks folder is outside of `AxonDeepSeg`.

The CLI commands are working as expected.

Related discussion in #109

mathieuboudreau:
Yes, readons 3 and 4 I was aware of, and were some of the main reasons why we gave up on pipy reases previously. We'd need to restructure the project (e.g. host notebooks in a different repository, same with the plugin) and provid additional instructions in order to make thus feasible. Or, we could just not release to pipy (don't think we have a wide audience on there) and just make people download the project on GitHub.

jcohenadad:
it's funny, we just had a long discussion in an SCT internal meeting about a related issue https://github.com/neuropoly/spinalcordtoolbox/issues/2959. The strategy we will be testing is to package external deps as wheels available on PyPI (and those wheels can be automatically pushed from separate GH repos). That will require a bit of restructuring, but it might be a strategical approach for the future (ie: we don't need to maintain our own package manager as we [currently do](https://github.com/neuropoly/axondeepseg/blob/master/AxonDeepSeg/models/download_model.py))

[...]

PR #376, opened 2020-10-30
Remove PyPI install instructions from RTD

mariehbourget:
Related to #368 
Removed the instructions for the installation of the stable release via PyPI in RTD.
The only installation methods are now the development version via git clone, or via a link to download a zip from github.

### DIFF
--- a/docs/source/documentation.rst
+++ b/docs/source/documentation.rst
@@ -68,16 +68,15 @@ installer compatible with your Windows system (64 bit recommended). After the do
 downloaded file, and follow the instructions. If you are unsure about any of the
 installation options, we recommend you use the default settings.
 
-Git (optional)
---------------
-Git is a software version control system. Because AxonDeepSeg is hosted on GitHub, a 
-service that hosts Git repositories, having Git installed on your system allows you
+git
+---
+``git`` is a software version control system. Because AxonDeepSeg is hosted on GitHub, a 
+service that hosts ``git`` repositories, having ``git`` installed on your system allows you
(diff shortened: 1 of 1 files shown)


jcohenadad (approved):
i've made slight modifications-- the rest looks ok 👍

mariehbourget:
If git is no longer "optional", we should also remove the link for download and the note about the main folder having a different name than "axondeepseg" i.e. the only option will now be to clone via git.
@jcohenadad, can you confirm that this is what you want and I'll make the changes, thanks!

jcohenadad:
> If git is no longer "optional", we should also remove the link for download and the note about the main folder having a different name than "axondeepseg" i.e. the only option will now be to clone via git.
> @jcohenadad, can you confirm that this is what you want and I'll make the changes, thanks!

makes sense to me-- it also simplifies the install procedure (the less text there is, the happier the users).  @mathieuboudreau are you ok with that?

mariehbourget:
> > If git is no longer "optional", we should also remove the link for download and the note about the main folder having a different name than "axondeepseg" i.e. the only option will now be to clone via git.
> > @jcohenadad, can you confirm that this is what you want and I'll make the changes, thanks!
> 
> makes sense to me-- it also simplifies the install procedure (the less text there is, the happier the users). @mathieuboudreau are you ok with that?

@mathieuboudreau, is it ok with you if we remove the link for download?

mathieuboudreau:
Sounds good to me !

mariehbourget:
This PR is ready for re-review.

jcohenadad (approved):

PR #382, opened 2020-11-02
Pin h5py~=2.10.0 as a workaround for upstream Keras/TensorFlow issues

### LINKED ISSUE #380: test_train_network.py failing on Travis CI
The builds (#379, #376) are failing for the recent PR's, as the test `(test_train_network.py`) is failing on Travis.
Here is the error for the failed test : - 
```
=================================== FAILURES ===================================
_______ TestCore.test_train_model_runs_successfully_for_simplified_case ________
self = <test_train_network.TestCore object at 0x7f82f4afb790>
    @pytest.mark.integration
    def test_train_model_runs_successfully_for_simplified_case(self):
        # Note: This test is simply a mock test to ensure that the pipeline
        # runs successfully, and is not a test of the quality of the model
        # itself.
    
        train_model(
            str(self.trainingPath),
            str(self.modelPath),
            self.config_network,
>           debug_mode=True
            )
test/test_train_network.py:135: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
AxonDeepSeg/train_network.py:277: in train_model
    model.save(str(path_model) + "/model.hdf5")
../../../miniconda/envs/ads_venv/lib/python3.7/site-packages/keras/engine/network.py:1090: in save
    save_model(self, filepath, overwrite, include_optimizer)
../../../miniconda/envs/ads_venv/lib/python3.7/site-packages/keras/engine/saving.py:382: in save_model
    _serialize_model(model, f, include_optimizer)
../../../miniconda/envs/ads_venv/lib/python3.7/site-packages/keras/engine/saving.py:114: in _serialize_model
    layer_group[name] = val
../../../miniconda/envs/ads_venv/lib/python3.7/site-packages/keras/utils/io_utils.py:218: in __setitem__
    dataset = self.data.create_dataset(attr, val.shape, dtype=val.dtype)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <Closed HDF5 group>, name = b'cconv-d0-c0/convolution/conv2d_1/kernel:0'
shape = (3, 3, 1, 5), dtype = dtype('float32'), data = None
kwds = {'track_order': False}, group = <Closed HDF5 group>
    def create_dataset(self, name, shape=None, dtype=None, data=None, **kwds):
        """ Create a new HDF5 dataset
    
        name
            Name of the dataset (absolute or relative).  Provide None to make
            an anonymous dataset.
        shape
            Dataset shape.  Use "()" for scalar datasets.  Required if "data"
            isn't provided.
        dtype
            Numpy dtype or string.  If omitted, dtype('f') will be used.
            Required if "data" isn't provided; otherwise, overrides data
            array's dtype.
        data
            Provide data to initialize the dataset.  If used, you can omit
            shape and dtype arguments.
    
        Keyword-only arguments:
    
        chunks
            (Tuple or int) Chunk shape, or True to enable auto-chunking. Integers can
            be used for 1D shape.
    
        maxshape
            (Tuple or int) Make the dataset resizable up to this shape. Use None for
            axes you want to be unlimited. Integers can be used for 1D shape.
        compression
            (String or int) Compression strategy.  Legal values are 'gzip',
            'szip', 'lzf'.  If an integer in range(10), this indicates gzip
            compression level. Otherwise, an integer indicates the number of a
            dynamically loaded compression filter.
        compression_opts
            Compression settings.  This is an integer for gzip, 2-tuple for
            szip, etc. If specifying a dynamically loaded compression filter
            number, this must be a tuple of values.
        scaleoffset
            (Integer) Enable scale/offset filter for (usually) lossy
            compression of integer or floating-point data. For integer
            data, the value of scaleoffset is the number of bits to
            retain (pass 0 to let HDF5 determine the minimum number of
            bits necessary for lossless compression). For floating point
            data, scaleoffset is the number of digits after the decimal
            place to retain; stored values thus have absolute error
            less than 0.5*10**(-scaleoffset).
        shuffle
            (T/F) Enable shuffle filter.
        fletcher32
            (T/F) Enable fletcher32 error detection. Not permitted in
            conjunction with the scale/offset filter.
        fillvalue
            (Scalar) Use this value for uninitialized parts of the dataset.
        track_times
            (T/F) Enable dataset creation timestamps.
        track_order
            (T/F) Track attribute creation order if True. If omitted use
            global default h5.get_config().track_order.
        external
            (Iterable of tuples) Sets the external storage property, thus
            designating that the dataset will be stored in one or more
            non-HDF5 files external to the HDF5 file.  Adds each tuple
            of (name, offset, size) to the dataset's list of external files.
            Each name must be a str, bytes, or os.PathLike; each offset and
            size, an integer.  If only a name is given instead of an iterable
            of tuples, it is equivalent to [(name, 0, h5py.h5f.UNLIMITED)].
        allow_unknown_filter
            (T/F) Do not check that the requested filter is available for use.
            This should only be used with ``write_direct_chunk``, where the caller
            compresses the data before handing it to h5py.
        """
        if 'track_order' not in kwds:
            kwds['track_order'] = h5.get_config().track_order
    
        with phil:
            group = self
            if name:
>               if '/' in name:
E               TypeError: a bytes-like object is required, not 'str'
../../../miniconda/envs/ads_venv/lib/python3.7/site-packages/h5py/_hl/group.py:143: TypeError
```

However, when I'm running the tests on my local machine, all tests, including the test (`test_train_network.py`), passes. I don't understand this disparity. Anyone has any clue?

On my local machine, I ran `pytest -v` and all the tests ran successfully. 

**System Requirements**
**OS:** macOS Mojave

mathieuboudreau:
Resolves #380 

Solution from https://github.com/neuropoly/spinalcordtoolbox/pull/2988

### DIFF
--- a/AxonDeepSeg/train_network.py
+++ b/AxonDeepSeg/train_network.py
@@ -274,7 +274,7 @@ def train_model(
 
     ########################## Save the model after Training ###########
 
-    model.save(str(path_model) + "/model.hdf5")
+    model.save(path_model / "model.hdf5")
 
     # Add ops to save and restore all the variables.
     saver = tf.train.Saver()
--- a/AxonDeepSeg/train_network.py
+++ b/AxonDeepSeg/train_network.py
@@ -274,7 +274,7 @@ def train_model(
 
     ########################## Save the model after Training ###########
 
-    model.save(path_model / "model.hdf5")
+    model.save(str(path_model / "model.hdf5"))
 
     # Add ops to save and restore all the variables.
     saver = tf.train.Saver()
--- a/AxonDeepSeg/train_network.py
+++ b/AxonDeepSeg/train_network.py
@@ -273,8 +273,11 @@ def train_model(
     )
 
     ########################## Save the model after Training ###########
-
-    model.save(str(path_model / "model.hdf5"))
+    model_output = str(path_model / "model.hdf5")
+    model_output2 = path_model / "model.hdf5"
+    print(model_output)
+    print(model_output2)
+    model.save(model_output)
 
     # Add ops to save and restore all the variables.
     saver = tf.train.Saver()
--- a/AxonDeepSeg/train_network.py
+++ b/AxonDeepSeg/train_network.py
@@ -273,11 +273,8 @@ def train_model(
     )
 
     ########################## Save the model after Training ###########
-    model_output = str(path_model / "model.hdf5")
-    model_output2 = path_model / "model.hdf5"
-    print(model_output)
-    print(model_output2)
-    model.save(model_output)
+
+    model.save(str(path_model / "model.hdf5"))
 
     # Add ops to save and restore all the variables.
     saver = tf.train.Saver()
--- a/test/test_train_network.py
+++ b/test/test_train_network.py
@@ -122,7 +122,7 @@ class TestCore(object):
                 print("Could not clean up {} - you may want to remove it manually.".format(modelPath))
 
     # --------------train_model tests-------------- #
-    @pytest.mark.debug
+    @pytest.mark.integration
     def test_train_model_runs_successfully_for_simplified_case(self):
         # Note: This test is simply a mock test to ensure that the pipeline
         # runs successfully, and is not a test of the quality of the model
--- a/test/test_train_network.py
+++ b/test/test_train_network.py
@@ -122,7 +122,7 @@ class TestCore(object):
                 print("Could not clean up {} - you may want to remove it manually.".format(modelPath))
 
     # --------------train_model tests-------------- #
-    @pytest.mark.integration
+    @pytest.mark.debug
     def test_train_model_runs_successfully_for_simplified_case(self):
         # Note: This test is simply a mock test to ensure that the pipeline
         # runs successfully, and is not a test of the quality of the model
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,9 +16,10 @@ pytest-cov==2.5.1
 prettytable==0.7.2
 raven==6.8.0
 jupyter>=1.0.0
-Keras==2.2.5
+Keras==2.2.4
 Keras-Applications==1.0.8
 Keras-Preprocessing==1.1.0
 opencv-contrib-python==4.1.0.25
 albumentations==0.3.0
 openpyxl==2.6.3
+h5py~=2.10.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,7 +16,7 @@ pytest-cov==2.5.1
 prettytable==0.7.2
 raven==6.8.0
 jupyter>=1.0.0
-Keras==2.2.4
+Keras==2.2.5
 Keras-Applications==1.0.8
 Keras-Preprocessing==1.1.0
 opencv-contrib-python==4.1.0.25


vs74:
@mathieuboudreau the build for the Mac operating system seems to have failed because of jobs have exceeded the minimum time for execution. However, for the linux operating system, the build passed. Also, for travis CI the time taken to run all the tests is too long ( around **40 min** for linux OS and around **50 min** for Mac OS, leading to a total run time of **1 hour 40 min**, which is too much). Is it normal ? I think, we need to find an efficient way to deal with such an issue.

jcohenadad:
> @mathieuboudreau the build for the Mac operating system seems to have failed because of jobs have exceeded the minimum time for execution. However, for the linux operating system, the build passed. Also, for travis CI the time taken to run all the tests is too long ( around **40 min** for linux OS and around **50 min** for Mac OS, leading to a total run time of **1 hour 40 min**, which is too much). Is it normal ? I think, we need to find an efficient way to deal with such an issue.

@vs74 could you hunt down where the bottleneck is?

vs74:
Sure, I'll look into it.

mathieuboudreau:
To me today it was just Travis acting funny (it ran 15 minutes before first starting to pip install our tools), maybe they're overloaded. I don't think it's the test themselves, which typically take 10-20 minutes to run

mariehbourget (commented):
I checked out the branch and installed ADS.
The `h5py` version installed is `2.10.0` as expected.
I ran the tests locally, all 101 tests passed.

However, as mentioned by @vs74, the tests on Travis timed out.

mathieuboudreau:
Maybe the Travis Linux builds are failing because of the large backlog that happened for those builds today:

Capture d’écran 2020-11-02 à 18 07 51

mathieuboudreau:
Oh nevermind, it's the Mac builds that timed out. I just restarted them to see if they'll run better now regardless.

mathieuboudreau:
The build issue is very likely to an update to opencv-python-headless made 6 hours ago

Capture d’écran 2020-11-02 à 19 17 15

The wheel build stalls locally for me. I'll try setting it to a previous version.

vs74:
I was monitoring the travis CI build for the before this [recent commit](https://github.com/neuropoly/axondeepseg/pull/382/commits/7a5969f9aaa284a2eaba1573a425b34b02b56d62), and saw that building wheel for `opencv-python-headless`was taking long enough. I hope this [recent commit](https://github.com/neuropoly/axondeepseg/pull/382/commits/7a5969f9aaa284a2eaba1573a425b34b02b56d62) would solve the issue.

test/test_train_network.py
```diff
@@ -122,7 +122,7 @@ def teardown_class(cls):
                 print("Could not clean up {} - you may want to remove it manually.".format(modelPath))
 
     # --------------train_model tests-------------- #
-    @pytest.mark.integration
+    @pytest.mark.debug
```
vs74, line 125: Do we need to change the marker of `test_train_network`.py from `integration` to `debug`?

AxonDeepSeg/train_network.py
```diff
@@ -274,7 +274,7 @@ def train_model(
 
     ########################## Save the model after Training ###########
 
-    model.save(str(path_model) + "/model.hdf5")
+    model.save(str(path_model / "model.hdf5"))
```
vs74, line 277: I presume the reason for this commit is being consistent in ADS code base on string concatenation. Is that right?

vs74 (commented):
Indeed, `opencv-python-headless` was the reason for the build to take too long. Thank you @mathieuboudreau for the quick turnaround in addressing this issue. I'll approve this PR once you address my comments.

test/test_train_network.py
mathieuboudreau: Yes i'll change it, thanks for catching it

AxonDeepSeg/train_network.py
mathieuboudreau: Yup, exactly. I noticed it because that's the line that the test was failing on.

vs74 (approved):

PR #386, opened 2020-11-06
Add zoomed window on masks examples in RTD

### LINKED ISSUE #354: Good mask example is resampled on RTD
We updated our documentation with an example of a good and bad masks for [manual labelling](https://axondeepseg.readthedocs.io/en/latest/documentation.html#guide-for-manual-labelling).

However, because of the resampling on RTD, the good mask example has intermediate values that are not 0, 127 or 255. So the example may not be as clear as we intended.

The image is 375x250 pixels, the resampled size on RTD is 750x500.
We should resize the masks in nearest neighbor to the real size on RTD.

mariehbourget:
Fixes #354 
Added zoomed window on good/bad mask examples to clarify instructions and avoid confusion with the resampling of images on RTD.

Doc rendered here: https://axondeepseg.readthedocs.io/en/mhb-354-fix-mask-doc/documentation.html#guide-for-manual-labelling


jcohenadad (approved):
*very* nice addition! thank you @mariehbourget 🙏  !!

Issue #396, opened 2020-11-16
Visually Distinguishing Adjacent Myelin Wraps and Ellipse Minor Access in Fsleyes


n-amb:
Hello, 

Within my image set, there are cases of adjacent myelin wraps contacting each other (see screen grab below). The morphometric calculation typically yields reasonable measurements in these cases. However, I think it would be useful to visually confirm that adjacent myelin wraps are in fact distinct within Fsleyes. Perhaps this means being able to select a specific ROI/axon-myelin instance and toggle them off and on, or that the myelin overlay created includes multiple colors that can be used to distinguish discrete myelin wraps. 
<img width="364" alt="Screen Shot 2020-11-15 at 8 49 29 PM" src="https://user-images.githubusercontent.com/71902782/99214046-1071ec80-2784-11eb-90f8-4d309b95bd96.png">

In addition, a previous post alluded to the implementation of the ellipse minor axis morphometric measurement to the Fsleyes plugin. Is this option available with the current version? 

Thanks for considering these feature requests!

Stoyan-I-A:
Coincidentally, the team discussed about these topics today.

For the visual distinction between two axon-myelin objects, we used to have a feature that did something like that a while ago. We removed it because its purpose wasn't clear at the time but now we would like to explore the possibility of adding such this feature to a new interactive user interface similar to this one: https://dash-gallery.plotly.host/dash-label-properties/ .
That being said, this won't be done on the short term.

As for the ellipse minor axis, we are working on it. You can follow the progress in #381 .

mathieuboudreau:
Stale

PR #397, opened 2020-11-16
Use the myelin centroids as seed points for the floodfill

### LINKED ISSUE #384: Improve procedure for manual correction of output segmentations
### Context

Currently, the [procedure for correcting output segmentations](https://axondeepseg.readthedocs.io/en/latest/documentation.html#manual-correction-of-segmentation-masks) is based on the filling of myelin masks, based on a watershed initialized by the center of mass (COM) of the axons.

This is prone to issues because sometimes, lots of "wrong" single axon pixels are spread out at many places in the output segmentation image.  

### Suggestion

A more efficient workflow would consist in *not* relying on the axon mask, and only on the myelin mask (less work to do):
- Manually fixing the myelin mask only (ie: closing it, adjusting borders, remove isolated pixels)
- Estimate COM of the myelin (not axons)
- Use that COM to run the watershed.

Stoyan-I-A:
Check this presentation for more information on the tool: https://docs.google.com/presentation/d/1aRl7C1IzzdhXzUBHk4tnIzU3L2r-DJ1540hnQdivzJE/edit#slide=id.p

fixes #384 

In FSLeyes, the axon mask is no longer required to use the "Fill axons" button

### DIFF
--- a/AxonDeepSeg/postprocessing.py
+++ b/AxonDeepSeg/postprocessing.py
@@ -26,17 +26,17 @@ def get_centroids(mask):
     )
     return ind_centroid
 
-def floodfill_axons(axon_array, myelin_array):
+def floodfill_axons(myelin_array):
     """
-    This function does a floodfill operation on the myelin_array. The seed points are the centroids of the axon objects
-    in the axon_array. The goal is to fill the center of the myelin objects with the axon mask.
+    This function does a floodfill operation on the myelin_array. The seed points are the centroids of the myelin
+    objects. The goal is to fill the center of the myelin objects with the axon mask.
     Note: The myelin objects need to be closed in order to prevent the axon_mask from flooding the entire image.
-    :param axon_array: the binary array corresponding to the axon mask
+    The myelin objects also need to be separated from each other, otherwise they will be counted as one object.
     :param myelin_array: the binary array corresponding to the myelin mask
     :return: the binary axon array corresponding to the axon mask after the floodfill
     """
     # Get the centroid indexes
-    centroid_index_map = get_centroids(axon_array)
+    centroid_index_map = get_centroids(myelin_array)
 
     # Create an image with the myelinMask and floodfill at the coordinates of the centroids
     # Note: The floodfill algorithm only works on RGB images. Thus, the mask must be colorized before applying
--- a/ads_plugin.py
+++ b/ads_plugin.py
@@ -35,7 +35,7 @@ import imageio
 
 from AxonDeepSeg.morphometrics.compute_morphometrics import *
 
-VERSION = "0.2.14"
+VERSION = "0.2.15"
 
 class ADScontrol(ctrlpanel.ControlPanel):
     """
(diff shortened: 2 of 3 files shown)


Stoyan-I-A:
@mariehbourget You can use this to fill the axons in the Simeon images. 
Remember,
- No open myelin
- No touching myelin

Stoyan-I-A:
By the way, in #396 , there's an exemple where this feature might not suit the user.

mariehbourget:
Thank you @Stoyan-I-A!

I tried the feature, it works well, but there are a few axons where the “Fill axons” fills the myelin itself instead of the interior part. For those axons, the center is black, and the myelin is both blue (in the axon mask) and red (in the myelin mask).
fsleyes_axon_example_50 fsleyes_myelin_example_50

When exporting, the overlap is detected so the “myelin” mask is fine, but those axons are missing on the “axon” and “axonmyelin” masks.

I’m not sure if I can share the myelin mask I used here, so I’ll send you a link privately if you want to test it on your side.

For the time being, I was able to correct those few axons manually.

As you mentioned, the “no touching myelin” might be a problem in several cases, with our datasets and other users datasets. In my opinion, the feature is worth pursuing because it simplify greatly the manual correction procedure, however, we should keep looking for solutions with touching myelin before changing the behavior.

Stoyan-I-A:
For the axon on the top left, this happens because the center of mass of the myelin object is actually on the myelin itself.
image
I am not sure how we should deal with those cases.

Stoyan-I-A:
Same deal with the other one
image

jcohenadad:
> For the axon on the top left, this happens because the center of mass of the myelin object is actually on the myelin itself.

Hum, that's an interesting problem. IIUC, you only need a single pixel _inside_ the myelin, to start the flood fill process, right? So, it doesn't necessarily need to be at the center. If that's the case, how about applying a closure of the myelin mask, followed by the subtraction between the closure and the myelin mask --> the output should _only_ give pixels inside the myelin (ie: which are axons). Then, pick one of those pixels as the seed for the flood fill. Does that make sense?

jcohenadad:
An additional comment about the problem of touching myelin: how about applying a [squeletonization](https://scikit-image.org/docs/dev/auto_examples/edges/plot_skeleton.html) on the myelin masks, followed by the finding of seeds that will be used by the flood fill algo (this seed finding could be center of mass, or another method as pointed out in https://github.com/neuropoly/axondeepseg/pull/397#issuecomment-730465543).

Stoyan-I-A:
> you only need a single pixel inside the myelin, to start the flood fill process, right?

Yes.

I have done a bit of experimentation with this myelin mask:
image

Here's what a closing operation does to the mask:
image

With the two images overlapped:
image

It is definitely something we could work with, but I felt like there were probably easier solutions. So I looked for alternatives and found the [remove_small_holes](https://scikit-image.org/docs/dev/api/skimage.morphology.html#remove-small-holes) function from scikit-image. With this, I can define how large the maximum area of an axon should be, and it will fill the image as long as the myelin is closed.

Here's what it looks like if I set the maximum area of one axon to about 10% of the total image size:
image

With the two images overlapped:
image

So far, this looks nice.

Stoyan-I-A:
I am closing this PR because we chose to use the remove_small_holes for filling the myelin mask with axon objects. The development of this tool is in #585.

Issue #349, opened 2020-09-10
Diameter measurements discussion


mariehbourget:
This issue is opened to discuss the different methods used for diameter measurements between ADS and other users protocols.

In ADS, the calculation of the diameter is based on the axon area of the mask. For each axons, the equivalent diameter is computed, which is the diameter of a circle with the same area as the axon. Similarly, ADS calculates the area of each axon+myelin pairs on the mask and its equivalent diameter. The myelin thickness is then calculated from the difference between the axon+myelin equivalent diameter and the axon equivalent diameter.

We recommend to use this method to ensure that a mean diameter is computed that is not bias in any specific direction, and also takes into account that axons do not have a perfectly cylindrical shape.

We received feedback from users with different protocols for manual measurements:
1. Axon and cell diameters are taken from the shortest distance passing through the center of the axon [here](https://github.com/neuropoly/axondeepseg/issues/347#issuecomment-689961633).
2. Another user mentioned by e-mail that they treat the axon as an ellipse and take the minor axis of the ellipse as the axon diameter to account for the sectioning for TEM that is never perfectly orthogonal to the axon resulting in the axons being usually in oblong shapes.

mathieuboudreau:
What is the current consensus on how it's currently (or used to be) done manually in the histology world?

Paasu:
The measurements I performed were done on TEM images of rat optic nerves using ImageJ. The method echoed what the other user tries; measuring the minor axis of the 'ellipse' mainly because it gets the minimum axon/fibre diameters. The minimum axon diameter was used to reduce the skewing of data accounted by the oblong-like shapes of some of the fibres. When it comes to injured nerves however, they tend to produced decompaction within the myelin while regenerating; myelin thickness is very different and g-ratio doesn't always reflect accurately the relationship between axon and myelin. Here are a few studies that used the methodology as described above when looking at rat neurons in TEM:

Payne et al: https://pubmed.ncbi.nlm.nih.gov/21381867/
 - Measured minimum axon and fibre diameter
 - Calculated myelin thickness by subtracting axon diameter from fibre diameter and dividing by 2
 - Minimum was calculated to avoid artificial skewing of data due to axons projecting at an oblique angle
 - Only myelin thickness of normally myelinated axons were calculated to reduce skew in data
 
Payne et al: https://pubmed.ncbi.nlm.nih.gov/22879411/
 - Measured minimum axon diameter
[...]

Issue #347, opened 2020-09-07
Inaccurate data extraction


Paasu:
**Describe the bug**
I tried extracting the morphometric data on the manual masks to test the efficacy of the data compared to my manual measurements. I do understand that it was the manual masks and not the segmented images so it might be the reason the data differs significantly. The reason I haven't used the masks to train and the use the segmented images to analyze is because of #346. 
[Neuron Morphometrics.pdf](https://github.com/neuropoly/axondeepseg/files/5183710/Neuron.Morphometrics.pdf)

I only used a very small sample of 4 masks but there seems to be a significant difference in g-ratio, and myelin thickness with smaller but present differences in axon diameter. Is it possible because my masks aren't high quality enough?
[Images.zip](https://github.com/neuropoly/axondeepseg/files/5183715/Images.zip)

Feel free to let me know if you guys need more information as I do understand it might be a little unclear.

**Desktop (please complete the following information):**
 - OS: [OpenSUSE]
 - Version [15.1]

jcohenadad:
@Paasu this is the names of your images:
![image](https://user-images.githubusercontent.com/2482071/92402198-2799bb80-f0fd-11ea-9ac1-2502ed072cd3.png)

and this is the content of the pdf:
![image](https://user-images.githubusercontent.com/2482071/92402224-2f596000-f0fd-11ea-9448-43e17306cc16.png)

it is impossible for us to know which sample number corresponds to which image.

so, what you need to do is: 
- send us: one manual mask
- send us: the exact syntax you typed to compute automatic morphometrics + results of automatic morphometrics
- send us your manual measures

then, we will be able to help you.

Paasu:
@jcohenadad 
Sorry about that. Here contains what you requested:
[data.zip](https://github.com/neuropoly/axondeepseg/files/5184017/data.zip)

jcohenadad:
the file `pixel_size_in_micrometer.txt` is missing. Also, do you have a screenshot or any recording of your manual measurements? e.g.: which axons did you measure and how did you measure them.

Paasu:
[pixel_size_in_micrometer.txt](https://github.com/neuropoly/axondeepseg/files/5184058/pixel_size_in_micrometer.txt)
Woops, I thought I included it.

jcohenadad:
[...]

PR #179, opened 2019-03-08
Add Axon mask simulator

mathieuboudreau:
Resolves #178 , and as mentioned in #176

Initial functionality:

```python

from AxonDeepSeg.visualization.simulate_axons import SimulateAxons
a=SimulateAxons()
a.generate_axon(50)
a.generate_axon(75, center=[100, 200])
a.generate_axon(100, center=[800, 150])
a.generate_axon(100, center=[600, 700], gratio=0.9)
a.save()

```

Output (`SimulatedAxons.png`):

simulatedaxons

No documentation yet, but the code should be self-evident on what functions do and how to use them.

PR #399, opened 2020-11-19
Implement Ellipse Minor Axis  as Diameter

### LINKED ISSUE #363: Use ellipse minor axis as diameter in morphometrics
Following discussions on issues #347  and #349 , we want to implement a new feature for computing morphometrics.
This feature should:
- Use the minor axis of a fitted ellipse as axon diameter and axon+myelin diameter.
- Give the choice to the user to compute morphometrics with the current method (equivalent diameter) or the new method (minor axis).
- Integrate the change in FSLeyes (should we open a separate issue?)

vs74:
Following the discussions in #363, #349, this PR aims to implement the diameter of axon using minor axis of ellipse.

**DONE:**

- [x] Implement minor axis as an additional feature to compute the diameter of an axon, thickness of myelin, diameter of axon_myelin
- [x] In order to allow the user to chose **minor axis** or **equivalent diameter** while computing morphometrics as its diameter, user can manually set the boolen variable `ellipse` to be `True` or `False`.
- [x] Made the necessary changes in `04-compute-morphometrics.ipynb` notebook file, allowing the user to set his choice for diameter computation.
- [x] Added comprehensive tests to test this new feature.
- [x] Implement the similar behaviour in **FSLeyes plugin**, where the user is prompted to set his preferred choice of diameter either as **equivalent diameter** or **ellipse minor axis**.  Opened seperate issue for this see #432  and it will be dealt in seperate PR. 
- [x] Add documentation for this feature in notebook [04-morphometrics_extraction.ipynb](https://github.com/neuropoly/axondeepseg/blob/master/notebooks/04-morphometrics_extraction.ipynb)
- [x] Add a flag to select the shape of the axons
- [x] Documentation: Add literature for axon shape (circle and ellipse)
- [x] Add cli tests for the axon shape `-a` flag

## What are the main contributions of this PR?
1. Implements Ellipse Minor axis as an additional way to compute morphometrics
2. For generating morphometrics via CLI, adds an flag `-a` to select the shape of axon. Usage: Refer to the [docs](https://axondeepseg--399.org.readthedocs.build/en/399/documentation.html#ellipse) for usage.
3. Updated the [RTD](https://axondeepseg--399.org.readthedocs.build/en/399/documentation.html#morphometrics-of-a-single-image)

**NOTE: -** The default behaviour is set as **equivalent diameter**(circle) for measuring morphometrics. However, if the user wants to consider the axon as an "oblong ellipse", the user can set `ellipse` boolean variable to `True`.

Fix #363, #349


vs74:
As discussed in the meeting, to validate this new feature,  I tested it on a good **TEM mask**, that is,  located in our default TEM model. Manual measurements were done using the **protocol** as mentioned in this [comment](https://github.com/neuropoly/axondeepseg/issues/347#issuecomment-689961633). **4** manual measurements were done so that we can compare it with our new feature ( `ellipse minor axis`) and current method of calculating diamter (`equivalent diameter`). 

Since this is my first time working with **Fiji**, I am including the mask on which I did manual measurements. So, if either of @mariehbourget  / @jcohenadad  could check the correctness of it superficially, that would be great. 

mask.


The instances which were manually measured are **15, 16, 33, 40**. 
Screenshot 2020-11-19 at 7 49 53 PM

vs74:
Here are the diameter measurements for 4 labels **[15, 16, 33, 40]** . 
**Manual Measurements** :  

  |   |   |  
-- | -- | -- | --
Instance | Axon Diameter | Myelin Thickness | Axon + Myelin Diameter
15 | 0.45361796 | 0.07566632 | 0.62155792
16 | 0.4111002 | 0.08243008 | 0.61844272
33 | 0.53261188 | 0.08286904 | 0.7177232
40 | 0.3493508 | 0.0708 | 0.51692732

**Ellipse Minor Axis**

  | axon_diam | myelin_thickness
-- | -- | --
15 | 0.46739465 | 0.085082158
16 | 0.443414718 | 0.084947787
33 | 0.54835999 | 0.093597934
40 | 0.364738882 | 0.079313725

**Equivalent Diameter**


  | axon_diam | myelin_thickness
-- | -- | --
15 | 0.566857219 | 0.092142902
16 | 0.565692604 | 0.087108396
33 | 0.634559572 | 0.091233045
40 | 0.400306314 | 0.090037793

Looking at the results, when **ellipse minor axis** is used to measure diameter, the results seems close to the manual measured values of the labels as compared to using **equivalent diameter**.

mariehbourget:
> Since this is my first time working with **Fiji**, I am including the mask on which I did manual measurements. So, if either of @mariehbourget / @jcohenadad could check the correctness of it superficially, that would be great.

@vs74, yes the measurements are correct, and they are matching more closely with the minor axis method, which was expected, thanks!

We should also add a test for this new ellipse feature, similar to [this one](https://github.com/neuropoly/axondeepseg/blob/e2cd6fcc7502db6241cbea4b7b47e1201599315d/test/morphometrics/test_compute_morphometrics.py#L94).

vs74:
Added tests for this new feature. Anyone wants to give this PR a shot, please feel free to review it. Also,  I've added a couple of tests and changes in [compute_morphometrics.py](https://github.com/neuropoly/axondeepseg/blob/vs74/ellipse_minor_axis/AxonDeepSeg/morphometrics/compute_morphometrics.py) and [launch_morphometrics_computation.py](url), so if we include changes in FSLeyes as well, this PR would be too huge to review.  I suggest, we open a separate PR to brainstorm for the FSLeyes, the design choices regarding how we should display to the user about using either **ellipse_minor_axis**  or **equivalent diameter** to compute the diameter.

.travis.yml
```diff
@@ -24,6 +24,7 @@ jobs:
       os: linux
       script: 
         - yes | conda install -c conda-forge fsleyes=0.33.1
+        - yes | conda install -c conda-forge h5py=2.10.0
```
jcohenadad, line 27: why this change? pls explain in your PR-- seems out of scope when not justified

.travis.yml
```diff
@@ -33,6 +34,7 @@ jobs:
       os: osx
       script: 
         - yes | conda install -c conda-forge fsleyes=0.33.1
+        - yes | conda install -c conda-forge h5py=2.10.0
```
jcohenadad, line 37: same as above

jcohenadad (changes requested):
why is "docs/source/_static/bad_mask_example.png" in this PR?

jcohenadad (changes requested):
@vs74 this branch seems to include older commits that have already been merged with master

.travis.yml
mathieuboudreau: Yup  - if necessary, it needs to be done in a new PR please.

mariehbourget:
This PR seems to be to merge between the branch `vs74/ellipse_minor_axis` on Vasudev fork, to the corresponding branch in ADS (and not to `master` in ADS). I think this is why it includes older commits and can be merged automatically.
To be able to review properly, we would have to first merge to the branch `vs74/ellipse_minor_axis` in ADS, then open another PR from this branch to `master`, no?

.travis.yml
mariehbourget: This is an older commit from PR #392

mathieuboudreau:
> This PR seems to be to merge between the branch `vs74/ellipse_minor_axis` on Vasudev fork, to the corresponding branch in ADS (and not to `master` in ADS). I think this is why it includes older commits and can be merged automatically.
> To be able to review properly, we would have to first merge to the branch `vs74/ellipse_minor_axis` in ADS, then open another PR from this branch to `master`, no?

No, we can change the base to be master on the PR page. I just did so.

AxonDeepSeg/morphometrics/compute_morphometrics.py
```diff
@@ -45,15 +46,17 @@ def get_pixelsize(path_pixelsize_file):
...
 
-def get_axon_morphometrics(im_axon, path_folder=None, im_myelin=None, pixel_size=None):
+def get_axon_morphometrics(im_axon, path_folder=None, im_myelin=None, pixel_size=None, ellipse = False):
     """
     Find each axon and compute axon-wise morphometric data, e.g., equivalent diameter, eccentricity, etc.
     If a mask of myelin is provided, also compute myelin-related metrics (myelin thickness, g-ratio, etc.).
     :param im_axon: Array: axon binary mask, output of axondeepseg
     :param path_folder: str: absolute path of folder containing pixel size file
     :param im_myelin: Array: myelin binary mask, output of axondeepseg
+    :param ellipse: Boolean: if axon is considered ellipse, set to True else if considered circle, set to False.
```
jcohenadad, line 56: a circle *is* an ellipse, so i would personally go with the inverse logic, e.g: `circle_approx={True, False}`

vs74: That seems reasonable, I will make changes.

vs74: Done in commit 7d85055

jcohenadad:
@vs74 if a PR is ready to review pls make sure to remove the "[WIP]" in the title-- also: in general, do not use WIP, but use draft mode-- see our guidelines

AxonDeepSeg/morphometrics/compute_morphometrics.py
```diff
@@ -108,8 +111,12 @@ def get_axon_morphometrics(im_axon, path_folder=None, im_myelin=None, pixel_size
         solidity = prop_axon.solidity
         # Eccentricity
         eccentricity = prop_axon.eccentricity
-        # Axon equivalent diameter in micrometers
-        axon_diam = prop_axon.equivalent_diameter * pixelsize
+        if  circle_approx:
```
jcohenadad, line 114:
```suggestion
        if circle_approx:
```

AxonDeepSeg/morphometrics/compute_morphometrics.py
```diff
@@ -154,33 +161,37 @@ def get_axon_morphometrics(im_axon, path_folder=None, im_myelin=None, pixel_size
 
     return stats_array
 
-def evaluate_myelin_thickness_in_px(axon_object, axonmyelin_object):
+def evaluate_myelin_thickness_in_px(axon_object, axonmyelin_object ,circle_approx):
```
jcohenadad, line 164:
```suggestion
def evaluate_myelin_thickness_in_px(axon_object, axonmyelin_object, circle_approx):
```

AxonDeepSeg/morphometrics/compute_morphometrics.py
```diff
@@ -280,7 +291,7 @@ def draw_axon_diameter(img, path_prediction, pred_axon, pred_myelin):
 
     path_folder = path_prediction.parent
 
-    stats_array = get_axon_morphometrics(pred_axon, path_folder)
+    stats_array = get_axon_morphometrics(pred_axon, path_folder , circle_approx = circle_approx)
```
jcohenadad, line 294: please be careful with PEP8... it's the 3rd typo...
```suggestion
    stats_array = get_axon_morphometrics(pred_axon, path_folder, circle_approx=circle_approx)
```

AxonDeepSeg/morphometrics/compute_morphometrics.py
```diff
@@ -324,11 +335,12 @@ def save_map_of_axon_diameters(path_folder, axon_diameter_figure):
     axon_diameter_figure.savefig(file_path)
 
 
-def get_aggregate_morphometrics(pred_axon, pred_myelin, path_folder):
+def get_aggregate_morphometrics(pred_axon, pred_myelin, path_folder, circle_approx = True):
```
jcohenadad, line 338: fix PEP8-- no space around "="-- and do it everywhere in your changes

AxonDeepSeg/morphometrics/launch_morphometrics_computation.py
```diff
@@ -10,13 +10,15 @@
...
 
-def launch_morphometrics_computation(path_img, path_prediction):
+def launch_morphometrics_computation(path_img, path_prediction, circle_approx = True):
     """
     This function is equivalent to the morphometrics_extraction notebook of AxonDeepSeg.
     It automatically performs all steps (computations, savings, displays,...) of the
     morphometrics extraction of a given sample.
     :param path_img: path of the input image (microscopy sample)
     :param path_prediction: path of the segmented image (output of AxonDeepSeg)
+    :param circle_approx: Boolean: if axon is considered circle, set to True else it is considered ellipse, and set to False.
```
jcohenadad, line 20: there is a *lot* of functions where the parameter `circle_approx` was added. Isn't there a better (less invasive) way to introduce that parameter without having to change the input params of all these functions?

vs74: I agree, will see a more elgant way to do this.

AxonDeepSeg/morphometrics/compute_morphometrics.py
vs74: Resolved in 3ba42bc

AxonDeepSeg/morphometrics/compute_morphometrics.py
vs74: Resolved in 3ba42bc

AxonDeepSeg/morphometrics/compute_morphometrics.py
vs74: Fixed in 3ba42bc

AxonDeepSeg/morphometrics/launch_morphometrics_computation.py
vs74: We could

1. Create a `config_morphometrics.json` file which will contain the circle_approx param, similar to `config_network.json file` which contains model parameters. The only downside I feel is that it would not make sense to add an exra config just to store one parameter. 

2. Encapsulte the functions of `compute_morpometrics.py` script in a class, which could have circle_approx as its attribute member. But for this we've to refactor this script and would include a couple of changes. 

On top of my head, I could only think of these ways to deal with this. @mathieuboudreau do you've any insights on this?

vs74: PS: Personally, I feel the default way is better.  

@mathieuboudreau  for your reference: The functions: [evaluate_myelin_thickness_in_px](https://github.com/vs74/axondeepseg/blob/3ba42bcaff974cebf3a816ea6302c6effc81e6d6/AxonDeepSeg/morphometrics/compute_morphometrics.py#L164), [get_axon_morphometrics](https://github.com/vs74/axondeepseg/blob/3ba42bcaff974cebf3a816ea6302c6effc81e6d6/AxonDeepSeg/morphometrics/compute_morphometrics.py#L49), [draw_axon_diameter](https://github.com/vs74/axondeepseg/blob/3ba42bcaff974cebf3a816ea6302c6effc81e6d6/AxonDeepSeg/morphometrics/compute_morphometrics.py#L278), [get_aggregate_morphometrics](https://github.com/vs74/axondeepseg/blob/3ba42bcaff974cebf3a816ea6302c6effc81e6d6/AxonDeepSeg/morphometrics/compute_morphometrics.py#L338) and [launch_morphometrics_computation](https://github.com/vs74/axondeepseg/blob/3ba42bcaff974cebf3a816ea6302c6effc81e6d6/AxonDeepSeg/morphometrics/launch_morphometrics_computation.py#L13) call [get_axon_morphometrics](https://github.com/vs74/axondeepseg/blob/3ba42bcaff974cebf3a816ea6302c6effc81e6d6/AxonDeepSeg/morphometrics/compute_morphometrics.py#L49) function, which has `circle_approx` param to let the user choose between **equivalent_diameter** or **minor_axis of ellipse** to compute diameter. For this reason, I added additional parameter `circle_approx` in all these functions. Now, the question is [this](https://github.com/neuropoly/axondeepseg/pull/399#discussion_r530437803)

jcohenadad: that should probably be the purpose of another PR. We should move the discussion to a new issue. Let's see what @mathieuboudreau thinks first

mathieuboudreau: I'm going to reflect on this a bit and will either respond here if I have a strong opinion or we can discuss how to move forward at the next meeting.

vs74:
This PR is ready for review.

mathieuboudreau:
> This PR is ready for review.

Starting to review now.+

AxonDeepSeg/morphometrics/compute_morphometrics.py
```diff
@@ -45,15 +46,17 @@ def get_pixelsize(path_pixelsize_file):
         return pixelsize
 
 
-def get_axon_morphometrics(im_axon, path_folder=None, im_myelin=None, pixel_size=None):
+def get_axon_morphometrics(im_axon, path_folder=None, im_myelin=None, pixel_size=None, circle_approx=True):
```
mathieuboudreau, line 49: Suggestion: I think setting the variable a boolean and calling it "circle_approx" may not be sufficiently clear to future devs. When the value is false, there's nothing really indicating that it will then be set to ellipse. We could maybe instead call it `axon_shape` and set it to `"circle"` or `"ellipses"`, or something of the sort.

test/morphometrics/test_compute_morphometrics.py
```diff
@@ -170,6 +253,32 @@ def test_get_axon_morphometrics_with_unexpected_myelin_mask_simulated_axons(self
             assert axon_prop['myelin_thickness'] == pytest.approx(0.0, rel=0.01)
             assert axon_prop['myelin_area'] == pytest.approx(0.0, rel=0.01)
             assert axon_prop['gratio'] == pytest.approx(1.0, rel=0.01)
+    
+    @pytest.mark.unit
+    def test_get_axon_morphometrics_with_unexpected_myelin_mask_simulated_axons_with_axon_as_ellipse(self):
```
mathieuboudreau, line 258: Since all the test images generated for this test are perfect circles, I feel like all these tests dealing with the the ellipse settings are missing some of the functionality. We should have a test image with some ellipses with known minor and major axes, and test that your functionality returns these parameters. i.e. I know all your tests pass, but they may not be testing the primary functionality of this new feature, so it's difficult for me as a new user read the tests and understand the expected behaviour/functionality.

mathieuboudreau:
On naming it `circle_approx` in the code; yeah I'm quite certain we need to rename this, because now in the morphometrics Jupyter Notebook, we are instead setting the variable `ellipse=true`, so there's a lack of consistency between the API and the internal code here with the naming I think.

Capture d’écran 2020-12-10 à 23 31 04

mathieuboudreau:
And actually, that functionality is broken too:

Capture d’écran 2020-12-10 à 23 33 02

mathieuboudreau:
Nevermind for the previous error; I think there was a problem between conda env and Jupyter

AxonDeepSeg/morphometrics/compute_morphometrics.py
```diff
@@ -108,8 +111,12 @@ def get_axon_morphometrics(im_axon, path_folder=None, im_myelin=None, pixel_size
...
         # Eccentricity
         eccentricity = prop_axon.eccentricity
-        # Axon equivalent diameter in micrometers
-        axon_diam = prop_axon.equivalent_diameter * pixelsize
+        if circle_approx:
+            # Axon equivalent diameter in micrometers
+            axon_diam = prop_axon.equivalent_diameter * pixelsize
+        else: 
+            # Axon diameter using ellipse minor axis in micrometers
+            axon_diam = prop_axon.minor_axis_length * pixelsize
```
mathieuboudreau, line 119: So for the ellipse case, are we not storing the major and minor axes and providing it to the user? That's what I was expected, is that what we wanted?

mathieuboudreau: If I understand correctly, the current implementation only saves the minor axes as the equivalent diameter for the ellipse case, and the major axes is never used or saved, correct? Is this the behaviour that you wanted @mariehbourget ?

AxonDeepSeg/morphometrics/compute_morphometrics.py
vs74: Thanks @mathieuboudreau for the suggestion, I agree with the renaming of the variable. I've fixed this in commit 8a42b35

test/morphometrics/test_compute_morphometrics.py
vs74: That's a good catch. I've made a test image containing axons which are ellipse. 

Untitled Diagram (4)

Let me know what do you think about this image. In the tests, I'm planning to add this test image for computing the diameter using `ellipse_minor_axis`. For this image, I'll do manual measurements on them  which should be similar to the new functionality.

jcohenadad: there are black pixels between white and gray (shouldn't be the case)

vs74: Right; corrected the image.

jcohenadad:
The notebook **04-morphometrics_extraction.ipynb**  is 1.16 MB, which suggests it probably includes binaries/images --> we need to clear that otherwise repos accumulates fat
--> meaning we should squash before merging

AxonDeepSeg/morphometrics/compute_morphometrics.py
vs74: As discussed in the meeting, we are only interested in the minor axis of ellipse for computing morphometrics, so we needn't worry about the major axis.

vs74:
For the documentation concerned regarding this feature, I'm not sure what to add in the documentation becuase there are only changes done in the functions of  [compute_morphometrics](https://github.com/neuropoly/axondeepseg/blob/master/AxonDeepSeg/morphometrics/compute_morphometrics.py) script and [launch_morphometrics_computation](https://github.com/neuropoly/axondeepseg/blob/master/AxonDeepSeg/morphometrics/launch_morphometrics_computation.py) script. 

According to me, it would make more sense to have documentation for the CLI to generate morphometrics file (see PR #434 ), wherin I've added  `-a --axonshape` flag and documentation for that in PR #434 which will allow the user can select the desired `axon_shape` ( `ellipse` or `circle`).

 So, I don't think documentation for this PR would be needed.

mathieuboudreau:
@vs74 PR #179 was merged, so you can now generate test images with ellipses using it.

vs74:
Shall we open a new PR for automatically generating synthetic images (  using #179 recently merged  into master) when axon shape is **`circle`** or we shall open another PR for the same? Currently, we are using a synthetic mask image in the test files hosted on OSF and testing the behaviour using the synthetic mask.

mathieuboudreau:
> Shall we open a new PR for automatically generating synthetic images ( using #179 recently merged into master) when axon shape is **`circle`** or we shall open another PR for the same? Currently, we are using a synthetic mask image in the test files hosted on OSF and testing the behaviour using the synthetic mask.

No, not a new PR, but a new issue is warranted. Those masks are not directly related to this PR, so it doesn't belong here.

vs74:
So, I tested added some tests which generate simulated axons ellipsoidal in shape with varying `plane angle`. The simulated axons are very much identical to the test image of simulated axons, which were circular except for the fact that they have varying `plane angle`, which makes the axon's shape ellipse. 

``` 
    c.generate_axon(axon_radius=50, center=[100, 100], gratio=0.9, plane_angle=0)
    c.generate_axon(axon_radius=45, center=[200, 200], gratio=0.8, plane_angle=10)
    c.generate_axon(axon_radius=40, center=[300, 300], gratio=0.7, plane_angle=20)
    c.generate_axon(axon_radius=35, center=[400, 400], gratio=0.6, plane_angle=30)
    c.generate_axon(axon_radius=30, center=[520, 520], gratio=0.5, plane_angle=40)
    c.generate_axon(axon_radius=23, center=[650, 650], gratio=0.4, plane_angle=50)
    c.generate_axon(axon_radius=18, center=[750, 750], gratio=0.3, plane_angle=60)
    c.generate_axon(axon_radius=12, center=[830, 830], gratio=0.2, plane_angle=70)
    c.generate_axon(axon_radius=6, center=[920, 920], gratio=0.1, plane_angle=80)

    c.save()
```


image



     
The tests seem to fail for the axons with higher plane angle or in other words, the more the axons are ellipsoidal in shape, the measurements such as `g_ratio` and `myelin_thickness` are different from the ones it was generated on.

For example, for the axon instances [5, 6, 7, 8] there is a difference in the `gratio` and `myelin_thickness` measurements computed by the new feature and the measurements (`gratio` and `myelin_thickness`) of synthetic axons. Hence, the test cases fail. 

I manually measured the `axon_diameter`, `myelin_thickness` and `gratio` for this test image following the protocol mentioned [here](https://github.com/neuropoly/axondeepseg/discussions/347#discussioncomment-232032). These manual measurements were much similar to the measurements generated by the new feature as compared to the measurements of synthetic axons. 

So a question for @mathieuboudreau, shall I set the [measurements](https://github.com/vs74/axondeepseg/blob/3ee1d5715bc1ee894506dcf3f3a4d5c13579c77c/test/morphometrics/test_compute_morphometrics.py#L216-L239) (`gratio` and `myelin_thickness`) to the manual measurements for testing purposes?

vs74:
@mariehbourget, I've added some literature studies[ here](https://axondeepseg--434.org.readthedocs.build/en/434/documentation.html#axon-shape-circle-vs-ellipse), but these are not sufficient. If you happen to know literature where axon shape being `circle` or `ellipse` are used for morphometrics, feel free to contribute in the documentation of this PR or add them in the comments, which I will add on in the RTD. Thanks!!

vs74:
@mathieuboudreau did you get a chance to see my previous comment? I'd love to hear your thoughts on this.

mathieuboudreau:
@vs74 I would recommend doing one image where only the gratio varies, and another where only the plane angle varies. The first case should definitely agree when getting morphometrics from the code, however I'm not convinced that the morphometrics calculation (well, the g-ratio value) is supposed to match for the case where the plane angle varies; in this case though, the major and minor axes should match with the ones I simulated (see [here](https://github.com/neuropoly/axondeepseg/blob/28666b8be47889df24b6839cefd247e06de16102/AxonDeepSeg/visualization/simulate_axons.py#L74-L75) to see how they're calculated from the plane angle).

vs74:
@mathieuboudreau correct me if I am wrong. Would the myelin thickness see[ here ](https://github.com/neuropoly/axondeepseg/blob/28666b8be47889df24b6839cefd247e06de16102/AxonDeepSeg/visualization/simulate_axons.py#L72) be `axon_radius * (1/gratio - 1)`?

AxonDeepSeg/morphometrics/compute_morphometrics.py
```diff
@@ -143,7 +152,7 @@ def get_axon_morphometrics(im_axon, path_folder=None, im_myelin=None, pixel_size
                 stats['myelin_thickness'] = myelin_thickness
                 stats['myelin_area'] = myelin_area
                 stats['axonmyelin_area'] = axonmyelin_area
-                stats['gratio'] = np.sqrt(axon_area / axonmyelin_area)
+                stats['gratio'] = (axon_diam / 2) / (axon_diam / 2 + myelin_thickness)
```
vs74, line 155: I realized that in the last meeting I wasn't able to properly put my point forward of why I went on to change the gratio from `np.sqrt(axon_area / axonmyelin_area)` to `(axon_diam / 2) / (axon_diam / 2 + myelin_thickness)`. Well, gratio is defined as `axon_radius / (axon_radius + myelin thickness)`, which is the general formula that would fit in for the both **ellipse** as well as **circle** shaped axons.  

If we set the gratio as `np.sqrt(axon_area / axonmyelin_area)`, this wouldn't work for the case when the axons are ellipsoidal in shape and the gratio measured in this case would be inaccurate because it would be similar to that when axons are circular in shape ( as the area of the axons and the axon + myelin would remain same for the ellipsoidal shape)

Hence, I decided to drop off the earlier formula of gratio which was dependent on the area of the axon and axon + myelin.

vs74:
> @mathieuboudreau correct me if I am wrong: wouldn't the myelin thickness see[ here ](https://github.com/neuropoly/axondeepseg/blob/28666b8be47889df24b6839cefd247e06de16102/AxonDeepSeg/visualization/simulate_axons.py#L72) be `axon_radius * (1/gratio - 1)`?

@mathieuboudreau could you confirm the formula for myelin thickness so that I will proceed with the tests? Thanks!!

mathieuboudreau:
Hi Vasudev,

I'm currently working on a project that has a deadline tomorrow, so I will recheck the equation after that.

Thanks for the reminder,

Best,

Mathieu

--
Mathieu Boudreau, PhD
Research Fellow | Montreal Heart Institute
Software Developer | qMRLab and AxonDeepSeg
Lead Editor | MRM Highlights<http://linkedin.com/company/mrm-highlights/><https://www.linkedin.com/company/mrm-highlights/>

vs74:
No worries, it's not that urgent.

mathieuboudreau:
> > @mathieuboudreau correct me if I am wrong: wouldn't the myelin thickness see[ here ](https://github.com/neuropoly/axondeepseg/blob/28666b8be47889df24b6839cefd247e06de16102/AxonDeepSeg/visualization/simulate_axons.py#L72) be `axon_radius * (1/gratio - 1)`?
> 
> @mathieuboudreau could you confirm the formula for myelin thickness so that I will proceed with the tests? Thanks!!

@vs74 I had the time to revisit this, and indeed it appears you are correct; there is an error in the equation:

PXL_20210201_170058836

I will create a branch to patch this and add tests that would have caught this in the first place. I'll let you know shortly if I encounter any other problems (conceptually or in the algorithm) prior to creating a pull request for you to review.

vs74:
> I will create a branch to patch this and add tests that would have caught this in the first place. I'll let you know shortly if I encounter any other problems (conceptually or in the algorithm) prior to creating a pull request for you to review.

Thank you @mathieuboudreau for revisiting this formula. I'm not sure if I as a contributor can review the PR of the repo. Is it possible to provide rights to contributors to review PR's? Also, note that the patched PR you will submit will result in failure of the tests because of PR #441, so we should deal with that first.

mathieuboudreau:
Ahh ok, whether or not you can approve the review, I'll ask you to check the PR since you're the one who discovered the bug!

vs74:
An update on this PR: the tests seem to pass when the parameters **"gratio"**, **"radius"** and **"plane angle"** are varied in isolation while keeping other parameters constant. However, as discussed before in the meeting, when all the parameters are varied, the tests seem to fail [PR #399 (comment)](https://github.com/neuropoly/axondeepseg/pull/399#issuecomment-762059816)

vs74:
@mathieuboudreau did you get a chance to revisit this PR?

test/morphometrics/test_compute_morphometrics.py
```diff
@@ -147,6 +209,224 @@ def test_get_axon_morphometrics_with_myelin_mask_simulated_axons(self):
...
+        # Simulated image of axon and myelin to test morphoterics generated by ADS when axon shape is ellipse
+        
+        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[80, 80], gratio=0.9, plane_angle=0)
+        self.image_sim_ellipse.generate_axon(axon_radius=45, center=[180, 180], gratio=0.8, plane_angle=10)
+        self.image_sim_ellipse.generate_axon(axon_radius=40, center=[280, 280], gratio=0.7, plane_angle=20)
+        self.image_sim_ellipse.generate_axon(axon_radius=35, center=[380, 380], gratio=0.6, plane_angle=30)
+        self.image_sim_ellipse.generate_axon(axon_radius=30, center=[480, 480], gratio=0.5, plane_angle=40)
+        self.image_sim_ellipse.generate_axon(axon_radius=23, center=[580, 580], gratio=0.4, plane_angle=50)
+        self.image_sim_ellipse.generate_axon(axon_radius=18, center=[700, 700], gratio=0.3, plane_angle=60)
+        self.image_sim_ellipse.generate_axon(axon_radius=12, center=[800, 850], gratio=0.2, plane_angle=70)
```
mathieuboudreau, line 224: @vs74  I think here `axon_radius` should be 10 and not 12, as `axon_diam_sim` below is 20 and not 24.

mathieuboudreau:
@vs74 I've finished doing as much digging as I could for now. I think I found one bug in your test (see comment above), but unsure if this was the sole reason for your test failing. 

Here are my analytical derivations, and they all appear to be consistent with my code:

<details>

PXL_20210324_163459215

PXL_20210324_163502516

PXL_20210324_163506298

PXL_20210324_164259068

</details>

The only thing I had a minor concern about was the fact that myelin_major_radius requires the radius, plane_angle, and *myelin_thickness* values, because *myelin_thickness* already uses the plane angle to calculate an updated value, so I'm wondering if maybe there's some duplication in the logic here. To test this out, instead of varying all three parameters at a time, I've added 4 more tests in a branch that's a copy of this PR (in commit 4fff10af) to test the combinations of pairs of two parameters varying (plus one more for the case of zero plane angle). If one or some of these pass, that will give us more information into the conflicting parameters and could help debug this. TBD.

jcohenadad:
> Here are my analytical derivations, and they all appear to be consistent with my code:

so beautiful (and rare) to see derivations on paper these days 😍

mathieuboudreau:
@vs74 your "single parameter varying" tests appear to be sensitive to the values given. i.e., I changed the parameters of your test `test_get_axon_morphometrics_with_myelin_mask_simulated_axons_with_axon_as_ellipse_gratio_varies` to the following:

Capture d’écran 2021-03-24 à 19 44 47

Capture d’écran 2021-03-24 à 19 44 47

And after that the test fails, 

Capture d’écran 2021-03-24 à 19 42 55

Which is super odd actually, because those parameters are all the same as the first set of parameters of your `test_get_axon_morphometrics_with_myelin_mask_simulated_axons_with_axon_as_ellipse` test, which didn't fail at `myelin_thickness` check but at the `gratio` check,

Capture d’écran 2021-03-24 à 19 47 58

and I don't know why that would be, even if there was a bug in my simulation code... ???

At this point I'm stuck again and out of ideas of where to look to easily fix the state of this issue. I think we may want to go back to the drawing board and start fresh with baby steps, slowly building the functions/features one feature/test at a time :/

vs74:
> @vs74 your "single parameter varying" tests appear to be sensitive to the values given. i.e., I changed the parameters of your test test_get_axon_morphometrics_with_myelin_mask_simulated_axons_with_axon_as_ellipse_gratio_varies to the following:

I didn't test the single parameter tests extensively with a wider range of values, so that could be the reason for the failure of the tests. Perhaps I got lucky with my previous numbers.

>Which is super odd actually, because those parameters are all the same as the first set of parameters of your test_get_axon_morphometrics_with_myelin_mask_simulated_axons_with_axon_as_ellipse test, which didn't fail at myelin_thickness check but at the gratio check,

Yes, it is weird. I need to revisit the tests again and I'll try to figure out the cause of this problem. 

If it doesn't work out, then you are right that we have to take baby steps to figure out the cause of failure of the tests. I'll let you know once I have looked at it.

Thanks @mathieuboudreau  for looking into it.

test/morphometrics/test_compute_morphometrics.py
vs74: Yeah, you are right. It should be 10 instead of 12. Thanks for catching it.

vs74:
@mathieuboudreau I revisited this PR and simulated your case where you mentioned that the tests in this PR were sensitive to the gratio values https://github.com/neuropoly/axondeepseg/pull/399#issuecomment-806222808

## UNIT TEST (@mathieuboudreau tried)
```
    @pytest.mark.unit
    def test_get_axon_morphometrics_with_myelin_mask_simulated_axons_with_axon_as_ellipse_gratio_varies(self):

        # varying g ratio 
        self.image_sim_ellipse = SimulateAxons()

        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[115, 90], gratio=0.9, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[200, 200], gratio=0.8, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[305, 305], gratio=0.7, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[410, 410], gratio=0.6, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[510, 520], gratio=0.5, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[650, 650], gratio=0.4, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[800, 800], gratio=0.3, plane_angle=0)
    
        self.image_sim_ellipse.save(self.image_sim_ellipse_path)

        gratio_sim = np.array([
                                0.9,
                                0.8,
                                0.7,
                                0.6,
                                0.5,
                                0.4,
                                0.3,
                                ])

        axon_diam_sim = np.array([
                                100,
                                100,
                                100,
                                100,
                                100,
                                100,
                                100,
                                ])

        myelin_thickness_sim = (axon_diam_sim / 2) * (1/gratio_sim - 1)

        # Read paths and compute axon/myelin masks
        pred = imageio_imread(self.image_sim_ellipse_path)
        pred_axon = pred > 200
        pred_myelin = np.logical_and(pred >= 50, pred <= 200)

        # compute axon morphometrics
        stats_array = get_axon_morphometrics(pred_axon, str(self.image_sim_ellipse_path.parent), im_myelin=pred_myelin, axon_shape=self.axon_shape)

        for ii in range(0,len(gratio_sim)):

            assert stats_array[ii]['gratio'] == pytest.approx(gratio_sim[ii], rel=0.1)
            assert stats_array[ii]['axon_diam'] == pytest.approx(axon_diam_sim[ii], rel=0.1)
            assert stats_array[ii]['myelin_thickness'] == pytest.approx(myelin_thickness_sim[ii], rel=0.1)

```


Here is the simulated image of the gratio values you tested with

image



Could it possible due to overlapping axons on the simulated image the tests are failing? I tried axon_radius ranging from 5-38, where the axonmyelin objects were not overlapping, and the tests passed. However, when the axonmyelin objects were overlapping the test cases were failing, that is for **axon_radius > 41**.

**Note** The tests often fail when the axonmyelin objects overalps in the simulated image. This is true for not only `test_get_axon_morphometrics_with_myelin_mask_simulated_axons_with_axon_as_ellipse_gratio_varies`, but also for other tests as well. 

Let me know what do you think.

jcohenadad:
i would recommend to avoid overlap as it creates an indeterminacy on the true solution

mathieuboudreau:
>Could it be possible that due to overlapping axons on the simulated image the tests are failing? 

Yes, that's very likely at least one cause of it. I'd recommend we remove every other axon (axon #2, #4, etc) and try again to see if the results of the larger axons agree better. This way, we'll still cover the same range, but will avoid overlapping. Thanks for checking this!

vs74:
> > Could it be possible that due to overlapping axons on the simulated image the tests are failing?
> 
> Yes, that's very likely at least one cause of it. I'd recommend we remove every other axon (axon #2, #4, etc) and try again to see if the results of the larger axons agree better. This way, we'll still cover the same range, but will avoid overlapping. Thanks for checking this!

## Test 1
I commented axon # 4 and # 6 with the same **axon_radius = 50** and **plane_angle = 0** for the simulated image given below
<details>
<summary> Commented unit test </summary>

```
@pytest.mark.unit
    def test_get_axon_morphometrics_with_myelin_mask_simulated_axons_with_axon_as_ellipse_gratio_varies(self):

        # varying g ratio 
        self.image_sim_ellipse = SimulateAxons()

        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[115, 90], gratio=0.9, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[200, 200], gratio=0.8, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[305, 305], gratio=0.7, plane_angle=0)
        # self.image_sim_ellipse.generate_axon(axon_radius=50, center=[410, 410], gratio=0.6, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[510, 520], gratio=0.5, plane_angle=0)
        # self.image_sim_ellipse.generate_axon(axon_radius=50, center=[650, 650], gratio=0.4, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[800, 800], gratio=0.3, plane_angle=0)
    
        self.image_sim_ellipse.save(self.image_sim_ellipse_path)

        gratio_sim = np.array([
                                0.9,
                                0.8,
                                0.7,
                                #0.6,
                                0.5,
                                #0.4,
                                0.3,
                                ])

        axon_diam_sim = np.array([
                                100,
                                100,
                                #100,
                                100,
                                100,
                                #100,
                                100,
                                ])

        myelin_thickness_sim = (axon_diam_sim / 2) * (1/gratio_sim - 1)

        # Read paths and compute axon/myelin masks
        pred = imageio_imread(self.image_sim_ellipse_path)
        pred_axon = pred > 200
        pred_myelin = np.logical_and(pred >= 50, pred <= 200)

        # compute axon morphometrics
        stats_array = get_axon_morphometrics(pred_axon, str(self.image_sim_ellipse_path.parent), im_myelin=pred_myelin, axon_shape=self.axon_shape)

        for ii in range(0,len(gratio_sim)):

            assert stats_array[ii]['gratio'] == pytest.approx(gratio_sim[ii], rel=0.1)
            assert stats_array[ii]['axon_diam'] == pytest.approx(axon_diam_sim[ii], rel=0.1)
            assert stats_array[ii]['myelin_thickness'] == pytest.approx(myelin_thickness_sim[ii], rel=0.1)

```
</details>


image


## Test 2

I even tried with mostly large axons by commenting axon # 2 and axon # 3 for the simulated image given below, and the tests passed.
 
<details>
<summary> Unit test with mostly large axon myelin instances</summary>

```
   @pytest.mark.unit
    def test_get_axon_morphometrics_with_myelin_mask_simulated_axons_with_axon_as_ellipse_gratio_varies(self):

        # varying g ratio 
        self.image_sim_ellipse = SimulateAxons()

        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[115, 90], gratio=0.9, plane_angle=0)
        #self.image_sim_ellipse.generate_axon(axon_radius=50, center=[200, 200], gratio=0.8, plane_angle=0)
        # self.image_sim_ellipse.generate_axon(axon_radius=50, center=[305, 305], gratio=0.7, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[220, 220], gratio=0.6, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[360, 360], gratio=0.5, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[550, 550], gratio=0.4, plane_angle=0)
        self.image_sim_ellipse.generate_axon(axon_radius=50, center=[800, 800], gratio=0.3, plane_angle=0)
    
        self.image_sim_ellipse.save(self.image_sim_ellipse_path)

        gratio_sim = np.array([
                                0.9,
                                #0.8,
                                #0.7,
                                0.6,
                                0.5,
                                0.4,
                                0.3,
                                ])

        axon_diam_sim = np.array([
                                100,
                                #100,
                                #100,
                                100,
                                100,
                                100,
                                100,
                                ])

        myelin_thickness_sim = (axon_diam_sim / 2) * (1/gratio_sim - 1)

        # Read paths and compute axon/myelin masks
        pred = imageio_imread(self.image_sim_ellipse_path)
        pred_axon = pred > 200
        pred_myelin = np.logical_and(pred >= 50, pred <= 200)

        # compute axon morphometrics
        stats_array = get_axon_morphometrics(pred_axon, str(self.image_sim_ellipse_path.parent), im_myelin=pred_myelin, axon_shape=self.axon_shape)

        for ii in range(0,len(gratio_sim)):

            assert stats_array[ii]['gratio'] == pytest.approx(gratio_sim[ii], rel=0.1)
            assert stats_array[ii]['axon_diam'] == pytest.approx(axon_diam_sim[ii], rel=0.1)
            assert stats_array[ii]['myelin_thickness'] == pytest.approx(myelin_thickness_sim[ii], rel=0.1)

```
</details>

image


## **Ananlysis:** 

I guess because of the overlapping axon myelin instances the tests were failing. All the tests for this PR are passing when the simulated image has no overlapping axonmyelin instances. We have tested on [only varying the **plane angle**](https://github.com/vs74/axondeepseg/blob/b2f77432303eed0e39872ec32ffecdfcc954f5be/test/morphometrics/test_compute_morphometrics.py#L318), [only varying **gratio**](https://github.com/vs74/axondeepseg/blob/b2f77432303eed0e39872ec32ffecdfcc954f5be/test/morphometrics/test_compute_morphometrics.py#L266-L315), [only varying **axon_radius**](https://github.com/vs74/axondeepseg/blob/b2f77432303eed0e39872ec32ffecdfcc954f5be/test/morphometrics/test_compute_morphometrics.py#L373-L428), and lastly varying all the three parameters [(**axon_radius**, **plane_angle**, and **gratio**)](https://github.com/vs74/axondeepseg/blob/b2f77432303eed0e39872ec32ffecdfcc954f5be/test/morphometrics/test_compute_morphometrics.py#L210-L262).

@mathieuboudreau do you want to try out some additional tests on this feature? If not, then I will go ahead with cleaning this PR and would then ask for final reivew.

vs74:
This PR is up for a reivew.

mathieuboudreau:
Excellent debugging @vs74 – thank you so much for discovering this bug in our test datasets! I'll review ASAP.

test/morphometrics/test_compute_morphometrics.py
mathieuboudreau: I see you changed the axon_diam_sim to 24 instead of the axon radius here to 10

mathieuboudreau:
@vs74 did you merge your changes to the test to avoid overlap? I don't see them there; there's always 8-9 axons instead of your proposed 5. Also, why are the tests currently passing if these are note included in the tests? This is the last thing I need to see before merging.

docs/source/documentation.rst
```diff
@@ -364,6 +371,38 @@ This will generate **'axon_morphometrics.xlsx'** file in each of folders::
...
+* Payne et al: https://pubmed.ncbi.nlm.nih.gov/22879411/
+* Fehily et al: https://pubmed.ncbi.nlm.nih.gov/30702755/
+
+
+.. NOTE :: In the literature, both equivalent diameter and ellipse minor axis are used to compute the morphometrics. 
+           Thus, depending on the usecase, the user is advised to choose axon shape accordingly.
+           
+
 
 Morphometrics file
```
mariehbourget, line 407: Now that the diameter is not always the equivalent diameter of a circle, we should update the definitions of `gratio`, `axon_diameter` and `myelin_thickness` in the new headings table.

vs74:
> @vs74 did you merge your changes to the test to avoid overlap? I don't see them there; there's always 8-9 axons instead of your proposed 5. Also, why are the tests currently passing if these are note included in the tests? This is the last thing I need to see before merging.

In order to avoid overlap, for the test [only varying **gratio**](https://github.com/vs74/axondeepseg/blob/b2f77432303eed0e39872ec32ffecdfcc954f5be/test/morphometrics/test_compute_morphometrics.py#L266-L315), I switched to the old parameter values `axon_radius=25, plane_angle=60`(8-9 axons) with varying gratio values for the axon, which were passing earlier too. However, when you tested on this PR with `axon_radius=50, plane_angle=0` (https://github.com/neuropoly/axondeepseg/pull/399#issuecomment-806222808) on this tests itself, it failed because the **axon_radius** was big enough to cause overlap. 

Thus, I switched to old values `axon_radius=25, plane_angle=60`(8-9 axons) for the test [only varying **gratio**](https://github.com/vs74/axondeepseg/blob/b2f77432303eed0e39872ec32ffecdfcc954f5be/test/morphometrics/test_compute_morphometrics.py#L266-L315). Also, you tried testing it for `plane_angle=0` for this test which falied, and this **plane_angle** value wasn't making sense because this test was specifically designed to test **elliptical-shaped** axon myelin simulated image. 

Lastly, I changed the **center** parameter of other tests ( I guess for 1-2 tests) while keeping the simulated axons to 8-9 axons, which were preventing the tests from failing due to the overlapping nature of axons. 

I have two suggestions: 

1. You can try locally testing the test snippet for the particular value you tested on https://github.com/neuropoly/axondeepseg/pull/399#issuecomment-850205734, it will fail. However, when you will try test for non-overlapping axons https://github.com/neuropoly/axondeepseg/pull/399#issuecomment-850486505, it will pass.

2. Or, I can change the test (only varying gratio) to 5 axons with **axon_radius = 50** (non-overlapping) but with different plane_angle

docs/source/documentation.rst
vs74: Nice catch. I'll fix this.

docs/source/documentation.rst
```diff
@@ -326,7 +326,7 @@ Morphometrics of a single image
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 Before computing the morphometrics of an image, make sure it has been segmented using AxonDeepSeg ::
```
mariehbourget, line 327: I think the -a argument should be listed as optional argument above.

mathieuboudreau:
@vs74 please go with option number 2; I don't have my old laptop capable of running ADS on hand at the moment, so I wouldn't be able to do number 1 before tomorrow.

mariehbourget (commented):
Thank you @vs74 for the implementation of this feature. I tested it in CLI and everything LGTM. I left a couple comment regarding documentation.

vs74:
> @vs74 please go with option number 2; I don't have my old laptop capable of running ADS on hand at the moment, so I wouldn't be able to do number 1 before tomorrow.

Sure, I'll do the changes for the test after the meeting.

docs/source/documentation.rst
vs74: Yes, I agree. I'll fix this.

vs74: Addressed these changes in the latest commit of this PR.

docs/source/documentation.rst
vs74: Fixed.

vs74:
@mathieuboudreau I updated the test  [only varying **gratio**](https://github.com/vs74/axondeepseg/blob/b2f77432303eed0e39872ec32ffecdfcc954f5be/test/morphometrics/test_compute_morphometrics.py#L266-L315) to **Test 2 (non-overlapping axons with large radius)** https://github.com/neuropoly/axondeepseg/pull/399#issuecomment-850486505 with just **plane angle== 20** instead of **plane_angle==0** because the tests are for elliptical simulated image.

mathieuboudreau:
> @mathieuboudreau I updated the test [only varying **gratio**](https://github.com/vs74/axondeepseg/blob/b2f77432303eed0e39872ec32ffecdfcc954f5be/test/morphometrics/test_compute_morphometrics.py#L266-L315) to **Test 2 (non-overlapping axons with large radius)** [#399 (comment)](https://github.com/neuropoly/axondeepseg/pull/399#issuecomment-850486505) with just **plane angle== 20** instead of **plane_angle==0** because the tests are for elliptical simulated image.

The plane angle were 10 before you changed them to 20, not 0, so not sure if this modification changed anything.

mathieuboudreau (approved):
Approved; thank you for your hard work on this issue @vs74 !

Issue #400, opened 2020-11-19
Use squeletonization to close myelin


jcohenadad:
As pointed out in #397, there is a need to manually correct myelin masks, including closing them so that the flood fill algo works. There might be ways to "close" myelin masks automatically after a [squeletonization](https://scikit-image.org/docs/dev/auto_examples/edges/plot_skeleton.html) algo. To investigate...

Stoyan-I-A:
For some visualization:
![image](https://user-images.githubusercontent.com/43790661/99916353-d2228300-2cd7-11eb-979d-eda3ee9f8a90.png)

mathieuboudreau:
@hermancollin this old issue may be relevant to the user issue you were discussing today.

mathieuboudreau:
Stale

PR #402, opened 2020-11-22
Fill the myelin with the "remove_small_holes" tool

### LINKED ISSUE #401: Use 'remove_small_holes' to fill the myelin
As discussed in #397, the `remove_small_holes` function from scikit-image might be a nice way to fill the myelin without relying on the `center of mass` of myelin objects.

Stoyan-I-A:
Check this presentation for more information on the tool: https://docs.google.com/presentation/d/1aRl7C1IzzdhXzUBHk4tnIzU3L2r-DJ1540hnQdivzJE/edit#slide=id.p

Fixes #401 and #384.

This is an alternative to #397 and #400.

The limits of this tool will be axons which have a very large area (more than 10% of the image's size, we can ajust this) and images in which the background represents a very small area (less than 10% of the image, we can ajust this).

### DIFF
--- a/AxonDeepSeg/postprocessing.py
+++ b/AxonDeepSeg/postprocessing.py
@@ -135,3 +135,24 @@ def generate_axon_numbers_image(centroid_index, x0_array, y0_array, image_size, 
     image_array = np.asarray(number_image)
 
     return image_array.astype(np.uint8)
+
+def fill_myelin_holes(myelin_array):
+    """
+    This function uses the fill_small_holes function from scikit-image to fill closed myelin objects with the axon mask.
+    The maximum area of an axon is defined as 10% of the image's size.
+    :param myelin_array: the binary array corresponding to the myelin mask
+    :return: the binary axon array corresponding to the axon mask after the floodfill
+    """
+    # Get the dimensions of the image
+    image_dims = myelin_array.shape
+
+    # Define the maximum axon area as 10% of the image's size
+    maximum_axon_area = 0.1 * image_dims[0] * image_dims[1]
+
+    #Fill the myelin array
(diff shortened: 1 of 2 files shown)


Stoyan-I-A:
@mariehbourget I tried this implementation with the myelin mask you shared with me. I obtained a good result.

mariehbourget:
> @mariehbourget I tried this implementation with the myelin mask you shared with me. I obtained a good result.

I tested it and got good results too on that specific myelin mask.

However, the point you mentioned about small background area will be an issue for many images where myelin are touching, example from our SEM test images:
remove_small_holes

Same thing happens with our TEM test image.

jcohenadad:
> However, the point you mentioned about small background area will be an issue for many images where myelin are touching, example from our SEM test images:

how does the squeletization look like for these cases?

Stoyan-I-A:
image
Here's the myelin mask from MH's image and its skeleton. The touching myelin objects are still problematic.

jcohenadad:
> Here's the myelin mask from MH's image and its skeleton. The touching myelin objects are still problematic.

Indeed :sad. But i'm fairly confident that if we use the uncertainty output of axon segmentation we will be able to obtain a robust axon identification (by thresholding axon predictions with low uncertainty). But for that we'll have to wait for the connection with ivadomed

mariehbourget:
> But i'm fairly confident that if we use the uncertainty output of axon segmentation we will be able to obtain a robust axon identification (by thresholding axon predictions with low uncertainty).

And we have to keep in mind that the feature will be used to fill axons from a manual mask (without uncertainty available).
Thanks @Stoyan-I-A for the investigation.

Stoyan-I-A:
Since this PR is old, I have created #585, which is the same implementation but is using the most up-to-date version of ADS. I am closing this PR.

PR #405, opened 2020-11-26
Add Python 3.8 compatibility

### LINKED ISSUE #403: Update ADS to be compatible with python 3.8
ADS is currently supported only with python 3.7.
We would like ADS to be compatible with python 3.8 as well in preparation for the integration with ivadomed (which supports 3.6, 3.7 and 3.8).

mathieuboudreau:
Resolves #403 

WIP
* [x] Add Python 3.8 builds to Travis tests.
* [ ] Resolve fails that arrise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,5 +54,49 @@ jobs:
         - yes | conda install -c conda-forge fsleyes=0.33.1
         - yes | conda install -c conda-forge h5py=2.10.0
         - py.test . -v --cov AxonDeepSeg/ --cov-report term-missing
+    - stage: Basic install
+      os: linux
+      script:
+        - yes | conda create -n ads_venv python=3.7
+        - conda activate ads_venv
+        - echo "Installation of the repository"
+        - cd ~/build/neuropoly/axondeepseg
+        - pip install -e .
+        - pip install python-coveralls  # Add me to install python-coveralls
+        - py.test . -v --cov AxonDeepSeg/ --cov-report term-missing
+    - stage: FSLeyes install
+      os: linux
+      script: 
+        - yes | conda create -n ads_venv python=3.7
+        - conda activate ads_venv
+        - echo "Installation of the repository"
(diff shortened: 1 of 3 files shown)


mathieuboudreau:
Ah yes, now I remember why I didn't include Python 3.8 compatibility last time,

Capture d’écran 2020-11-26 à 19 16 09

Tensorflow 1.13.x isn't compatible with it.

We'll need to upgrade to at least Tensorflow 2.2, see https://www.tensorflow.org/install/pip

However if I recall correctly, moving to Tensorflow 2.x is incompatible with the API of our current Keras version, and upgrading that is going to be complicated if I recall our last conversation with @vs74 about this correctly from last year.

All this to say, @mariehbourget I think this PR might take a while and we'll need to completely retrain all our models again after it's complete.

mathieuboudreau:
We'll also need to stop using bireli, as moving up Tensorflow versions means we'll need too change CUDA version too

mathieuboudreau:
p.s. big picture-wise, is it even worth it spending a few days/weeks on upgrading to Tensorflow 2.2 and the accompanying  new Keras API if we're not going to be using any of it when moving to IVADOMED, since it uses PyTorch and not Tensorflow? Maybe we should just deal with the Python 3.8 upgrade whenever we're ready to get rid of Tensorflow and move to IVADOMED? @jcohenadad thoughts?

jcohenadad:
> p.s. big picture-wise, is it even worth it spending a few days/weeks on upgrading to Tensorflow 2.2 and the accompanying new Keras API if we're not going to be using any of it when moving to IVADOMED, since it uses PyTorch and not Tensorflow? Maybe we should just deal with the Python 3.8 upgrade whenever we're ready to get rid of Tensorflow and move to IVADOMED? @jcohenadad thoughts?

agreed, not worth it

mathieuboudreau:
Ok great, I'll close this PR now and we'll handle it when making the move to IVADOMED.

Issue #406, opened 2020-11-27
Move repo from travis-ci.org


alexfoias:

mathieuboudreau:
Done. It's now on travis-ci.com

PR #407, opened 2020-11-27
Travis Migration test

mathieuboudreau:
Just testing to see if the migration worked and the configuration was well done, per #406.


mathieuboudreau:
Ok the builds work on Travis-ci.com, and when the tests pass it sends that info back to the PR so approve the checks. Looks like the migration is done.

Issue #408, opened 2020-12-03
Migration to GH Actions from Travis-CI for testing 


vs74:
The purpose of this issue is to start a discussion on whether to migrate to GH actions for testing purpose as done in other projects such as SCT and Ivadomed. 

Lately, we've been facing issues with travis-CI. After migrating to [Travis.com ](https://github.com/neuropoly/axondeepseg/pull/407)from Travis.org, the PR builds are stuck and a message is being displayed: **"Expected — Waiting for status to be reported"**
-->See here: 
![image](https://user-images.githubusercontent.com/31419210/101020425-82c03a80-3594-11eb-988c-a513a483c6ae.png)

Furthermore, Travis-CI.com is a paid service, which could work for now, but in the near future it could pose problems such as only limited number of builds allowed per day in the free version. 

Since the builds are not passing, we should prioritize on this issue. @mathieuboudreau is there a quick fix on travis-ci.com builds? If there is, that would be great, but still we should have a discussion on whether we want to move to GH actions or  still want to stick to Travis-CI. 

If there is not a quick fix, then we should prioritize this issue.

mathieuboudreau:
Is it just old PRs that have this issue? The settings are exactly as they should be, and they worked for my initial test one. See: https://travis-ci.community/t/known-issue-travis-ci-reports-expected-waiting-for-status-to-be-reported-on-the-github-status-api-but-the-status-never-arrives/1154 and  https://github.com/neuropoly/axondeepseg/pull/407.

If it's only affecting old PRs that were initially built on travis-ci.org, I could override and approve them for merging.

Let's test on a new branch a PR and see if travis-ci.com returns the proper status.

mathieuboudreau:
Travis works as expected in new PRs: https://github.com/neuropoly/axondeepseg/pull/409

So the problem is just woth old PRs. @jcohenadad this might be relevant info for your other projects.

I'll force approve merges of old PRs, just tag me when they are ready.

PR #409, opened 2020-12-03
Test change for #408

mathieuboudreau:
Simply testing if the travis only fails to return the pass status on old PRs or on newer ones too.


mathieuboudreau:
Finished doing the test. See discussion in #408

PR #413, opened 2020-12-04
Fix notebook links in doc

### LINKED ISSUE #412: Links broken for Jupyter Notebooks in RTD
The links in RTD for Jupyter Notebooks have not been updated since the cleaning and renaming of the notebooks in #226.
I'll open a PR to make the changes.

mariehbourget:
Fixes #412

### DIFF
--- a/docs/source/documentation.rst
+++ b/docs/source/documentation.rst
@@ -356,19 +356,19 @@ Jupyter notebooks
 
 Here is a list of useful Jupyter notebooks available with AxonDeepSeg:
 
-* `getting_started.ipynb <https://github.com/neuropoly/axondeepseg/blob/master/notebooks/getting_started.ipynb>`_: 
-    Notebook that shows how to perform axon and myelin segmentation of a given sample using a Jupyter notebook (i.e. not using the command line tool of AxonDeepSeg). You can also launch this specific notebook without installing and/or cloning the repository by using the `Binder link <https://mybinder.org/v2/gh/neuropoly/axondeepseg/master?filepath=notebooks%2Fgetting_started.ipynb>`_.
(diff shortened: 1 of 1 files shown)


mathieuboudreau (approved):
Thanks for catching this @mariehbourget !

I just checked all the links, and they all point to the correct notebook.

However, I ran the Getting Started notebook, and it is broken in MyBinder because the model wasn't downloaded. Can you check locally on your side?

Capture d’écran 2020-12-04 à 11 54 55

Capture d’écran 2020-12-04 à 11 55 09

Regardless, this PR works and I'll approve. We should investigate the other issue in another PR.

mariehbourget:
> However, I ran the Getting Started notebook, and it is broken in MyBinder because the model wasn't downloaded. Can you check locally on your side?

Oups, I had not thought about that!
I have the same error as you. Since the model is not in the repo anymore, I don't think it can work with that notebook. Maybe we could try adding the download in another notebook specific to MyBinder? Or remove that option?

I suggest to remove it for now and open an issue dedicated to bring it back, what do you think?

mathieuboudreau:
> > However, I ran the Getting Started notebook, and it is broken in MyBinder because the model wasn't downloaded. Can you check locally on your side?
> 
> Oups, I had not thought about that!
> I have the same error as you. Since the model is not in the repo anymore, I don't think it can work with that notebook. Maybe we could try adding the download in another notebook specific to MyBinder? Or remove that option?
> 
> I suggest to remove it for now and open an issue dedicated to bring it back, what do you think?

Let's see if it can be quickly fixed, and if it can't then we'll remove it from the doc.

mathieuboudreau:
@mariehbourget I found a solution, and resolved it in #415. We can go ahead and merge this PR as-is.

mariehbourget:
Thanks @mathieuboudreau!

PR #430, opened 2020-12-19
Update issue templates

nekohayo:
This pull request replaces PR 429 by a new one with a manual commit without GitHub messing it up. Also a branch name I can reuse.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,21 +1,22 @@
 ---
 name: Bug report
-about: Create a report to help us improve
+about: Create a report to help us fix a software bug
 title: ''
 labels: category:BUG
 assignees: ''
 
 ---
+<!-- Use this form to report software defects (inaccurate/erratic behavior, crashes, etc.).
+If you have questions about using the software or are encountering difficulties with installation, please create a topic in the forum (see the "Discussions" tab at the top of this page). -->
 
 **Describe the bug**
-<!-- This form is used to report software defects such as inaccurate/erratic behavior, crashes, etc. If you have general questions about using the software or are encountering difficulties with installation, please create a topic in the forum (see the "Discussions" tab at the top of this page). -->
 A clear and concise description of what the bug is.
 Feel free to add screenshots to illustrate the problem!
 
 **To Reproduce**
 <!--
 Please try to reproduce your issue using "./AxonDeepSeg/data_test" as inputs, and to provide a sequence of commands that can reproduce it.
-If this is not possible, try to isolate a minimal input on which the issue happens (eg. one file among a dataset), and provide this file publicly, or if not possible, privately (coordinate with @jcohenadad).
+If this is not possible, try to isolate a minimal input on which the issue happens (e.g. one file among a dataset) and provide this file publicly. If the data can't be shared publicly, please coordinate with @jcohenadad in private.
 -->
 Steps to reproduce the behavior:
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -8,15 +8,22 @@ assignees: ''
 ---
 
 **Describe the bug**
+<!-- This form is used to report software defects such as inaccurate/erratic behavior, crashes, etc. If you have general questions about using the software or are encountering difficulties with installation, please create a topic in the forum (see the "Discussions" tab at the top of this page). -->
 A clear and concise description of what the bug is.
+Feel free to add screenshots to illustrate the problem!
 
 **To Reproduce**
+<!--
+Please try to reproduce your issue using "./AxonDeepSeg/data_test" as inputs, and to provide a sequence of commands that can reproduce it.
+If this is not possible, try to isolate a minimal input on which the issue happens (eg. one file among a dataset), and provide this file publicly, or if not possible, privately (coordinate with @jcohenadad).
+-->
 Steps to reproduce the behavior:
+
 1. Go to '...'
 2. Click on '....'
 3. Scroll down to '....'
-4. See error
+4. See "Error 1955: Out of Time"
 
-**Desktop (please complete the following information):**
- - OS: [e.g. iOS]
- - Version [e.g. 22]
+**Platform (please complete the following information):**
+ - OS: [e.g. Mac OS "Flaming Flamingo", Fedora 99, Ubuntu 48.04]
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -8,15 +8,22 @@ assignees: ''
 ---
 
 **Describe the bug**
+<!-- This form is used to report software defects such as inaccurate/erratic behavior, crashes, etc. If you have general questions about using the software or are encountering difficulties with installation, please create a topic in the forum (see the "Discussions" tab at the top of this page). -->
```
joshuacwnewton, line 11: I'm worried the nudge to go to the Discussions page might be too subtle. Especially because, if someone is thinking of creating an issue for a non-bug question, they'll arrive [here](https://github.com/neuropoly/axondeepseg/issues/new/choose) first:

Screenshot from 2020-12-19 09-56-16

They might just choose "Open a blank issue" and never even see this little note here about Discussions.

What do you think about creating a third (dummy) issue template called "Everything else" or "Questions and concerns" or something like that, then in the description, include a link to the Discussions page? Basically to say "don't open a blank issue, go here instead", to fill the gaps in GitHub's UI design. :wink:

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -1,14 +1,18 @@
...
+about: Describe a well-defined idea of a how a new functionality should work
 title: ''
 labels: category:FEATURE
 assignees: ''
 
 ---
 
-**Is your feature request related to a problem? Please describe.**
-A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+<!-- Is your feature request not fully defined yet, are there pending questions in how to address the problem, or are you simply looking to find a way to do something with the application? Use the "Discussions" tab at the top of this page to ask your questions or brainstorm ideas, before filing a feature request. Feature requests should be ideas that have been precisely fleshed out.
```
joshuacwnewton, line 10: Small clarity nitpick.

```suggestion
<!-- Is your feature request not fully defined yet? Are there pending questions in how to address the problem? Are you simply looking to find a way to do something with the application? If any of these are the case, please use the "Discussions" tab at the top of this page to ask your questions or brainstorm ideas before filing a feature request. Feature requests should be ideas that have been precisely fleshed out.
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -8,15 +8,22 @@ assignees: ''
...
 A clear and concise description of what the bug is.
+Feel free to add screenshots to illustrate the problem!
 
 **To Reproduce**
+<!--
+Please try to reproduce your issue using "./AxonDeepSeg/data_test" as inputs, and to provide a sequence of commands that can reproduce it.
+If this is not possible, try to isolate a minimal input on which the issue happens (eg. one file among a dataset), and provide this file publicly, or if not possible, privately (coordinate with @jcohenadad).
```
joshuacwnewton, line 18: Small clarity nitpick.

```suggestion
If this is not possible, try to isolate a minimal input on which the issue happens (e.g. one file among a dataset) and provide this file publicly. If the data can't be shared publicly, please coordinate with @jcohenadad in private.
```

joshuacwnewton (commented):
Linking ivadomed/ivadomed#565, as it's similar work, but for Ivadomed.

(Not to suggest they should be identical of course, but it might be helpful for sharing ideas across projects. :slightly_smiling_face:)

.github/ISSUE_TEMPLATE/bug_report.md
jcohenadad: hum... in general i'm reluctant to offer ppl [too many choices](https://www.ted.com/talks/barry_schwartz_the_paradox_of_choice?language=en). I already find that 2 is too much. Also, too much text scares ppl. I would personally go with something like: 
- one button (bug/feature request)
- when ppl click there, there is one big message that says: if you are a user seeking help, please use the discussion forum instead --> URL)

joshuacwnewton: Regarding @mathieuboudreau's suggestion in Slack: 

> But I wouldn’t overthink the issues template too much; let’s just get something simple out there and see if it’s effective

I think I agree, and so maybe disregard my suggestion. _(One vs. two vs. three templates seems like something to check back in on, once we've seen how people treat the choices in practice. :stuck_out_tongue_closed_eyes:)_

joshuacwnewton: Ahh, good news. hopped over to scikit-learn and noticed they had found a way to include a link option in the "Issue Select" page. They configure this here: https://github.com/scikit-learn/scikit-learn/blob/master/.github/ISSUE_TEMPLATE/config.yml

@jcohenadad we can also use this for SCT, etc. to redirect users to the spinalcordmri forum.
Screenshot from 2020-12-23 10-34-22

jcohenadad: this is *awesome* 🙌
thank you for digging @joshuacwnewton

nekohayo: I'm working on a new commit that tries to address "everything" in one go; I'll push to this branch for review today

nekohayo:
I have addressed (I think) all your comments in the 2nd commit on top, which forces people to either go through the templates or to use the custom link to the forum. Those changes are theoretical, I have no way to know if/how they will work out in the GitHub UI in practice, but I think they solve the issue.

joshuacwnewton (approved):
LGTM! Thanks much for the contribution. :slightly_smiling_face: 

(Although, I work mostly on SCT, so I'm not sure I should be giving merge approval for ADS PRs. Does this need @mathieuboudreau's approval? I'm probably overthinking things... :stuck_out_tongue_closed_eyes:)

jcohenadad:
> (Although, I work mostly on SCT, so I'm not sure I should be giving merge approval for ADS PRs. Does this need @mathieuboudreau's approval? I'm probably overthinking things... 😝)

you're part of the neuropoly team at large, so yes all your approvals are valid 😊

mathieuboudreau:
> > (Although, I work mostly on SCT, so I'm not sure I should be giving merge approval for ADS PRs. Does this need @mathieuboudreau's approval? I'm probably overthinking things... 😝)
> 
> you're part of the neuropoly team at large, so yes all your approvals are valid 😊

+1, and thanks for reviewing this PR while I was on holiday !

PR #448, opened 2021-02-01
Use pathlib in the ads_plugin instead of os.path

### LINKED ISSUE #442: FSLeyes: For path handling, use Pathlib rather than os.path
The ADS codebase uses `pathlib` to handle paths see [here](https://github.com/neuropoly/axondeepseg/wiki/How-to-handle-paths). The FSLeyes plugin, however, handles path using `os.path` module. To maintain consistency across the codebase, we should refactor how paths are handled in the ADS [FSLeyes plugin](https://github.com/neuropoly/axondeepseg/blob/master/ads_plugin.py).

Stoyan-I-A:
Fixes #442


Stoyan-I-A:
This PR is ready for review

vs74 (changes requested):
os.path is being still used [here](https://github.com/vs74/axondeepseg/blob/24080135f194f915b465a5f3a9673e09c614e56c/ads_plugin.py#L958)

ads_plugin.py
```diff
@@ -434,14 +429,15 @@ def on_save_segmentation_button(self, event):
 
         # Save the arrays as PNG files
         myelin_and_axon_array = (myelin_array // 2 + axon_array).astype(np.uint8)
-        ads_utils.imwrite(filename=save_dir + "/ADS_seg.png", img=myelin_and_axon_array)
-        ads_utils.imwrite(filename=save_dir + "/ADS_seg-myelin.png", img=myelin_array)
-        ads_utils.imwrite(filename=save_dir + "/ADS_seg-axon.png", img=axon_array)
+        ads_utils.imwrite(filename=(save_dir / "ADS_seg.png"), img=myelin_and_axon_array)
```
vs74, line 432: Either remove the brackets in the filename argument or keep the brackets for myelin and axon mask as well .

vs74:
Thanks @Stoyan-I-A for fixing the issue. I manually tested all the FSLeyes functionalities and everything works as expected. 
There are some minor changes which I've requested. Once these are fixed, this PR is good to go.

vs74:
**NOTE:** This comment is not related to this PR but in general about the master branch. 

While I was testing **Fill Axons**  button in the control panel of FSLeyes on our default SEM model's test data, I encountered a weird behavior.

1. First, I loaded the image(`AxonDeepSeg/models/default_SEM_model/data_test/image.png`) in FSLeyes using **Load PNG or TIF image** button.
2. Then, I segmented the SEM image using **Apply ADS prediction model** button ( I selected the modality as SEM).
3. Afterwards, I filled the axons using the **Fill Axons** button.

Here is the image (**image_seg-axon_cor**r) I got 
image

Is it normal??

On the other hand, when I tested for our default TEM model's test data, I got a corrected axon mask.

mariehbourget:
> Is it normal ??

That will happen if an "axon" pixel is outside of myelin borders in the segmentation.
In our cases, with our default models and data test, it happens for SEM but not for TEM.
The behavior ans way to correct it are documented [here](https://docs.google.com/document/d/1S8i96cJyWZogsMw4RrlQYwglcOWd3HrM5bpTOJE4RBQ/edit#heading=h.ddyep65c2yp) in our manual labelling procedure with FSLeyes.

Stoyan-I-A:
About the `Fill axon button` , I opened a PR (https://github.com/neuropoly/axondeepseg/pull/402) where I implemented another way to fill the axons, which in my opinion is better than the current feature.

ads_plugin.py
Stoyan-I-A: Done!

vs74 (changes requested):
I suggest removing `os` import since we have shifted to Pathlib. https://github.com/neuropoly/axondeepseg/blob/c98f882d46808dda441a5e03e8a390cbe169de18/ads_plugin.py#L17

vs74:
I retested this PR manually. All the functionalities look good to me. Before merging this PR, there is a small change I've requested. @mathieuboudreau if you wish you can also manually test the functionalities of FSLeyes before merging this PR.

Stoyan-I-A:
The `os` import is now removed.

vs74 (approved):
LGTM. Thanks, @Stoyan-I-A.

mathieuboudreau (approved):
The plugin works as expected after this refactoring, and the code looks good. Thanks a lot @Stoyan-I-A !

PR #453, opened 2021-02-04
Add rescaling in data augmentation

### LINKED ISSUE #331: "Rescaling" variable unused in data augmentation
This was first discussed in #326.

Affine scaling of the input image is not used anymore in data augmentation.
The associated variable "rescaling" still appears but is not used in train_network.py,  the "da-1-rescaling-activate" key also appears in config_files and in the 02-training_guidelines jupyter notebook.

If is it not included in elastic transfo, we should consider using affine scaling in data augmentation again.
If we don't want to use it, we should remove the unused variable and config file key in jupyter notebook.

mariehbourget:
Fixes #331 

This PR aims to re-integrate "Rescaling" in data augmentation.
The option was still available in the notebook (`da-1-rescaling-activate`) but not used in the codebase since Keras integration.

The rescaling is between a factor of 0.8 and 1.2 (similar to ADS paper with factor between 1/1.2 and 1.2).

@vs74, I re-ordered the transforms in the AUGMENTATIONS_TRAIN to reflect the order of the parameters in the notebook. Was there specific reason for the order to be different?

I also removed the now unused parameter `da-type` in the notebook.


vs74:
> @vs74, I re-ordered the transforms in the AUGMENTATIONS_TRAIN to reflect the order of the parameters in the notebook. Was there specific reason for the order to be different?

Ordering of the data augmentation operation impacts the performance of the model. I did some tests earlier and believe that the ordering of the operation in the **train_network.py** gave the best performance. 

To validate your ordering, I think it would be a good idea to include the validation metrics (like val_accuracy, val_dice_myelin,  val_loss, and val_dice_myelin) and one prediction for the modality (SEM and TEM) before ordering and the new ordering. If your ordering gives better results, we should switch to that. 

The same is the reason why I dropped rescaling. It would be a good idea to save/ log these experiments and so if in the future we wish to make a decision regarding the data augmentation,  we can refer back to them.

vs74:
@mariehbourget I believe this PR can be halted for now until IVADOMED and ADS integration as we would be doing experimentation with data augmentation, training, and all. So, right now converting this PR to **draft** makes sense to me as we will refer to this PR while doing data augmentation.

In case, if you feel that we should add a rescaling option and change the ordering, you should add the necessary validation metrics and segmentation for this PR to be merged.

mariehbourget:
Thank you for your insights @vs74! I converted to draft as you suggested and will use this branch to test rescaling with peripheral nerves training.

mariehbourget:
Closing this PR as the training is now done with ivadomed for ads v4.

PR #456, opened 2021-02-09
Remove old config file (`set_config`)

### LINKED ISSUE #447: Question: significance of script `set_config.py`?
Following the discussion in PR #441 regarding centralizing the suffixes' names into a single location, I thought of creating a config.py file where the hardcoded values will be stored and used by the developers whenever they need.

However, there is already a script named [set_config.py](https://github.com/neuropoly/axondeepseg/blob/master/set_config.py) in AxonDeepSeg which is not used anywhere in the project. This script contains paths of **axonseg** and **MATLAB**. May I know why in the first place this script was created? From what I can understand, it has something to do with **axonseg**, but I am not sure the exact purpose of it.

vs74:
This PR aims to remove old config file, which is not used in the AxonDeepSeg codebase. For context, refer to [#447 (Comment)](https://github.com/neuropoly/axondeepseg/issues/447#issuecomment-776125923)

Fixes #447


vs74:
This PR is ready for review.

mariehbourget (approved):

PR #474, opened 2021-03-12
Change the name of `-o` flag to `-overlap` flag

### LINKED ISSUE #460: Consider revisiting -o flag, which sounds like "output"
Currently, we use "-o" for "overlap", but intuitively it means "output". We should revisit this flag.

vs74:
## Checklist

## Description
As pointed out by Julien in the issue https://github.com/neuropoly/axondeepseg/issues/460, the `-o` flag for segmenting the images sounds like output, in this PR I've dropped the single dash overlap `-o` flag which is confusing.

## How to test this PR?
1. First, do `pip install -e .` in this branch.
2. Segment an image using changed flag name; `axondeepseg -i <image_name.png> -t <Modality type> --overlap <overlap_value>`

## Changed Documentation of this PR
https://axondeepseg--474.org.readthedocs.build/en/474/documentation.html#syntax

## Linked issues
Fixes #460

### DIFF
--- a/AxonDeepSeg/segment.py
+++ b/AxonDeepSeg/segment.py
@@ -280,7 +280,7 @@ def main(argv=None):
                                                             '   for the segmentation of current sample. \n'+
                                                             '3: Also displays the patch number being processed in the current sample.',
                                                             default=0)
-    ap.add_argument('-overlap', '--overlap', required=False, type=int, help='Overlap value (in pixels) of the patches when doing the segmentation. \n'+
+    ap.add_argument('--overlap', required=False, type=int, help='Overlap value (in pixels) of the patches when doing the segmentation. \n'+
                                                             'Higher values of overlap can improve the segmentation at patch borders, \n'+
                                                             'but also increase the segmentation time. \n'+
                                                             'Default value: '+str(default_overlap)+'\n'+
--- a/AxonDeepSeg/segment.py
+++ b/AxonDeepSeg/segment.py
@@ -280,7 +280,7 @@ def main(argv=None):
                                                             '   for the segmentation of current sample. \n'+
                                                             '3: Also displays the patch number being processed in the current sample.',
                                                             default=0)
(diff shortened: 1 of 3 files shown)


mathieuboudreau:
I *think* single dash CLI flags are usually meant for single letters. I don't see the point of using `-overlap` if we already also have `--overlap`.

Thoughts from anyone else?

vs74:
In some RTDs, I've seen the practice of single dashed even for words. That's why I added that. If it is true that single dashed are only for letters then perhaps we can remove the `-o` flag ( in CLI's of several other packages, `-o` flag usually means 'output'  so it can create confusion among the users) and keep the double dashed `--overlap` flag of the CLI.

jcohenadad:
> I think single dash CLI flags are usually meant for single letters. I don't see the point of using -overlap if we already also have --overlap.

indeed-- single letter after "-" is the POSIX convention.  I was not aware we already had "--overlap", in this case we can simply drop "-o". In general i'm in favour of getting rid of single dash single letter (becomes confusing).

For some context. It all started (in the lab, inspired by FSL) with SCT using single dash with long words, and no double-dash. We are having this discussion regularly in other projects. Recently for shimming toolbox we decided to stick to POSIX (eg double dash for long words), but avoid single dash. Few discussions:
- https://github.com/shimming-toolbox/shimming-toolbox/issues/222
- https://github.com/neuropoly/spinalcordtoolbox/issues/2939
- https://github.com/neuropoly/spinalcordtoolbox/issues/3159

mathieuboudreau:
Perfect - thank you for clarifying the new lab convention @jcohenadad !

So @vs74, I guess all that needs to be done in this PR is removing the `-o` flag and then it can be merged, correct?

vs74:
For now, I've dropped the `-o` flag, and this PR is ready for review. I've dropped the -o flag, changed the documentation added a unit test that was missing for the overlap flag. But we should open an issue too just for the case that if in future we wish to switch to single dash or double-dash entirely for CLI flags, we could refer to that issue.

mathieuboudreau (approved):
Looks great - once the test pass I'll merge!

PR #490, opened 2021-03-28
Update the documentation for installing the GPU compatible version of ADS

### LINKED ISSUE #470: Documentation: Correction on GPU compatible installation of axondeepseg
**Describe the bug**

- Current documentation on how to install GPU compatible ADS cause ADS cli to crash
- For Full log and context: https://github.com/neuropoly/axondeepseg/discussions/469#discussioncomment-429709

**Steps To Reproduce**

1. Clone ADS Repo
2. Make the conda environment (no need to setup CUDA yet to reproduce this issue)
3. ```pip install -e .``` while in the conda env. 
4. ```pip uninstall tensorflow``` as per documentation here: https://axondeepseg.readthedocs.io/en/latest/documentation.html#gpu-compatible-installation
5. ```pip install tensorflow-gpu==1.13.1```
6. Reproduce the error reported in discussion

**Potential Fix**

1. replace ```tensorflow==1.13.1``` with ```tensorflow-gpu==1.13.1``` in the requirement.txt file.  Then run  ```pip install -e .```  while in the root directory of the repo. 
2. **ReadTheDoc/Sphinx documentation ultimately needs to update to reflect the above correct steps for that section.** 
3. Long discussion about the role of ```requirements.txt``` and potential better deployment strategy discussion as started in https://github.com/neuropoly/axondeepseg/issues/327 should continue. This may become a moot point or bigger issue when IvadoMed integration starts. 

**Platform (please complete the following information):**
 - OS: [Win 10.0.19042 Build 19042]
 - AxonDeepSeg version number: [git master]

Stoyan-I-A:
## Checklist

- [X] I've given this PR a concise, self-descriptive, and meaningful title
- [X] I've linked relevant issues in the PR body
- [X] I've updated the documentation and/or added correct docstrings

## Description
Since we changed the way the environment is installed, we should also take a look at the way the GPU compatible version of ADS is installed. Currently, we suggest using pip to uninstall tensorflow and install tensorflow gpu. However, because we are now using the anaconda cloud/conda-forge to install our packages, it makes sense to use conda instead of pip to install the correct packages for using the GPU.

I have added instructions to use the GPU on Windows, but I believe the same instructions would work on Linux.

## Linked issues
Resolves #470


vs74:
I checked the instructions on a Linux system. The instructions seem to work and `Keras` does get uninstalled while installing `TensorFlow-gpu=1.13.1.` 

Here is the configuration on which I tested the updated documentation.

OS: Ubuntu 18.04
AxonDeepSeg version number: git master
CUDA: 11.0
GPU: NVIDIA A100

Stoyan-I-A:
> **CUDA:** 11.0

Thank you @vs74 for testing this. Did you also try to run `axondeepseg_test`? I it worked correctly, I suppose the CUDA version installed with anaconda (10.0) was used over the 11.0 one.

vs74:
I did try to run `axondeepseg_test` and it worked. Yes CUDA 10.0 was installed over 11.0 in the virtual environment.

docs/source/documentation.rst
```diff
@@ -155,13 +155,28 @@ GPU-compatible installation
 ---------------------------
 .. NOTE :: This feature is not available if you are using a macOS.
 
+Linux
+~~~~~~~
+
 By default, AxonDeepSeg installs the CPU version of TensorFlow. To train a model using your GPU, you need to uninstall the TensorFlow from your virtual environment, and install the GPU version of it::
 
     pip uninstall tensorflow
```
mathieuboudreau, line 163: Should we change the instructions to conda here too?

Stoyan-I-A: I think it's a good idea.

Stoyan-I-A:
In https://github.com/neuropoly/axondeepseg/pull/490/commits/bb5a8a7cc78e73ec41ae431896b78c1836ba059e, I updated the instructions for Linux. I also removed the warning about the CUDA version because Conda automatically installs the correct version.

I still need to check if installing tensorflow-gpu with conda could automatically uninstall the regular tensorflow without uninstalling keras.

Stoyan-I-A:
> I still need to check if installing tensorflow-gpu with conda could automatically uninstall the regular tensorflow without uninstalling keras.

I tested this, and unfortunately installing tensorflow-gpu with conda without uninstalling tensorflow (cpu) results in having both versions installed at the same time, and the cpu version will be used instead of the gpu one.

Therefore, won't make more changes to the documentation. 

This PR is ready for review.

vs74 (changes requested):
Looks like instructions on [02-training_guideline.ipynb ](https://github.com/neuropoly/axondeepseg/blob/sia/470/gpu_documentation/notebooks/02-training_guideline.ipynb) for running the notebook on GPU and the wiki page ([Training on Bireli](https://github.com/neuropoly/axondeepseg/wiki/Training-on-bireli)) also needs to be updated.

Stoyan-I-A:
> Looks like instructions on [02-training_guideline.ipynb ](https://github.com/neuropoly/axondeepseg/blob/sia/470/gpu_documentation/notebooks/02-training_guideline.ipynb) for running the notebook on GPU and the wiki page ([Training on Bireli](https://github.com/neuropoly/axondeepseg/wiki/Training-on-bireli)) also needs to be updated.

You're right. I suspect these instructions to change as well whenever we switch to Ivadomed. Do you think I could just add a link to the section of the documentation explaining how to use the GPU compatible version? This way, we would only have to update the documentation if the instructions ever change.

vs74:
Yes, most likely the instructions would change with ivadomed. I'm comfortable with that option. This way we don't have to maintain all the files/docs for instructions on how to access GPU with ADS.

notebooks/02-training_guideline.ipynb
```diff
@@ -11,12 +11,7 @@
     "* Train a model from scratch by defining the parameters of the network\n",
     "* Make inference using the trained model\n",
     "\n",
-    "**Important:** If you have access to a GPU card, we strongly recommend you use it. To do so, TensorFlow needs to be uninstalled and then re-installed using the GPU mode:\n",
-    "```bash\n",
-    "source activate ads_venv  # Activate virtual environment. This line might depend on your ADS installation\n",
-    "pip uninstall tensorflow\n",
-    "pip install tensorflow-gpu==1.13.1\n",
-    "```"
+    "**Important:** If you have access to a GPU card, we strongly recommend you use it. By default, AxonDeepSeg will only use the CPU. To install the GPU compatible version of AxonDeepSeg, refer to the documentation: https://axondeepseg.readthedocs.io/en/latest/documentation.html#gpu-compatible-installation"
```
vs74, line 14: The instructions look good to me.

docs/source/documentation.rst
```diff
@@ -168,12 +168,18 @@ GPU-compatible installation
 ---------------------------
 .. NOTE :: This feature is not available if you are using a macOS.
 
+Linux and Windows 10
```
vs74, line 171: Is there a specific reason for only specifying "Linux and Windows 10"? Is it because that we have tested these instructions on a GPU machine having base OS "Linux and Windows 10" only?

vs74 (commented):
Everything looks good to me. Once my questions have been addressed, I'll approve this PR. Thanks @Stoyan-I-A for taking care of the GPU installation issue.

docs/source/documentation.rst
Stoyan-I-A: I specified Linux and Windows 10 because I don't think there are many other options.
CUDA requires an Nvidia GPU and there haven't been any MacOS products with Nvidia GPUs in the last few years. Users could use an external GPU but I've never played with this kind of hardware.

vs74: I guess we shouldn't be very specific about only "Linux and Windows 10". Anyways, not a big issue, so I'm approving this PR. Everything looks good to me.

vs74 (approved):

Issue #508, opened 2021-04-19
Add integrity testing


jcohenadad:
**Describe the problem**
It would be nice to test integrity, ie. is the prediction correct?

**Proposed solution**
Add an assertion below this code:
https://github.com/neuropoly/axondeepseg/blob/e32dcfc582ab9cb0b886af57aa9075a95cf958ee/AxonDeepSeg/integrity_test.py#L61

PR #510, opened 2021-04-19
Fix black bar segmentation bug when width = height

### LINKED ISSUE #509: Black bar in segmentations
We have an issue with 4096x4096 images where the TEM segmentation produces a "black bar" of pixels at the right of the image except the corners. Upon further investigation, I found that it only happens when both the `width` and `height` of the image are ~~a multiple of `512` (patch size)~~ the same, unrelated to the content of the image.

I tried segmenting images with the native TEM pixel size of `0.01` and an overlap value of `25` pixels. In that case, the well segmented area above the black bar is `25x25` pixels, and the well segmented area below the black bar is `487x25` pixels (where 487 = 512 -25).

image

So, the issue seems to be related to image reconstruction from the patches, for all the last patches on the right except the one in the bottom right corner.

I may have found the culprit in patches2im_overlap [here](https://github.com/neuropoly/axondeepseg/blob/e32dcfc582ab9cb0b886af57aa9075a95cf958ee/AxonDeepSeg/patch_management_tools.py#L102), where `if L_pos[i][1] != h_l:` should be `if L_pos[i][0] != h_l:`.

I'll do some test and open a PR.

mariehbourget:
## Checklist

## Description
This PR fixes a bug where a black bar appears in the segmentation at the right of the image when the image width and height are the same, see #509 for details.

## To test the PR
1. On master branch, segment an image where width=height. For reference, the `Maxo03_img60` image from our SEM dataset is 2001x2001. There will be a black bar on the right in the segmentation.
2. Checkout this branch and redo the segmentation --> no black bar.
3. Segment another image where width != height --> no black bar.

I did not add a test for this fix. As discussed today, we would need to add an integrity test for the prediction first (#508).

## Linked issues
Fixes #509

### DIFF
--- a/AxonDeepSeg/patch_management_tools.py
+++ b/AxonDeepSeg/patch_management_tools.py
@@ -99,7 +99,7 @@ def patches2im_overlap(L_patches, L_pos, overlap_value=25, scw=512):
                                                                                                           -overlap_value:,
                                                                                                           overlap_value:-overlap_value]
         if L_pos[i][1] == w_l:
-            if L_pos[i][1] != h_l:
+            if L_pos[i][0] != h_l:
                 new_img[L_pos[i][0] + overlap_value:L_pos[i][0] + scw - overlap_value, -overlap_value:] = e[
                                                                                                           overlap_value:-overlap_value,
                                                                                                           -overlap_value:]


mathieuboudreau (approved):
I tested this fix using our test SEM image but resized to be exactly square, and did not experience the black bar after running the segmentation. Thanks for finding this odd bug and resolving it @mariehbourget !

image
image_seg-axonmyelin

vs74:
I tested it on Alan's image as well where @mariehbourget noticed this issue. This fix seems to work perfectly. Thanks for fixing it @mariehbourget.

# **Original image**

image



# **Segmentation**
image

vs74 (approved):

jcohenadad:
thank you @mariehbourget !

Issue #506, opened 2021-04-15
Move/Copy data from OSF and package them with git versioning


jcohenadad:
**Describe the problem**

Current issues with hosting data on OSF:
- OSF is cumbersome to use (too many clicks, specific APIs that the team needs to learn to use), 
- not compatible with git
- limited in size
- Datasets are spreadout across multiple OSF repos:
  - https://osf.io/yp4qg/
  - https://osf.io/bj9eu/
- Currently, [model packages](https://osf.io/sv7u2/) include testing data, which makes it difficult to update separately. Related to #508.

**Proposed solution**
Host all the datasets that are used for training models using git/git-annex technology. If data are private and/or too large for Github, they can be hosted on NeuroPoly's internal git server (data.neuro.polymtl.ca). If data are public and small size, they can go under example: **https://github.com/axondeepseg/data-testing**. 

Pros:
- Harmonization with current best practices in reproducible science;
- Harmonization with internal lab practice;
- Can accommodate private data;
- Can train using multiple BIDS datasets (e.g. ivadomed can take as input multiple BIDS datasets and aggregate them);
- We can easily encode in the ivadomed config file the git commit hashtag;

For inspiration: 
- https://github.com/spinalcordtoolbox/sct_testing_data
- https://github.com/ivadomed/data-testing

## Details of the strategy

- Move **test_segmentation.zip** to https://github.com/axondeepseg/data-example
- Move test files under: https://github.com/axondeepseg/data-testing
- Move models under: https://github.com/axondeepseg/XXX, with XXX being the name of the model. Create one repo per model.
Wait to move the raw/ data. They will ultimately be moved to our internal server with git-annex.

vs74:
> For inspiration: https://github.com/sct-data/

Is the repo a private one as I can't access it? It displays **404 error.**

jcohenadad:
> > For inspiration: https://github.com/sct-data/
> 
> Is the repo a private one as I can't access it? It displays **404 error.**

sorry @vasudev-sharma, the link has changed, it is now: https://github.com/spinalcordtoolbox/sct_testing_data
( I also updated the link in my original message: https://github.com/neuropoly/axondeepseg/issues/506#issue-858986482)

hermancollin:
Update:
- [x] Move test_segmentation.zip to https://github.com/axondeepseg/data-example
- [x] Move test files under: https://github.com/axondeepseg/data-testing
[...]

PR #518, opened 2021-05-24
Implement Batch morphometrics for images present in directory/directories

vasudev-sharma:
## Checklist

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've added relevant tests for my contribution
- [x] I've updated the documentation and/or added correct docstrings
- [x] I've assigned a reviewer
- [x] I've consulted [ADS's internal developer documentation](https://github.com/neuropoly/axondeepseg/wiki/Guidelines-for-contribution) to ensure my contribution is in line with any relevant design decisions

## Description
This PR aim to implement batch morphometrics of all the images located in a directory. 

**CLI command:**  `axondeepseg_morphometrics -i <directory>`

**Documentation:** https://axondeepseg--518.org.readthedocs.build/en/518/documentation.html#morphometrics-of-images-present-in-folder-s

**Test added for the feature -**> [test_main_cli_runs_successfully_for_generating_batches_morphometrics_multiple_images ](https://github.com/neuropoly/axondeepseg/blob/5f552acc81555dd2596ef17d6f4bc4db8de9436b/test/morphometrics/test_launch_morphometrics_computation.py#L125-L156)
## Linked issues

Fixes #514

PR #533, opened 2021-06-18
Download models from Github repo

hermancollin:
## Checklist

## Description

With this modification, models are now downloaded from Github instead of OSF.
I had to play around with the github API in order to get it to work properly. IMHO, it isn't very well documented and I had to find a way to communicate with the API without installing a new dependency. The `requests` module does the trick for us.
I think the way it worked with OSF was simpler, as now we have to fetch the release tag and insert it in the download URL, but at least it works and we will always get the latest version.

## How to review
You can do the following:
```bash
$ cd axondeepseg
$ py.test test/test_download_models.py
```
(or run all tests)

## Linked issues

Related to migration of repositories from OSF to Github #506 

BTW first PR here, feel free to tell me if something isn't right

### DIFF
--- a/AxonDeepSeg/download_model.py
+++ b/AxonDeepSeg/download_model.py
@@ -47,7 +47,7 @@ def download_model(destination = None):
         shutil.rmtree(str(model_seg_pns_bf_destination))
 
     shutil.move(folder_name_SEM_model.joinpath("default_SEM_model"), str(sem_destination))
-    shutil.move(folder_name_TEM_model.joinpath('default_TEM_model'), str(tem_destination))
+    shutil.move(folder_name_TEM_model.joinpath("default_TEM_model"), str(tem_destination))
     shutil.move(folder_name_OM_model.joinpath("model_seg_pns_bf"), str(model_seg_pns_bf_destination))
 
     # remove temporary folders
--- a/AxonDeepSeg/download_model.py
+++ b/AxonDeepSeg/download_model.py
@@ -1,7 +1,7 @@
 from AxonDeepSeg.ads_utils import convert_path, download_data
 from pathlib import Path
 import shutil
-import requests
+from os import listdir
 
 
 def download_model(destination = None):
@@ -15,17 +15,11 @@ def download_model(destination = None):
         tem_destination = destination / "default_TEM_model"
         model_seg_pns_bf_destination = destination / "model_seg_pns_bf"
 
-    # retrieve latest release identifier from github API
-    response_TEM = requests.get("https://api.github.com/repos/axondeepseg/default-TEM-model/releases/latest")
-    folder_name_TEM_model = Path("default-TEM-model-"+str(response_TEM.json()['tag_name']))
-    response_SEM = requests.get("https://api.github.com/repos/axondeepseg/default-SEM-model/releases/latest")
-    folder_name_SEM_model = Path("default-SEM-model-"+str(response_SEM.json()['tag_name']))
-    response_OM = requests.get("https://api.github.com/repos/axondeepseg/model-seg-pns-bf/releases/latest")
-    folder_name_OM_model = Path("model-seg-pns-bf-"+str(response_OM.json()['tag_name']))
+    url_TEM_model = "https://github.com/axondeepseg/default-TEM-model/archive/refs/tags/r20210615.zip" 
+    url_SEM_model = "https://github.com/axondeepseg/default-SEM-model/archive/refs/tags/r20210615.zip" 
+    url_model_seg_pns_bf = "https://github.com/axondeepseg/model-seg-pns-bf/archive/refs/tags/r20210615.zip"
 
-    url_TEM_model = "https://github.com/axondeepseg/default-TEM-model/archive/refs/tags/"+str(response_TEM.json()['tag_name'])+".zip" 
-    url_SEM_model = "https://github.com/axondeepseg/default-SEM-model/archive/refs/tags/"+str(response_SEM.json()['tag_name'])+".zip" 
-    url_model_seg_pns_bf = "https://github.com/axondeepseg/model-seg-pns-bf/archive/refs/tags/"+str(response_OM.json()['tag_name'])+".zip"
+    files_before = listdir()
 
     if (
         not download_data(url_TEM_model) and not download_data(url_SEM_model) and not download_data(url_model_seg_pns_bf)
@@ -35,6 +29,14 @@ def download_model(destination = None):
         print(
             "ERROR: Data was not succesfully downloaded and unzipped- please check your link and filename and try again."
         )
+
+    files_after = listdir()
+
+    # retrieving unknown model folder names
+    model_folders = list(set(files_after)-set(files_before))
+    folder_name_TEM_model = ''.join([x for x in model_folders if 'TEM' in x])
+    folder_name_SEM_model = ''.join([x for x in model_folders if 'SEM' in x])
+    folder_name_OM_model = ''.join([x for x in model_folders if 'pns-bf' in x])
 
     if sem_destination.exists():
         print('SEM model folder already existed - deleting old one.')
@@ -46,9 +48,9 @@ def download_model(destination = None):
         print('Bright Field Optical Microscopy model folder already existed - deleting old one')
         shutil.rmtree(str(model_seg_pns_bf_destination))
 
-    shutil.move(folder_name_SEM_model.joinpath("default_SEM_model"), str(sem_destination))
-    shutil.move(folder_name_TEM_model.joinpath("default_TEM_model"), str(tem_destination))
-    shutil.move(folder_name_OM_model.joinpath("model_seg_pns_bf"), str(model_seg_pns_bf_destination))
+    shutil.move(Path(folder_name_SEM_model).joinpath("default_SEM_model"), str(sem_destination))
+    shutil.move(Path(folder_name_TEM_model).joinpath("default_TEM_model"), str(tem_destination))
+    shutil.move(Path(folder_name_OM_model).joinpath("model_seg_pns_bf"), str(model_seg_pns_bf_destination))
 
     # remove temporary folders
     shutil.rmtree(folder_name_TEM_model)
--- a/AxonDeepSeg/download_model.py
+++ b/AxonDeepSeg/download_model.py
@@ -17,15 +17,15 @@ def download_model(destination = None):
 
     # retrieve latest release identifier from github API
     response_TEM = requests.get("https://api.github.com/repos/axondeepseg/default-TEM-model/releases/latest")
-    folder_name_TEM_model = Path("default-TEM-model-"+str(response_TEM.json()["tag_name"]))
+    folder_name_TEM_model = Path("default-TEM-model-"+str(response_TEM.json()['tag_name']))
     response_SEM = requests.get("https://api.github.com/repos/axondeepseg/default-SEM-model/releases/latest")
-    folder_name_SEM_model = Path("default-SEM-model-"+str(response_SEM.json()["tag_name"]))
+    folder_name_SEM_model = Path("default-SEM-model-"+str(response_SEM.json()['tag_name']))
     response_OM = requests.get("https://api.github.com/repos/axondeepseg/model-seg-pns-bf/releases/latest")
-    folder_name_OM_model = Path("model-seg-pns-bf-"+str(response_OM.json()["tag_name"]))
+    folder_name_OM_model = Path("model-seg-pns-bf-"+str(response_OM.json()['tag_name']))
 
-    url_TEM_model = "https://github.com/axondeepseg/default-TEM-model/archive/refs/tags/"+str(response_TEM.json()["tag_name"])+".zip" 
-    url_SEM_model = "https://github.com/axondeepseg/default-SEM-model/archive/refs/tags/"+str(response_SEM.json()["tag_name"])+".zip" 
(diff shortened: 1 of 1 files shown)


AxonDeepSeg/download_model.py
```diff
@@ -14,9 +15,17 @@ def download_model(destination = None):
         tem_destination = destination / "default_TEM_model"
         model_seg_pns_bf_destination = destination / "model_seg_pns_bf"
 
-    url_TEM_model = "https://osf.io/2hcfv/?action=download&version=5"  # URL of TEM model hosted on OSF storage with the appropriate versioning on OSF
-    url_SEM_model = "https://osf.io/sv7u2/?action=download&version=5"  # URL of SEM model hosted on OSF storage with the appropriate versioning on OSF
-    url_model_seg_pns_bf = "https://osf.io/hfeu2/?action=download&version=2" # URL of Bright Field Optical Microscopy model hosted on storage with the appropriate versioning on OSF
+    # retrieve latest release identifier from github API
+    response_TEM = requests.get("https://api.github.com/repos/axondeepseg/default-TEM-model/releases/latest")
+    folder_name_TEM_model = Path("default-TEM-model-"+str(response_TEM.json()['tag_name']))
```
vasudev-sharma, line 20: Thanks @hermancollin for creating your first PR 🎉 in ADS. 

I've a question: is it not possible to directly download the latest release without extracting the tag name? 

I was able to extract the asset zip without having  the tag name in my PR see here. https://github.com/neuropoly/axondeepseg/pull/518#issuecomment-863351415

Likewise, I was able to download the latest release for **SEM**, **TEM** and **OM** models

1. https://api.github.com/repos/axondeepseg/default-SEM-model/zipball
2.  https://api.github.com/repos/axondeepseg/default-TEM-model/zipball
3. https://api.github.com/repos/axondeepseg/model-seg-pns-bf/zipball

AxonDeepSeg/download_model.py
```diff
@@ -14,9 +15,17 @@ def download_model(destination = None):
...
+    response_SEM = requests.get("https://api.github.com/repos/axondeepseg/default-SEM-model/releases/latest")
+    folder_name_SEM_model = Path("default-SEM-model-"+str(response_SEM.json()['tag_name']))
+    response_OM = requests.get("https://api.github.com/repos/axondeepseg/model-seg-pns-bf/releases/latest")
+    folder_name_OM_model = Path("model-seg-pns-bf-"+str(response_OM.json()['tag_name']))
+
+    url_TEM_model = "https://github.com/axondeepseg/default-TEM-model/archive/refs/tags/"+str(response_TEM.json()['tag_name'])+".zip" 
+    url_SEM_model = "https://github.com/axondeepseg/default-SEM-model/archive/refs/tags/"+str(response_SEM.json()['tag_name'])+".zip" 
+    url_model_seg_pns_bf = "https://github.com/axondeepseg/model-seg-pns-bf/archive/refs/tags/"+str(response_OM.json()['tag_name'])+".zip"
```
vasudev-sharma, line 28: I was thinking that perhaps we should store the url in [config.py](https://github.com/neuropoly/axondeepseg/blob/master/config.py) script rather than hardcoing in this script. By doing so, all the hardcoded strings will be centralized in a single location. What do you think?

hermancollin: That makes a lot of sense. I will incorporate this change right away!

AxonDeepSeg/download_model.py
hermancollin: Yes, I agree with you that the method I suggested looks a bit tedious. The problem with these links is that they are not compatible with the `ads_utils.download_data()` function. I played around with variations of these URLs a lot and what I found is that we always get the same problem: the repo is zipped inside another folder, so when the `download_data` function retrieves the content from the URL, we need to know the name of the highest-level folder.

The URLs you suggested give us the following:
Screenshot_20210621_115911
As you can see, the folder name contains the first 6 characters from the commit hash, which is slightly more complicated to retrieve than the tag_name. I know this isn't ideal, but I haven't found a better way to deal with this.

I had another idea to deal with this problem. When creating the release, we can manually add the compressed file in the assets and this would be easier to retrieve directly than the default released zip file. However, this workflow is really not ideal because whenever we will want to change something in the repo, we'll have to create a new release and manually add the compressed file everytime (also the name of the added asset would need to be the same for every release to be compatible with a hardcoded asset name).

AxonDeepSeg/download_model.py
hermancollin: In fact I'll wait to close our discussion above (concerning the actual URLs we want to keep) before commiting these changes.

AxonDeepSeg/download_model.py
vasudev-sharma: > I had another idea to deal with this problem. When creating the release, we can manually add the compressed file in the assets and this would be easier to retrieve directly than the default released zip file. However, this workflow is really not ideal because whenever we will want to change something in the repo, we'll have to create a new release and manually add the compressed file everytime (also the name of the added asset would need to be the same for every release to be compatible with a hardcoded asset name).

I agree with you. Thats not the best way to go ahead with. However, I wish to know how does `ivadomed` deals with data and releases on GH. @jcohenadad any ideas? 

Your first solution makes sense and I'm favour of it.

jcohenadad: > I've a question: is it not possible to directly download the latest release without extracting the tag name?

We definitely want to keep the tag name in the URL, in order to ensure cross-compatibility. Eg: a user would like to analyse a dataset using a version of ADS that is 2 years old: the _latest_ models would be different from that one 2 years ago, and the users would not be able to reproduce results.

The way we do it in SCT is by pointing to the tagged URL (example [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/scripts/sct_download_data.py#L53)), downloading/unzipping in a temporary folder, and then copy the model/data in a folder that always has the same name (eg: model-tem). 

> I had another idea to deal with this problem. When creating the release, we can manually add the compressed file in the assets and this would be easier to retrieve directly than the default released zip file.

Every idea that has the name "manually" in it is a bad idea 😉

hermancollin: > The way we do it in SCT is by pointing to the tagged URL (example [here](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/scripts/sct_download_data.py#L53)), downloading/unzipping in a temporary folder, and then copy the model/data in a folder that always has the same name (eg: model-tem).

Should we put the links in the `config.py` script, as suggested by @vasudev-sharma or should we put them directly in the `download_models.py` script (like the *SCT way*)?

jcohenadad: hum... i find it less cumbersome from a developer standpoint to put the script URL where it is used (ie: in download_models.py).

mariehbourget: I also prefer to keep the links in `download_model.py`, this is where I would intuitively look for them.

AxonDeepSeg/download_model.py
```diff
@@ -36,6 +30,14 @@ def download_model(destination = None):
             "ERROR: Data was not succesfully downloaded and unzipped- please check your link and filename and try again."
         )
 
+    files_after = listdir()
+
+    # retrieving unknown model folder names
+    model_folders = list(set(files_after)-set(files_before))
```
hermancollin, line 36: This is the simplest way I found to retrieve the folder names automatically after the download.
Ideally, if the actual download wasn't processed by the `download_data` utility, retrieving the folder names would be trivial.

vasudev-sharma: I would personally use **Pathlib's** `iterdir()` function than **os** module` listdir() `function. Although both of have the same functionality, the reason for having preferance to `pathlib` module is that ADS uses `pathlib` to handle paths ( see [here](https://github.com/neuropoly/axondeepseg/wiki/How-to-handle-paths)) and it will maintain consistency in the codebase.

vasudev-sharma (commented):
I've added some comments. Let me know what do you think about them.

AxonDeepSeg/download_model.py
```diff
@@ -14,9 +15,11 @@ def download_model(destination = None):
         tem_destination = destination / "default_TEM_model"
         model_seg_pns_bf_destination = destination / "model_seg_pns_bf"
 
-    url_TEM_model = "https://osf.io/2hcfv/?action=download&version=5"  # URL of TEM model hosted on OSF storage with the appropriate versioning on OSF
-    url_SEM_model = "https://osf.io/sv7u2/?action=download&version=5"  # URL of SEM model hosted on OSF storage with the appropriate versioning on OSF
-    url_model_seg_pns_bf = "https://osf.io/hfeu2/?action=download&version=2" # URL of Bright Field Optical Microscopy model hosted on storage with the appropriate versioning on OSF
+    url_TEM_model = "https://github.com/axondeepseg/default-TEM-model/archive/refs/tags/r20210615.zip" 
+    url_SEM_model = "https://github.com/axondeepseg/default-SEM-model/archive/refs/tags/r20210615.zip" 
+    url_model_seg_pns_bf = "https://github.com/axondeepseg/model-seg-pns-bf/archive/refs/tags/r20210615.zip"
```
vasudev-sharma, line 20: Is it retrieving the latest release? I believe hardcoding the **tag** name with the urlwould make us change those links in this script whenever we have a new release.

hermancollin: No, this does not retrieve the latest release, but it is the behaviour we chose to ensure cross-compatibility and mostly reproducibility (e.g. if we want to update the models, it will require a PR on the ADS side)

AxonDeepSeg/download_model.py
hermancollin: Thanks for the link! I was looking for that exact document yesterday

vasudev-sharma (approved):
Everything works as expected. Thanks @hermancollin for your first and an important contribution. Since we have shifted the models to OSF, I've a small change request which has to made in WIKI ( see [here](https://github.com/neuropoly/axondeepseg/wiki/Instructions-on-How-to-Update-Models-in-Repo-and-OSF-Storage))

jcohenadad (approved):
Great contribution! @hermancollin 🏅

hermancollin:
> Since we have shifted the models to OSF, I've a small change request which has to made in WIKI ( see [here](https://github.com/neuropoly/axondeepseg/wiki/Instructions-on-How-to-Update-Models-in-Repo-and-OSF-Storage))

You are right I missed this. I updated the wiki for the new instructions. Also updated [this template for models READMEs](https://github.com/neuropoly/axondeepseg/wiki/Template-for-Readme.txt) which had instructions for upload on OSF.

@vasudev-sharma @jcohenadad if everything is OK with those updates, I'll merge this branch

jcohenadad:
> > Since we have shifted the models to OSF, I've a small change request which has to made in WIKI ( see [here](https://github.com/neuropoly/axondeepseg/wiki/Instructions-on-How-to-Update-Models-in-Repo-and-OSF-Storage))
> 
> You are right I missed this. I updated the wiki for the new instructions. Also updated [this template for models READMEs](https://github.com/neuropoly/axondeepseg/wiki/Template-for-Readme.txt) which had instructions for upload on OSF.
> 
> @vasudev-sharma @jcohenadad if everything is OK with those updates, I'll merge this branch

suggestion:
~~~
Steps to train this model
=========================

1. Get ADS version: [Enter the git commit hashtag that was used to train the model]
2. Get the data: [Enter the github URL pointing to the data-- make sure to include the proper tag name]
3. Run the notebook "guide_dataset_building.ipynb" to perform a split into training and validation set. 
4. Go to "training_guideline.ipynb" notebook and make necessary changes required in paths, URL’s and config. Finally run all the cells of the notebook.
5. After successfully running all the cells, the trained model file will be saved under the directory: AxonDeepSeg/axondeepseg/models/
6. Name the model according to this convention: https://github.com/ivadomed/ivadomed/wiki/repositories

Upload on Github
================
1. Create a repository on Github's ADS organization (https://github.com/axondeepseg). Name the repository as: model-XXX, with XXX being the name of the model (see step 6 above). 
2. Add this README.txt file alongside the model and upload the uncompressed folder on the Github repository. 
3. Create a new release. The creation of a release generates a ZIP package of the model. 
4. Copy the URL of the ZIP package and add it to ADS's function download_model.py. 
~~~

hermancollin:
> suggestion:
> 
> ```
> Steps to train this model
> =========================
> 
> 1. Get ADS version: [Enter the git commit hashtag that was used to train the model]
> 2. Get the data: [Enter the github URL pointing to the data-- make sure to include the proper tag name]
> 3. Run the notebook "guide_dataset_building.ipynb" to perform a split into training and validation set. 
> 4. Go to "training_guideline.ipynb" notebook and make necessary changes required in paths, URL’s and config. Finally run all the cells of the notebook.
> 5. After successfully running all the cells, the trained model file will be saved under the directory: AxonDeepSeg/axondeepseg/models/
> 6. Name the model according to this convention: https://github.com/ivadomed/ivadomed/wiki/repositories
> 
> Upload on Github
> ================
> 1. Create a repository on Github's ADS organization (https://github.com/axondeepseg). Name the repository as: model-XXX, with XXX being the name of the model (see step 6 above). 
> 2. Add this README.txt file alongside the model and upload the uncompressed folder on the Github repository. 
> 3. Create a new release. The creation of a release generates a ZIP package of the model. 
> 4. Copy the URL of the ZIP package and add it to ADS's function download_model.py. 
> ```

I edited the template and used your suggestion instead.

I think everything is ready for merging :tada:

Issue #523, opened 2021-06-07
AxonDeepSeg v4


mathieuboudreau:
@mariehbourget, @vasudev-sharma, and I just met to get the discussion started on thoughts/strategy for the version 4 release of AxonDeepSeg, which will be a major refactor that will use IVADOMED. Feel free to edit or write questions below.

## Preliminary work
* Reseach IVADOMED/ADS overlap/differences
  * Make a two-column table that lists/identifies ADS functions with their equivalent IVADOMED functions
    * [GDoc File](https://docs.google.com/document/d/1Ipd6_nK6I6CKlDI_KpkLXH_jY2C547W_ootalfKnFlU/edit?usp=sharing) 
* Leave ADS’s API as unchanged as possible
* List all main ADS user entrypoints & options
  * Command line
  * FSLeyes
  * Jupyter Notebook
* Tests
  * Identify tests that will have to be completely deleted (entire functionality is replaced by IVADOMED)
  * Identify tests that should be untouched (API, or files that have no overlap with IVADOMED).
  * Identify tests that will need to be modified for them to work with the new ADS version.
* Models/Segmentations
  * Create models using IVADOMED for all of our current models

## Repository

* Create a new branch (v4) that will be the “master branch” of all these changes, and will be what is merged for release
* Ramp down PRs on our current master branch
  * Only essential PRs should be merged or worked on (e.g. collaborations, bugs)
  * Ideally, these would also be merged somewhere too in our v4 tree (or worked on), if compatible.

## Task flow

* First identify all the API that will need to be untouched, meaning that we might need to keep some functions that wrap around other code or calls IVADOMED?
* Delete files and functions that have overlap with IVADOMED but not required for our API (delete their tests too).
  * See if any of the remaining test fail after this step, and plan accordinlgy
* Refactor files that have overlap with ivadomed, but are required for our API.
  * Ensure tests pass with minimal changes to them.
* Test all our API entrypoints, and fix our refactored code accordingly
* Compare our old models with our new models; ensure approximately equivalent quality of segmentations on test images.
  * Report the differences, or recommend improvements to IVADOMED code.

## New features?
* BIDS-app for segmentation?
* Git-annex bidsified datasets?

mariehbourget:
As references:

**Models:**
I put an example of model from `ivadomed` here: https://drive.google.com/drive/folders/1vOc-HazqyE4xeyBWwbl4OTpoCcEhafad. There are 3 files:

- json config file of the model
[...]

PR #539, opened 2021-07-09
Delete files/functions/tests we suspect will be no longer necessary after IVADOMED transition

mathieuboudreau:
See: https://docs.google.com/document/d/1Ipd6_nK6I6CKlDI_KpkLXH_jY2C547W_ootalfKnFlU/edit?usp=sharing

## Checklist

## Description
Delete files/functions/tests we suspect will be no longer necessary after IVADOMED transition

## Linked issues
#523


AxonDeepSeg/visualization/visualize.py
```diff
@@ -1,265 +0,0 @@
...
-        ax2.set_ylabel("Loss")
-        return fig
-
-    evolution = retrieve_training_data(path_model)
-    fig = _create_figure_helper(evolution)
-
-    return fig
-
-
-def visualize_segmentation(path):
```
mariehbourget, line 74: I think I would keep the `visualize_segmentation` function in the code for now and only remove the other functions related to training.

Similar to the functions of the other files in the visualization folder, it look like a useful tool but is not used at the moment. I would keep it for now, but we should re-assess its usage later in the integration process with ivadomed.

mariehbourget (commented):
Thank you @mathieuboudreau and @vasudev-sharma for the detailed comparison table and first PR. I went through the table again and add some comments following our meeting.

As for the PR, there is only the function `visualize_segmentation` from `visualize.py` that I would keep for now, everything else looks good to me and match the table.

AxonDeepSeg/visualization/visualize.py
mathieuboudreau: Done in be539d3

mathieuboudreau:
@mariehbourget @vasudev-sharma I think this PR is ready for merging, all that's needed is a quick look at all the files deletes to ensure that it agrees with the list that we curated in PR #523

mariehbourget (approved):
LGTM! I checked the new changes. I did not compare the entire table again because the last time, the only difference was `visualize_segmentation` which was fixed.
Thank you @mathieuboudreau!

PR #546, opened 2021-07-19
Update reqs for Mac M1 ARM. Remove dependencies: TF, Keras, and Albumentation. Remove FSLeyes (temporarily)

mathieuboudreau:
## Checklist

## Description
As we're moving to IVADOMED, I think we can scrap the following requirements: Tensorflow, Keras, and Albumentations. This should also conveniently make AxonDeepSeg compatible with Macs using the ARM M1 processor, as long as there's no issues on the IVADOMED side (?).

I've also updated the requirements for Mac M1 ARM compatibility (MiniForge isn't compatible with Python 3.7). I've increased it to Python 3.9, but am not sure if this will break FSLeyes compatibility. If it does, maybe we should continue anyways (or downgrade to 3.8) and see if we can resolve it on the FSLeyes side.

## Linked issues
#523

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ class PostDevelopCommand(develop):
 
 setup(
     name='AxonDeepSeg',
-    python_requires='>=3.9, <3.10',
+    python_requires='>=3.8, <3.9',
     version=AxonDeepSeg.__version__,
     description='Python tool for automatic axon and myelin segmentation',
     long_description=long_description,
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ class PostDevelopCommand(develop):
 
 setup(
     name='AxonDeepSeg',
-    python_requires='>=3.7, <3.8',
+    python_requires='>=3.9, <3.10',
     version=AxonDeepSeg.__version__,
     description='Python tool for automatic axon and myelin segmentation',
     long_description=long_description,
--- a/test/testing/test_statistics_generation.py
+++ b/test/testing/test_statistics_generation.py
@@ -2,7 +2,6 @@
 
 import json
 from pathlib import Path
-import tensorflow as tf
 import pandas as pd
 
 import pytest
@@ -58,7 +57,7 @@ class TestCore(object):
     @pytest.mark.integration
     def test_metrics_single_wrapper_runs_successfully_and_outfile_exists(self):
         # reset the tensorflow graph for new training
-        tf.reset_default_graph()
+        #tf.reset_default_graph()
 
         path_model_folder = self.modelPath
         path_images_folder = self.imagesPath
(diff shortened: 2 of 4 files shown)


test/testing/test_statistics_generation.py
```diff
@@ -58,7 +57,7 @@ def teardown_class(cls):
     @pytest.mark.integration
     def test_metrics_single_wrapper_runs_successfully_and_outfile_exists(self):
         # reset the tensorflow graph for new training
-        tf.reset_default_graph()
+        #tf.reset_default_graph()
```
mariehbourget, line 60: Just curious: why is it commented and not deleted?

test/testing/test_statistics_generation.py
```diff
@@ -80,7 +79,7 @@ def test_metrics_single_wrapper_runs_successfully_and_outfile_exists(self):
     @pytest.mark.integration
     def test_metrics_classic_wrapper_runs_successfully_and_outfile_exists(self):
         # reset the tensorflow graph for new training
-        tf.reset_default_graph()
+        #tf.reset_default_graph()
```
mariehbourget, line 82: Same question as above

mariehbourget (approved):
LGTM!
The same 4 tests as mentioned in https://github.com/neuropoly/axondeepseg/issues/523#issuecomment-882784073 are failing, so it seems normal at the moment. Is there anything specific you would like me to test/check?

mathieuboudreau:
@mariehbourget this PR is ready for review. I had to remove FSLeyes, hopefully just temporarily, as it doesn't seem to be compatible with either Python 3.9 or the upgraded sci-kit versions. However, now I'm finally able to run tests locally on my Mac M1 machine, which will help with developing ADS v4; we can (hopefully) reinstate FSLeyes after IVADOMED has been integrated.

mathieuboudreau:
Oops just saw you already left messages!

test/testing/test_statistics_generation.py
mathieuboudreau: Hmm I can - I just kept them because, I'm not quite sure what some of these functionalities will be at all moving forward; they seem to track the training evolution (or one of them does), so I wanted to keep these here as a reminder that more work might actually need to be done or we might scrap these functions altogether. I can just delete them here though if you'd like.

mariehbourget: OK no problem, let's keep them as a reminder. I'm not sure either what will happen with those functions/tests.

setup.py
```diff
@@ -26,7 +26,7 @@ def run(self):
 
 setup(
     name='AxonDeepSeg',
-    python_requires='>=3.7, <3.8',
+    python_requires='>=3.9, <3.10',
```
mariehbourget, line 29: If I'm not mistaken, `ivadomed` is currently compatible with python 3.6, 3.7 and 3.8. And there is a [PR](https://github.com/ivadomed/ivadomed/pull/819) open to also be compatible with 3.9 but it is not done yet. Is that required to be able to use the ARM M1?

mathieuboudreau: Ok I wasn't sure - thanks for letting me know! Let me try with Python 3.8 now; if it worked with Python 3.9, I don't forsee any issue.

mathieuboudreau: Done in f834a1c and 6deeab9.

mariehbourget (approved):

PR #587, opened 2022-01-12
Release version 4.0.0

mathieuboudreau:
## Checklist

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've added relevant tests for my contribution
- [x] I've updated the documentation and/or added correct docstrings
- [x] I've assigned a reviewer
- [x] I've consulted [ADS's internal developer documentation](https://github.com/neuropoly/axondeepseg/wiki/Guidelines-for-contribution) to ensure my contribution is in line with any relevant design decisions

## Description

Release version 4.0.0 of AxonDeepSeg that integrated IVADOMED into the project and provides Mac M1 compatibility.

## Linked issues

Resolves #523 #536

PR #592, opened 2022-01-23
Fix NaNs not appearing when generating morphometrics with the GUI

Stoyan-I-A:
Quick fix for the issue mentionned in #587 where NaNs wouldn't appear in the morphometrics file if it was generated in the GUI.


hermancollin (approved):
Tested it on MacOS 10.15.7 and it works well. Thank you for this quick fix

Issue #589, opened 2022-01-19
Discrepancy in morphometrics between v3 and v4


hermancollin:
**Describe the bug**
When comparing morphometrics obtained in v3 and v4, the solidity and orientation values differ.

**To Reproduce**
I computed morphometrics on the _image.png_ test image located in `axondeepseg/test/__test_files__/__test_demo_files__` and indeed every values are the same between versions except solidity and orientation.

![Screenshot_20220119_163714](https://user-images.githubusercontent.com/83031821/150218592-cb3c5dc7-b41d-4110-b394-716e1f55cee1.png)

jcohenadad:
Good catch @hermancollin ! thank you for reporting it (and finding it)!

mathieuboudreau:
Thanks @hermancollin ! After this is resolved but prior to merging, we'll need to add a test to catch a similar bug if it happens again.

hermancollin:
> Good catch @hermancollin ! thank you for reporting it (and finding it)!

To be fair @mariehbourget caught this yesterday and asked me to look into it.

Now, as @mathieuboudreau noted, `scikit-image` was bumped from 0.14.2 to 0.18.3 in the v4 branch. The orientation and solidity values are both properties of the list returned by `skimage.measure.regionprops`. From regionprops doc:

> Deprecated since version 0.16.0: Use “rc” coordinates everywhere. It may be sufficient to call numpy.transpose on your label image to get the same values as 0.15 and earlier. However, for some properties, the transformation will be less trivial. ___For example, the new orientation is pi/2 plus the old orientation___

Indeed, from what I can see, if we add pi/2 to the v4 orientation, we get the same values as in v3.

jcohenadad:
> To be fair @mariehbourget caught this yesterday and asked me to look into it.

ah, thank you for the clarification. Thank you both then 😊

mathieuboudreau:
Since scikit image is permanently using this new convention, should we maybe consider adopting it too, at the cost of differences in values for orientation and solidity between ADS v4 and <v3?

hermancollin:
For the solidity, it looks a bit more convoluted. This is how it's computed:
```python
def solidity(self):
        return self.area / self.area_convex
```
`area_convex` is in turn computed using `skimage.morphology.convex_hull_image`. Not sure what is happening exactly here but in skimage 0.18.0 this function was changed
> morphology.convex_hull_image now uses much less memory by checking hull inequalities in sequence (#5020)

[...]

PR #595, opened 2022-01-25
Add test for morphometrics consistency

hermancollin:
## Checklist

## Description

This PR adds a unit test that compares morphometrics values computed from `__test_files__/__test_demo_files__/image.png` with a reference axonlist.npy file added in the testing data.

## Linked issues

Discussed in #589

### DIFF
--- a/AxonDeepSeg/download_tests.py
+++ b/AxonDeepSeg/download_tests.py
@@ -10,7 +10,7 @@ def download_tests(destination=None):
         destination = convert_path(destination)
         test_files_destination = destination / "__test_files__"
 
-    url_tests = "https://github.com/axondeepseg/data-testing/archive/refs/tags/r20211210.zip"  
+    url_tests = "https://github.com/axondeepseg/data-testing/archive/refs/tags/r20220125.zip"  
     files_before = list(Path.cwd().iterdir())
 
     if (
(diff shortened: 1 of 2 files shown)


hermancollin:
Converted to draft until I figure out why the checks are failing. The cause of the failure is the new test, which is weird since it runs fine locally.

mathieuboudreau:
@hermancollin are there any NaN's in the morphometrics? You can't compare two arrays easily if one contains a NaN, it was giving me a lot of headaches a few weeks ago. If this is the case, I'll share a "hack" to help you test it.

hermancollin:
> @hermancollin are there any NaN's in the morphometrics? You can't compare two arrays easily if one contains a NaN, it was giving me a lot of headaches a few weeks ago. If this is the case, I'll share a "hack" to help you test it.

@mathieuboudreau I don't see any NaN values in the file. Besides, I used your method to assess equality:
```python
for row_ref, row_new in zip(reference_stats_array, new_stats_array):
   if any(math.isnan(val) for val in row_ref.values()) == False and any(math.isnan(val) for val in row_new.values()) == False:
      assert np.array_equal(row_ref, row_new)
```
So IIUC it should still work with NaN values.

I tried running the checks again and weirdly enough they are only successful in linux.

mathieuboudreau:
Gotcha! Thanks!

hermancollin:
So I switched the `np.array_equal` assertion to `np.allclose`, which allows a relative/absolute tolerance when checking equality. It seems the test were failing because of some form of numerical instability. For example, some morphometrics values were slightly different wrt. their last digit. I am not sure what is causing this and it may not be a problem.

Also I used an absolute tolerance of 1e-11 but it's somewhat arbitrary and we might prefer using relative tolerance.

mathieuboudreau:
Excellent - great solution @hermancollin !

Issue #334, opened 2020-07-23
FSLeyes: apply model with different resolutions than default SEM and TEM


mariehbourget:
When applying model in FSLeyes, the resolution is set by default to 0.1 [here](https://github.com/neuropoly/axondeepseg/blob/3bcc87d29816d881dcc4b0eda5f075cdcb564544/ads_plugin.py#L284), and then change to 0.01 [here](https://github.com/neuropoly/axondeepseg/blob/3bcc87d29816d881dcc4b0eda5f075cdcb564544/ads_plugin.py#L323) if a "TEM" model is selected.

It would be nice to be able to choose a resolution when applying the model.
(e.g. If a user trains his own model with a general_pixel_size different than 0.1 or 0.01, and wants to use FSLeyes to make manual corrections or compute morphometrics.)

Stoyan-I-A:
What would be the best way to implement this in the interface? One option would be to add a text line where the resolution could be modified by the user.

mariehbourget:
Yeah, that would work. Or maybe it could be a text box like the one for "Enter pixel size in micrometer"?
In any case, I think it should be mandatory only if the user choose something else than the default models.

mathieuboudreau:
>It would be nice to be able to choose a resolution when applying the model.
(e.g. If a user trains his own model with a general_pixel_size different than 0.1 or 0.01, and wants to use FSLeyes to make manual corrections or compute morphometrics.)

If a user trains a model with a different general_pixel_size, and wants to use FSleyes, we should simply add that model as another drop-down option. I think adding the option to the user of setting the general_pixel_size might actually make them mistakenly input the image pixel size, and not the pixel size that was used for training (which they might not even know).

mathieuboudreau:
Either that, or even better this value should actually be saved in the model file, and be read from it. (since the value will always be tied with the model)

mariehbourget:
Right now, if a model is in `AxonDeepSeg/models`, it appears in the drop down menu and can be used in FSLeyes. The problem is that the general_pixel_size is hardcoded in the plugin and cannot be changed (either 0.1, or 0.01 if "TEM" is in the name of the model).
I agree that it would be a good idea to save this value in the training config.
We will have to document this very well because the general_pixel_size is set in the dataset_building notebook (where we do patches) and not in the training notebook.

mariehbourget:
@Stoyan-I-A 
[...]

PR #611, opened 2022-03-07
Remove some unused variables in the plugin

Stoyan-I-A:
As mentionned [here](https://github.com/axondeepseg/axondeepseg/issues/334#issuecomment-1059519811), some resolution variables are no longer used since the Ivadomed integration. We just forgot to remove them when releasing the v4.

### DIFF
--- a/ads_plugin.py
+++ b/ads_plugin.py
@@ -380,9 +380,6 @@ class ADScontrol(ctrlpanel.ControlPanel):
         selected in the combobox. The segmentation masks are then loaded into FSLeyes
         """
 
-        # Declare the default resolution of the model
-        resolution = 0.1
-
         # Get the image name and directory
         image_overlay = self.get_visible_image_overlay()
         if self.get_visible_image_overlay() == None:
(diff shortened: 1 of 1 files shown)


mariehbourget (commented):
Thank you @Stoyan-I-A for looking into this.

I think there may be other variables (unused) that were related to the model resolution for #334 on lines [55-57](https://github.com/axondeepseg/axondeepseg/blob/6d745492cc091576922b579bc67c56a72905b838/ads_plugin.py#L55-L57) as well.

Stoyan-I-A:
> Thank you @Stoyan-I-A for looking into this.
> 
> I think there may be other variables (unused) that were related to the model resolution for #334 on lines [55-57](https://github.com/axondeepseg/axondeepseg/blob/6d745492cc091576922b579bc67c56a72905b838/ads_plugin.py#L55-L57) as well.

Ah yes I remember. I added them because it thought we planned on letting the user modify the model resolution manually.
We can remove them as well.

mariehbourget (approved):
LGTM, thanks @Stoyan-I-A!

Issue #610, opened 2022-03-04
Implement zoom factor sweep and segmentation auto-evaluation


mariehbourget:
**Describe the problem**

When segmenting images than have a different resolution and/or axon sizes than the model resolution, we often have to try different pixel sizes to optimize the results, see #604 for example.

**Proposed solution**

In #609, we plan to add a new "zoom factor" option to the CLI that would multiply the pixel size, instead of changing the real pixel size itself.

In addition, we would like to explore the possibility of doing an "sweep" of different zoom factor in order to find automatically an optimal range or estimate for the zoom factor.

We would first need to define a metric to estimate the "best" zoom factor. For example: some sort of "ring" image identification (to find closed myelin), and then find the zoom factor that maximizes the number of rings.

jcohenadad:
I *love* this idea!

hermancollin:
To address this issue, the first part would be to implement the sweep feature. We can then later add on top a metric like discussed above.

Should it be an option for the main segmentation CLI or a separate script?

mathieuboudreau:
I definitely think it should be an option in the main segmentation CLI, but could be written in a function outside of segment.py

hermancollin:
The next part for this feature will be to train a classifier to evaluate the masks produced by the sweep. Here is my idea for the components of the training set:
- generate purposefully "bad" segmentations, i.e. the zoom factor is clearly not right for the input image and it produces incomplete/bad segmentations. Label these masks as ~0.2
- generate ok segmentations, i.e. the mask is not perfect but its what ADS can optimally achieve or close to it. Label these masks as ~0.8
- use some available manually corrected masks available from our datasets. Label these as 1 to show the model ideal cases. I happen to know that the TEM dataset contains more than 1.3 billion pixels of manually corrected masks, so we probably don't want to use everything as we need to keep some class balance. 

Then, we add the trained model to ADS and apply it to every segmentation produced by the sweep. It could then output the processed zoom factors ranked by the score the model gave them.

mathieuboudreau:
Closing, implemented in https://github.com/axondeepseg/axondeepseg/pull/632

hermancollin:
@mathieuboudreau thank you for the extensive cleanup you did last week. However I would like to keep this specific issue opened because the "metric" we were talking about is still not implemented. The sweep feature was only the first "easy" part but the mask evaluation we discussed could be an interesting internship project .

mathieuboudreau:
[...]

PR #620, opened 2022-03-17
Zoom factor option for segmentation via CLI

### LINKED ISSUE #609: Add zoom factor option to CLI segmentation
As discussed in meeting, we already have a "zoom factor" option in FSLeyes (implemented in #462).
This zoom factor is applied to the pixel size while segmenting and it:
- helps get optimal segmentations when resolution and/or axon sizes are different from the model resolution
- avoids "artificially" changing the pixel size when launching segmentation

We should implement this zoom factor in the CLI as well and document it.

Related issue #604.

hermancollin:
## Checklist

## Description

This PR adds a zoom factor option for CLI segmentation.
This is a very minimal implementation. For #610, if we want to do a sweep on a range of zoom factor values, this option wouldn't even be used because we would be more likely to use `segment.py:segment_image()`.

Still need to update the documentation. Also, any comment on the help message of the argument is welcome.

```
» axondeepseg --help                                                                                              herman@tank
AxonDeepSeg v.4.0.0
usage: axondeepseg [-h] -t {SEM,TEM,BF} -i IMGPATH [IMGPATH ...] [-m MODEL] [-s SIZEPIXEL] [-v {0,1}] [--overlap OVERLAP] [-z ZOOM]

required arguments:
[...]

optional arguments:
[...]
  -z ZOOM, --zoom ZOOM  Zoom factor to adjust the acquired pixel size of the image(s). 
                        The pixel size used for the segmentation will be the product of the 
                        specified pixel size and the zoom factor.

```

## Linked issues

Resolves #609 
Also, this feature should be mentioned in #619 when the RuntimeError related to minimum patch size occurs.

### DIFF
--- a/AxonDeepSeg/segment.py
+++ b/AxonDeepSeg/segment.py
@@ -219,6 +219,10 @@ def main(argv=None):
                                                             'Default value: '+str(default_overlap)+'\n'+
                                                             'Recommended range of values: [10-100]. \n',
                                                             default=default_overlap)
+    ap.add_argument("-z", "--zoom", required=False, help='Zoom factor to adjust the acquired pixel size of the image(s). \n'+
+                                                            'The pixel size used for the segmentation will be the product of the \n'+
+                                                            'specified pixel size and the zoom factor.',
+                                                            default=None)
     ap._action_groups.reverse()
 
     # Processing the arguments
@@ -232,6 +236,10 @@ def main(argv=None):
         psm = None
     path_target_list = [Path(p) for p in args["imgpath"]]
     new_path = Path(args["model"]) if args["model"] else None 
+    if args["zoom"] is not None:
+        zoom_factor = float(args["zoom"])
+    else:
+        zoom_factor = 1.0
 
     # Preparing the arguments to axon_segmentation function
     path_model = generate_default_parameters(type_, new_path)
@@ -277,7 +285,7 @@ def main(argv=None):
                     path_testing_image=current_path_target,
                     path_model=path_model,
                     overlap_value=overlap_value,
-                    acquired_resolution=psm,
+                    acquired_resolution=psm * zoom_factor,
                     verbosity_level=verbosity_level
                     )
 
@@ -312,7 +320,7 @@ def main(argv=None):
                 path_testing_images_folder=current_path_target,
                 path_model=path_model,
                 overlap_value=overlap_value,
-                acquired_resolution=psm,
+                acquired_resolution=psm * zoom_factor,
                 verbosity_level=verbosity_level
                 )
 


mariehbourget:
> Also, any comment on the help message of the argument is welcome.

We could go with the same as in FSLeyes, minus the mention of morphometrics (which is a different CLI): https://github.com/axondeepseg/axondeepseg/blob/ca4baad96bed0d9c84b2dc63a8c4b464dc4d5a2c/ads_plugin.py#L79-L80

hermancollin:
> We could go with the same as in FSLeyes, minus the mention of morphometrics (which is a different CLI):
> 
> https://github.com/axondeepseg/axondeepseg/blob/ca4baad96bed0d9c84b2dc63a8c4b464dc4d5a2c/ads_plugin.py#L79-L80

Ah yes missed that one. Now the help message looks like this:
```bash
  -z ZOOM, --zoom ZOOM  Zoom factor. 
                        When applying the model, the pixel size of the image will be 
                        multiplied by this number.
```

hermancollin:
Ok so I refactored the `segment_image` and `segment_folders` functions to accept an additional argument for the zoom factor. This change slightly affects how the function is used, and I already updated the plugin file.

Regarding tests, I did not add a dedicated test in test_segment.py. However, I added the `-z` option to several pre-existing unit tests. In #619, some tests will be added to make sure the minimum patch size (length_2D) error is handled appropriately. At this point, I will make sure to add a second test to assert that the exception is *not* thrown when the zoom factor is used to fix this.

Finally, I added the `-z` argument description in the doc. Also, I took the liberty of adding a sub-section in the ***Using AxonDeepSeg/Segmentation*** to document why this feature is useful. Let me know what you think. The doc build is available [here](https://axondeepseg--620.org.readthedocs.build/en/620/documentation.html#segmentation).

mathieuboudreau:
>Regarding tests, I did not add a dedicated test in test_segment.py. However, I added the -z option to several pre-existing unit tests. In https://github.com/axondeepseg/axondeepseg/pull/619, some tests will be added to make sure the minimum patch size (length_2D) error is handled appropriately. At this point, I will make sure to add a second test to assert that the exception is not thrown when the zoom factor is used to fix this.

I think, because this feature is quite easily measurable (zoom by a certain %/factor, and the output of the zoom should be by that much enlargement), that maybe it should be factored out into a function/method that can be tested as a unit. This will give better confidence on the functionality that was implemented, and ensure that future refactoring doesn't change that behaviour (i.e. even though you've added the -z options to a bunch of tests, if a bug were to be introduced that a -z factor of 20% actually zooms by 40%, these tests won't pick up on that change in behaviour). 

I'll look into it during my review to see if it's something I could implemented easily, likely tomorrow.

mathieuboudreau:
Quickly glancing at the code, maybe writing a function for it is overboard. I wouldn't call the zoom factor in the tests with the wrong values though (i.e. 1.01), it should just be called with 1.0 since those are the true value.

What we could do, is make an image that is too small to be used with a zoom factor of 1.0, test that it throws an error (or the proper message), and then test that when you call it with a sufficient zoom factor that it then works.

mathieuboudreau:
@hermancollin @mariehbourget I've implemented some error handling for the case with too small images size/zoom factor, done if for the segment folder case as well, and also added more descriptive messages indicating the the minimum zoom factor  to use (I think we'll be able to thus close #619, since it's being done here).

You can test it out with this new test image https://github.com/axondeepseg/data-testing/blob/main/__test_files__/__test_segment_zoom__/image.png (TEM, image size ~2000x2000 and resolution of 0.00236)


image


Here is an example log without zoom factor:

```
(ads_venv) mathieuboudreau@Mathieus-MacBook-Pro axondeepseg % axondeepseg -t TEM -i test/__test_files__/__test_segment_zoom__/image.png -v 1 -s 0.00236 
AxonDeepSeg v.4.0.0
test/__test_files__/__test_segment_zoom__/image.png
ERROR: Due to your given image size, resolution, and zoom factor, the resampled image is smaller than the patch size during segmentation. To resolve this, please set a zoom factor greater than  1.1 . To do this on the command line, call the segmentation with the -z flag, i.e.  -z  1.1
(ads_venv) mathieuboudreau@Mathieus-MacBook-Pro axondeepseg % 
```

and then with:

```

(ads_venv) mathieuboudreau@Mathieus-MacBook-Pro axondeepseg % axondeepseg -t TEM -i test/__test_files__/__test_segment_zoom__/image.png -v 1 -s 0.00236 -z 1.1
AxonDeepSeg v.4.0.0
test/__test_files__/__test_segment_zoom__/image.png
2022-03-23 14:11:16.926 | INFO     | ivadomed.config_manager:_display_differing_keys:153 - Adding the following keys to the configuration file
2022-03-23 14:11:16.926 | INFO     | ivadomed.config_manager:_display_differing_keys:155 - 

2022-03-23 14:11:16.926 | WARNING  | ivadomed.inference:segment_volume:402 - fname_roi has not been specified, then the entire volume is processed.
2022-03-23 14:11:17.244 | INFO     | ivadomed.inference:segment_volume:448 - Loaded 4 axial patches of shape [512, 512].
2022-03-23 14:11:17.272 | INFO     | ivadomed.utils:define_device:129 - Cuda is not available.
2022-03-23 14:11:17.272 | INFO     | ivadomed.utils:define_device:130 - Working on cpu.
2022-03-23 14:11:17.272 | DEBUG    | ivadomed.inference:get_preds:88 - Likely ONNX model detected at: /Users/mathieuboudreau/neuropoly/github/axondeepseg/AxonDeepSeg/models/model_seg_mouse_axon-myelin_tem/model_seg_mouse_axon-myelin_tem.onnx
2022-03-23 14:11:17.272 | DEBUG    | ivadomed.inference:get_preds:89 - Conduct ONNX model inference... 
2022-03-23 14:11:30.835 | DEBUG    | ivadomed.inference:get_preds:92 - Sending predictions to CPU
Lossy conversion from float64 to uint8. Range [0, 1]. Convert image to uint8 prior to saving to suppress this warning.
Lossy conversion from float64 to uint8. Range [0, 1]. Convert image to uint8 prior to saving to suppress this warning.
Lossy conversion from float64 to uint8. Range [0.0, 255.0]. Convert image to uint8 prior to saving to suppress this warning.
Image /Users/mathieuboudreau/neuropoly/github/axondeepseg/test/__test_files__/__test_segment_zoom__/image.png segmented.
Segmentation finished.
(ads_venv) mathieuboudreau@Mathieus-MacBook-Pro axondeepseg % 

```

with output:

image_seg-axonmyelin

hermancollin:
> What we could do, is make an image that is too small to be used with a zoom factor of 1.0, test that it throws an error (or the proper message), and then test that when you call it with a sufficient zoom factor that it then works.

I guess I'm a bit late for this but we didn't need a new testing image to trigger the error. It can work with any image if the pixel size specified is small enough.

mathieuboudreau:
> > What we could do, is make an image that is too small to be used with a zoom factor of 1.0, test that it throws an error (or the proper message), and then test that when you call it with a sufficient zoom factor that it then works.
> 
> I guess I'm a bit late for this but we didn't need a new testing image to trigger the error. It can work with any image if the pixel size specified is small enough.

Yup I know, but I find this a bit "cleaner" this way (i.e. it was easier for me to develop the feature, knowing what the real resolution and etc were, just from a mental calculation point of view). Also, I wanted to test it in a folder where one image is OK, and the other is small.

hermancollin:
@mathieuboudreau Looking at your changes, I think the function you added in segment.py to check the model parameters is cleaner than what I did in #619. Also a great job on the tests.

AxonDeepSeg/segment.py
```diff
@@ -108,10 +139,38 @@ def segment_image(
...
+            # Round to 1 decimal, always up.
+            minimum_zoom_factor = ceil(minimum_zoom_factor*10)/10
+
+            print("ERROR: Due to your given image size, resolution, and zoom factor, the resampled image is smaller than",
+                   "the patch size during segmentation. To resolve this, please set a zoom factor greater than ",
+                   str(minimum_zoom_factor), ".",
+                   "To do this on the command line, call the segmentation with the -z flag, i.e. ",
+                   "-z ", str(minimum_zoom_factor),
+            )
+            sys.exit(4)
```
mariehbourget, line 167: The `sys.exit` makes `fsleyes` quit. Is there a way to display to avoid that and still have the error in the fsleyes terminal?

AxonDeepSeg/segment.py
```diff
@@ -108,10 +139,38 @@ def segment_image(
...
 
+        # Get the model's native resolution
+        model_resolution, patch_size = get_model_native_resolution_and_patch(path_model)
+
+        # Check that the resampled image will be of sufficient size, and if not throw an error.
+        im = Image.open(path_testing_image)
+        w, h = im.size
+
+        w_resampled = w*(acquired_resolution*zoom_factor)/model_resolution
+        h_resampled = h*(acquired_resolution*zoom_factor)/model_resolution
```
mariehbourget, line 150: The "/" is problematic here in case there is a mix between float values (isotropic acquired_resolution) and list values (anisotropic model_resolution)
```
TypeError: unsupported operand type(s) for /: 'float' and 'list'
```

AxonDeepSeg/segment.py
```diff
@@ -108,10 +139,38 @@ def segment_image(
...
+
+        if w_resampled < patch_size or h_resampled < patch_size:
```
mariehbourget, line 152: Same as above for "<" with list for patch_size
```
TypeError: '<' not supported between instances of 'float' and 'list'
```

AxonDeepSeg/segment.py
```diff
@@ -73,11 +76,38 @@ def generate_default_parameters(type_acquisition, new_path):
...
+    model_resolution[0] = model_resolution [0] * resolution_unit_conversion_factor
+    model_resolution[1] = model_resolution [1] * resolution_unit_conversion_factor
+    
+    patch_size = [model['default_model']['length_2D'][0], model['default_model']['length_2D'][1]]
+
+    if model_resolution[0] == model_resolution[1]: #isotropic pixels
+        model_resolution = model_resolution[0]
+
+    if patch_size[0] == patch_size[1]: #isotropic pixels
+        patch_size = patch_size[0]
```
mariehbourget, line 101: See comments below, the code throws errors when there is a mix of float (isotropic) and list (anisotropic) values coming from here.

AxonDeepSeg/segment.py
```diff
@@ -169,9 +233,36 @@ def segment_folders(path_testing_images_folder, path_model,
...
+
+            print("ERROR: Due to your given image size, resolution, and zoom factor, at least one image, ", 
+                   str(path_testing_images_folder / file_),
+                   ", is smaller than",
+                   "the patch size after it is resampled during segmentation. To resolve this, please set a zoom factor greater than ",
+                   str(minimum_zoom_factor), ".",
+                   "To do this on the command line, call the segmentation with the -z flag, i.e. ",
+                   "-z ", str(minimum_zoom_factor),
+            )
+            sys.exit(4)
```
mariehbourget, line 260: This check is done in the for loop, so the error will be raised only when 1 of the file fails, after segmenting previous X images that were OK and leaving the next Y images not segmented.

Suggestions:
1. Check the whole folder for "bad" resolution before starting the segmentation.
2. Or skip to the next file and provide a list of "failed" segmentation at the end of the folder.

mathieuboudreau:
@hermancollin Thanks!

I've also slightly modified the text in the documentation. Although in the code, we are in fact multiplying a resolution by the zoom factor, what actually happens later on is that the image is resampled so that the patches are "zoomed into" the image, so I think I prefer this less technical explanation for users. We can keep working on the semantics of the explanation in another issue if you don't completely agree with this approach.

Everything else looks great, thanks a lot for working on this issue and #619 (even though we closed it without merging, sorry about cutting you off there!).

mathieuboudreau (approved):
Previous review left in comments. Looks good to go!

mariehbourget (commented):
Thanks @hermancollin and @mathieuboudreau for working on this.
I tested the CLI and fsleyes and left a few comments and suggestions.

Also, I found something odd.
There is one "corner-case" where the zoom factor would not be applied correctly, i.e. if someone use the `segment_image` function directly (ex: in cell 2.2 of the `00-getting_started notebook`), but without the `acquired_resolution` argument. In that case the pixel size is extracted from the `pixel_in_micrometer` file in `apply_model` [here](https://github.com/axondeepseg/axondeepseg/blob/ca4baad96bed0d9c84b2dc63a8c4b464dc4d5a2c/AxonDeepSeg/apply_model.py#L31), so after the "new" pixel size is computed with the zoom factor.
We may want to rethink that particular case.

AxonDeepSeg/segment.py
```diff
@@ -108,10 +139,38 @@ def segment_image(
 
         img_name_original = acquisition_name.stem
 
+        # Get the model's native resolution
+        model_resolution, patch_size = get_model_native_resolution_and_patch(path_model)
+
+        # Check that the resampled image will be of sufficient size, and if not throw an error.
+        im = Image.open(path_testing_image)
```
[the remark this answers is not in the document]
mathieuboudreau: Ah right, I thought about it while I was writing but forgot to change it to that function, thanks for pointing it out!

AxonDeepSeg/segment.py
mathieuboudreau: Hum, I used this because we were already using this elsewhere already, see here for an example: 

https://github.com/axondeepseg/axondeepseg/blob/ca4baad96bed0d9c84b2dc63a8c4b464dc4d5a2c/AxonDeepSeg/segment.py#L268-L272

I guess there, it's only called when the CLI is called. Should we handle the zoom factor error there as well, and then in FSLeyes seperately in the plugin?

AxonDeepSeg/segment.py
mathieuboudreau: I was thinking about that, but felt that programming for the anisotropic case here now was going overboard, as I still have yet to have a model/test image to work with during development. It is difficult to continue supporting an edge case like this only in abstract terms. If you can provide me with a model trained on anisotropic image and a test image to go with it, then I think we could continue supporting this edge case, as then I can write tests for it, otherwise, I'd simply write an error/exception message that we don't support it yet through an if/else.

hermancollin: Maybe we could fix this by adding an else clause in `get_model_native_resolution_and_patch()` like so:
```python
if model_resolution[0] == model_resolution[1]: #isotropic pixels
   model_resolution = model_resolution[0]
else:
   model_resolution = max(model_resolution)
```
As for the anisotropic `acquired_resolution` we would need to check it before applying this formula.

AxonDeepSeg/segment.py
mathieuboudreau: I think we need to do option 2, as we would have to load each images once just to do 1), and then reload them again when we actually do the segmentation.

Should we start to have a log for these types of situations?

AxonDeepSeg/segment.py
```diff
@@ -108,10 +139,38 @@ def segment_image(
...
+                minimum_zoom_factor = patch_size*model_resolution/(h*acquired_resolution)
+
+            # Round to 1 decimal, always up.
+            minimum_zoom_factor = ceil(minimum_zoom_factor*10)/10
+
+            print("ERROR: Due to your given image size, resolution, and zoom factor, the resampled image is smaller than",
+                   "the patch size during segmentation. To resolve this, please set a zoom factor greater than ",
+                   str(minimum_zoom_factor), ".",
+                   "To do this on the command line, call the segmentation with the -z flag, i.e. ",
+                   "-z ", str(minimum_zoom_factor),
```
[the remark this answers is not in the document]
mathieuboudreau: I think maybe we should add a popup in the plugin with the same message, so we'd need to add it inside the plugin file itself first?

AxonDeepSeg/segment.py
hermancollin: Same fix as above, but this time using min instead of max:
```python
if patch_size[0] == patch_size[1]: #isotropic pixels
   patch_size = patch_size[0]
else:
   patch_size = min(patch_size)
```

AxonDeepSeg/segment.py
hermancollin: We can wrap the call to segment_image in the plugin around a try-catch block and intercept the `SystemExit` exception.

mathieuboudreau: Ahh good idea, and we could in fact trigger the FSLeyes popup with the message if the sys error code 4 is triggered.

AxonDeepSeg/segment.py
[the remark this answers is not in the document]
hermancollin: Hm I think we should catch the exception in the plugin and display a message in the GUI instead because the FSLeyes users may not notice the error.

AxonDeepSeg/segment.py
hermancollin: @mathieuboudreau yep that's what I had in mind. I can take care of it (for some reason fsleyes suddently works on my linux)

AxonDeepSeg/segment.py
mariehbourget: I'm not sure that would work because it's the combination of "model resolution" and "patch_size" for a _given dimension_ that is problematic, not only the max model_resolution and min patch_size of the 2 dimensions.

My point was more that the current implementation is not consistent between the model_resolution and patch_size assignement on lines 97-100 (accepting both isotropic and anisotropic) with the computation here that assumes isotropic.

I agree with Mathieu that we are not dealing with anisotropic pixels elsewhere in ADS, so throwing an error here for the moment would be sufficient.

AxonDeepSeg/segment.py
hermancollin: Only problem is that the minimum zoom factor value is lost

AxonDeepSeg/segment.py
mariehbourget: Yes, a log is a good idea!
Saving the terminal output in a file (with date and time in the name) along with the segmentations would be helpful.

This may be out of scope for this PR. The current behavior is not ideal, but I would be comfortable merging like this and dealing with the loop and log separately if you prefer.

AxonDeepSeg/segment.py
hermancollin: Ah but we can tell the user to take a look at his terminal to check the minimum zoom factor because the "error" message doesn't actually comes from an error so it is still printed.

hermancollin: Ok so this is now fixed.

Screenshot_20220323_173148

mathieuboudreau:
>[here](https://github.com/axondeepseg/axondeepseg/blob/ca4baad96bed0d9c84b2dc63a8c4b464dc4d5a2c/AxonDeepSeg/apply_model.py#L31) is one "corner-case" where the zoom factor would not be applied correctly, i.e. if someone use the segment_image function directly (ex: in cell 2.2 of the 00-getting_started notebook), but without the acquired_resolution argument. In that case the pixel size is extracted from the pixel_in_micrometer file in apply_model here, so after the "new" pixel size is computed with the zoom factor.
We may want to rethink that particular case.

Good catch! 

I'll write a test for that case, let it fail, and then see how we can resolve it.

hermancollin:
> > [here](https://github.com/axondeepseg/axondeepseg/blob/ca4baad96bed0d9c84b2dc63a8c4b464dc4d5a2c/AxonDeepSeg/apply_model.py#L31) is one "corner-case" where the zoom factor would not be applied correctly, i.e. if someone use the segment_image function directly (ex: in cell 2.2 of the 00-getting_started notebook), but without the acquired_resolution argument. In that case the pixel size is extracted from the pixel_in_micrometer file in apply_model here, so after the "new" pixel size is computed with the zoom factor.
> > We may want to rethink that particular case.
> 
> Good catch!
> 
> I'll write a test for that case, let it fail, and then see how we can resolve it.

Wouldn't it make more sense to read the txt file in segment.py instead of apply_model? After a quick glance, the only time `axon_segmentation` (from *apply_model.py*) is called is in `segment_image` and `segment_folders` (theres an import for it in the plugin but its unused).

mathieuboudreau:
> > > [here](https://github.com/axondeepseg/axondeepseg/blob/ca4baad96bed0d9c84b2dc63a8c4b464dc4d5a2c/AxonDeepSeg/apply_model.py#L31) is one "corner-case" where the zoom factor would not be applied correctly, i.e. if someone use the segment_image function directly (ex: in cell 2.2 of the 00-getting_started notebook), but without the acquired_resolution argument. In that case the pixel size is extracted from the pixel_in_micrometer file in apply_model here, so after the "new" pixel size is computed with the zoom factor.
> > > We may want to rethink that particular case.
> > 
> > 
> > Good catch!
> > I'll write a test for that case, let it fail, and then see how we can resolve it.
> 
> Wouldn't it make more sense to read the txt file in segment.py instead of apply_model? After a quick glance, the only time `axon_segmentation` (from _apply_model.py_) is called is in `segment_image` and `segment_folders` (theres an import for it in the plugin but its unused).

Great point @hermancollin ! I agree. I think why I originally wrote it inside apply_model is because that allowed me to resolve this pixel value in one location; otherwise, we'll need to write it twice (once in segment_image and once in segment_folders). But then, that introduced this bug, se the other route seems to be the best way forward.

hermancollin:
> > Wouldn't it make more sense to read the txt file in segment.py instead of apply_model? After a quick glance, the only time `axon_segmentation` (from _apply_model.py_) is called is in `segment_image` and `segment_folders` (theres an import for it in the plugin but its unused).
> 
> Great point @hermancollin ! I agree. I think why I originally wrote it inside apply_model is because that allowed me to resolve this pixel value in one location; otherwise, we'll need to write it twice (once in segment_image and once in segment_folders). But then, that introduced this bug, se the other route seems to be the best way forward.

I see. We can always put this inside a function in segment.py for less redundancy.

In that case, we should change the `acquired_resolution` argument of `axon_segmentation()` to be required instead of optional as it will always be handled before its called.

mathieuboudreau:
> Also, I found something odd. There is one "corner-case" where the zoom factor would not be applied correctly, i.e. if someone use the `segment_image` function directly (ex: in cell 2.2 of the `00-getting_started notebook`), but without the `acquired_resolution` argument. In that case the pixel size is extracted from the `pixel_in_micrometer` file in `apply_model` [here](https://github.com/axondeepseg/axondeepseg/blob/ca4baad96bed0d9c84b2dc63a8c4b464dc4d5a2c/AxonDeepSeg/apply_model.py#L31), so after the "new" pixel size is computed with the zoom factor. We may want to rethink that particular case.

Resolved in [c0838b0](https://github.com/axondeepseg/axondeepseg/pull/620/commits/c0838b0cf9aa8773d31d1b3fa82ede46dc4b013e)

AxonDeepSeg/segment.py
```diff
@@ -127,6 +127,19 @@ def segment_image(
...
+    if acquired_resolution == None:
+        if (Path(path_testing_image.parents[0]) / 'pixel_size_in_micrometer.txt').exists():
+            resolutions_file = open(Path(path_testing_image.parents[0]) / 'pixel_size_in_micrometer.txt', 'r')
+            str_resolution = float(resolutions_file.read())
+            acquired_resolution = float(str_resolution)
+        else:
+            exception_msg = "ERROR: No pixel size is provided, and there is no pixel_size_in_micrometer.txt file in image folder. " \
+                            "Please provide a pixel size (using argument acquired_resolution), or add a pixel_size_in_micrometer.txt file " \
+                            "containing the pixel size value."
+            raise Exception(exception_msg)
```
hermancollin, line 141: Shouldn't we use `sys.exit(3)` again for this? It's essentially the same error.

mathieuboudreau: I'll check to see if doing that makes any test fail; I don't recall why we wrote it this way here. Maybe because here we're not calling directly through a command line (i.e. to reach this error, we *know* it will be a Python call to the function, so raising a Python exception allows us to catch and throw it in other loops, while in the other case we wrote the error inside the CLI call, so we'll always want to exit it completely).

mathieuboudreau: But, I wouldn't dwell too much on these tangential details, we'll start to go off track from the aim of this PR (which is the zoom factor); new issues could be opened to deal with this type of question/inconsistency.

hermancollin: I agree (the PR is already bigger than expected)

For what it's worth this error was probably never thrown (except in the edge case of calling directly `segment_image`) because the pixel size text files are already read inside the main of segment.py (before the call to `segment_image`/`segment_folder`). So, in total, we read the text files in 4 different places in segment.py.

AxonDeepSeg/segment.py
hermancollin: The `def get_model_native_resolution_and_patch` function now throws an error if the model uses an anisotropic pixel size, as a temporary fix until we decide to add support for it. I tried it with an anisotropic resolution in the model json config and it works for now.

mathieuboudreau:
@mariehbourget I think it's time for you to do a re-review, as all your comments have been addressed I believe.

mariehbourget (approved):
Thanks @hermancollin and @mathieuboudreau, everything is working great!

Question: I don't see the RTD build in the CI checks anymore, is that normal?

hermancollin:
> Thanks @hermancollin and @mathieuboudreau, everything is working great!
> 
> Question: I don't see the RTD build in the CI checks anymore, is that normal?

All the builds are here: https://readthedocs.org/projects/axondeepseg/builds/

New optional argument description [here](https://axondeepseg--620.org.readthedocs.build/en/620/documentation.html#using-axondeepseg) and new subsection [here](https://axondeepseg--620.org.readthedocs.build/en/620/documentation.html#segment-images-using-a-zoom-factor)

Issue #599, opened 2022-02-04
Update models README and add link in documentation


mariehbourget:
Following the release of version 4, the models were updated on their respective repos:
https://github.com/axondeepseg/default-SEM-model
https://github.com/axondeepseg/default-TEM-model
https://github.com/axondeepseg/default-BF-model

As discussed in meeting, we would like to update the README files with more information about training and usage of the models.
Example of another ivadomed model: https://github.com/ivadomed/model_seg_ms_mp2rage

When updated, we will also add the links to the repos in our documentation here: https://axondeepseg.readthedocs.io/en/latest/documentation.html#existing-models

hermancollin:
Also, a good starting point for this issue would be to update the template here: https://github.com/neuropoly/axondeepseg/wiki/Template-for-Readme-of-new-models

mariehbourget:
> I noticed the BF model lack a license file (other models have a MIT license). Is that intended?

Good catch, I don't think that is intended, @mathieuboudreau?

hermancollin:
Now that the SEM model readme is merged, I updated both the TEM and BF ones. Also, the template is updated.

hermancollin:
I added the version files in both the [TEM model repo](https://github.com/axondeepseg/default-TEM-model) and the [BF model repo](https://github.com/axondeepseg/default-BF-model). 

Also opened #621 to quickly add the repository links in the doc. Everything is ready for review/merge and then we can close this issue.

Issue #573, opened 2021-10-12
Add model names in the documentation


jcohenadad:
**Describe the problem**

Currently in the doc, we only refer to "EM models", "OM models", etc. It would be wise to also refer to the exact name of the model, so that people can refer to it when they publish, ask a question, etc.

**Proposed solution**

List the models in the documentation. We should consider an automated way of doing this. Eg: crawling in a hard-coded list of available models to download?  Also, we should do that after ADS is using ivadomed #547.

For "old" models that need to be renamed (eg: ppl citing a pre-ivadomed model in a paper), I suggest referencing them as: 
~~~
- `my_awesome_model` (formerly called `model_seg_pns_bf`): Segment blablabla...
~~~

PR #621, opened 2022-03-21
Add model repo links in the doc

hermancollin:
## Checklist

## Description

This simply adds the model repository links in the doc.

## Linked issues

Following the models README update from #599

### DIFF
--- a/docs/source/documentation.rst
+++ b/docs/source/documentation.rst
@@ -184,7 +184,7 @@ The three models are described below:
 
 * **SEM** model (*model_seg_rat_axon-myelin_sem*), that works at a resolution of 0.1 micrometer per pixel. For more information, please visit the `SEM model repository <https://github.com/axondeepseg/default-SEM-model>`_.
 * **TEM** model (*model_seg_mouse_axon-myelin_tem*), that works at a resolution of 0.01 micrometer per pixel. For more information, please visit the `TEM model repository <https://github.com/axondeepseg/default-TEM-model>`_.
-* **BF** (bright-field) model (*model_seg_rat_axon-myelin_bf*, formerly known as *model_seg_pns_bf*), that works at a resolution of 0.1 micrometer per pixel. For more information, please visit the `BF model repository <https://github.com/axondeepseg/default-BF-model>`_.
+* **BF** (bright-field) model (*model_seg_rat_axon-myelin_bf*, formerly called *model_seg_pns_bf*), that works at a resolution of 0.1 micrometer per pixel. For more information, please visit the `BF model repository <https://github.com/axondeepseg/default-BF-model>`_.
 
 Using AxonDeepSeg
 =================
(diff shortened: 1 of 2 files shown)


hermancollin:
Link to the compiled doc: https://axondeepseg--621.org.readthedocs.build/en/621/documentation.html#existing-models

docs/source/documentation.rst
```diff
@@ -182,9 +182,9 @@ Existing models
 Three models are available and shipped together with the installation package, so you don't need to install them separately.
 The three models are described below:
 
-* A SEM model, that works at a resolution of 0.1 micrometer per pixel.
-* A TEM model, that works at a resolution of 0.01 micrometer per pixel.
-* A BF (bright-field) model, that works at a resolution of 0.1 micrometer per pixel.
+* A SEM model, that works at a resolution of 0.1 micrometer per pixel. For more information, please visit the `SEM model repository <https://github.com/axondeepseg/default-SEM-model>`_.
```
mariehbourget, line 185: Maybe we could add the name of the model itself, ex:
```suggestion
* A SEM model (model_seg_rat_axon-myelin_sem), that works at a resolution of 0.1 micrometer per pixel. For more information, please visit the `SEM model repository <https://github.com/axondeepseg/default-SEM-model>`_.
```

hermancollin: Good suggestion this is a good opportunity to document their names. We had a question about these names recently.

mariehbourget: I just remembered, we had an issue about this as well #573. It might be a good time to think about how to track "old" model names when we changed them, as discussed in the issue.

hermancollin: Hmm that's interesting. I can add it manually easily but I don't know how we can automate this with RTD. Are the model names likely to be changed again? 
The problem is that I don't even know where we would fetch this information. For example, at this point, where is the `model_seg_pns_bf` name referenced? The "crawler" would need to look into old commits and check for hardcoded model names, but it sounds overkill to me. For example, this command:
```bash
git log --pretty=short -u -L 12,14:AxonDeepSeg/download_model.py
```
tells us that `default_SEM_model`, `default_TEM_model` and `model_seg_pns_bf` were resp. changed for `model_seg_rat_axon-myelin_sem`, `model_seg_mouse_axon-myelin_tem` and `model_seg_rat_axon-myelin_bf`. However, I don't think this is very robust.

hermancollin: Maybe we could add a note in https://github.com/axondeepseg/axondeepseg/wiki/Instructions-on-How-to-Update-Models-in-Repo-and-Github-Storage that the doc needs to be updated after a model name change but will the next person that updates a model read this wiki?

mariehbourget: Hi @hermancollin, sorry for the delay.
As a first step, I think we should add the mention "formely called ..." for our models names in the documentation.
In fsleyes for example, the displayed name changed between v3 and v4 and can create confusion for the user following the video tutorial (ex: in this [comment](https://github.com/axondeepseg/axondeepseg/discussions/614#discussion-3940183)).
In addition, we are publishing a paper with the "old name" of the BF model and it would be best to document it (if I recall correctly, this is what triggered the issue #573 in the first place).

hermancollin: I added the BF model "legacy" name.

mariehbourget (approved):
Looks good! Thanks @hermancollin!

PR #622, opened 2022-03-22
Fix morphometrics bug when writing index/axonmyelin_index files

### LINKED ISSUE #553: CLI morphometrics doesn't generate index images
**Describe the bug**
Using the `axondeepseg_morphometrics` script and specifying folders with the `-i` option produces an unwanted behavior.
1. we get the following message (which is peculiar because the xlsx file ___is___ saved and look normal)
```bash
Cannot save morphometrics data in file 'paasu_img2_axon_morphometrics.xlsx'.`
```
2. "index" images are not generated

The first point was reported by @mathieuboudreau last week when he tried to run morphometrics via the CLI with subfolders, similar to what is described here.

When running the script directly in the folder containing the segmentation, we do not encounter this problem and the index files are generated as expected. 

**Platform (please complete the following information):**
 - Linux 5.10.52-3-rt47-MANJARO
 - AxonDeepSeg v3.2.0

hermancollin:
## Checklist

## Description

Tying up some loose ends here. This PR addresses a bug when using the `axondeepseg_morphometrics` command with the `-i` arg pointing to a *folder*. In these cases, the xlsx/csv file was correctly written but the index + axonmyelin_index images could not be saved. Note that when launching the script from the actual directory and using `-i .`, this behavior does not manifest.

Also, the related exception was poorly handled and actually ignored. Not sure if this is the desired behavior (maybe when doing morphometrics for multiple files it's useful to ignore the exception and go on to the next folder?) I changed it but now the catch block is useless.

## Linked issues

Resolves #553

### DIFF
--- a/AxonDeepSeg/morphometrics/launch_morphometrics_computation.py
+++ b/AxonDeepSeg/morphometrics/launch_morphometrics_computation.py
@@ -238,8 +238,7 @@ def main(argv=None):
                     
                 print(f"Morphometrics file: {morph_filename} has been saved in the {str(current_path_target.parent.absolute())} directory")
             except IOError:
-                print("Cannot save morphometrics data in file '%s'." % morph_filename)
-                raise
+                print(f"Cannot save morphometrics data or associated index images for file {morph_filename}.")
 
         else: 
             print("The path(s) specified is/are not image(s). Please update the input path(s) and try again.")
--- a/AxonDeepSeg/morphometrics/launch_morphometrics_computation.py
+++ b/AxonDeepSeg/morphometrics/launch_morphometrics_computation.py
@@ -222,13 +222,16 @@ def main(argv=None):
                     pd.DataFrame(x).to_csv(current_path_target.parent / morph_filename, na_rep='NaN')
 
                 # Generate the index image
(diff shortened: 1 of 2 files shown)


Stoyan-I-A:
Looks good to me.

As for the exception, could there be a way to trigger it in a normal usage and see what it currently does?

hermancollin:
> Looks good to me.
> 
> As for the exception, could there be a way to trigger it in a normal usage and see what it currently does?

IIUC your question, one can call the morphometrics script and use the `-i` option with 2 or more folders. In that case, morphometrics will be computed and saved in the xlsx files, but the index and index_axonmyelin files will be missing. Also, the script will output the following statement for every folder:
```
Cannot save morphometrics data in file '%s'." % morph_filename
```

hermancollin:
I removed the `raise` statement I had previously added because I think it makes more sense to resume the computation if an error happens when writing one file, especially because it would stop batch morphometrics. 

Also improved the message a little bit in case something goes wrong, but I don't even know what could trigger it at this point.

mathieuboudreau (approved):
Looks great to me! 

I just added a test assertion to evaluate what was previously missed by this bug, to ensure we don't overlook it in the future.

hermancollin:
Thank you @mathieuboudreau for the tests, as always!

PR #629, opened 2022-04-09
Resolve "tif in filename" imread bug & refactor valid image exensions

### LINKED ISSUE #618: imread bug when "tif" is present in filename
**Describe the bug**

While debugging for #617, I found an issue with [ads_utils.py/imread](https://github.com/axondeepseg/axondeepseg/blob/ca4baad96bed0d9c84b2dc63a8c4b464dc4d5a2c/AxonDeepSeg/ads_utils.py#L254).

The condition to use tiff-pil to open TIF files is `if 'tif' in str(filename):`.
This fails when the file is not a tif, but the letter `tif` are present anywhere in the path of the file.
We should use a "endswith" condition instead (for `.tif `and `.tiff`) or an exact match after splitting the filename at the last dot.

Note: this can be addressed independently from #617 on ADS side.

**To Reproduce**
- Rename any image (png or tif 8bit) with `tif` in the filename, ex: `tifabcd.png`
- Run the segmentation with ADS CLI.
- The error message comes from `merge_masks` which is trying to use `imread` to read the "tifabcd_seg-axon.png" and "tifabcd_seg-myelin.png" segmentation files but fails because it's using tif-pill to open a PNG.

```
  File "/home/mhbourget/miniconda/envs/ads_venv_v4/lib/python3.8/site-packages/imageio/core/imopen.py", line 188, in imopen
    return loader(request, **kwargs)
  File "/home/mhbourget/miniconda/envs/ads_venv_v4/lib/python3.8/site-packages/imageio/core/imopen.py", line 177, in loader
    return config.plugin_class(request, **kwargs)
  File "/home/mhbourget/miniconda/envs/ads_venv_v4/lib/python3.8/site-packages/imageio/config/plugins.py", line 108, in partial_legacy_plugin
    return LegacyPlugin(request, legacy_plugin)
  File "/home/mhbourget/miniconda/envs/ads_venv_v4/lib/python3.8/site-packages/imageio/core/legacy_plugin_wrapper.py", line 69, in __init__
    raise InitializationError(
imageio.core.request.InitializationError: `TIFF-PIL` can not read `/home/mhbourget/Documents/20220316-16bit/tifabcd_seg-axon.png`.

  File "/home/mhbourget/miniconda/envs/ads_venv_v4/bin/axondeepseg", line 33, in <module>
    sys.exit(load_entry_point('AxonDeepSeg', 'console_scripts', 'axondeepseg')())
  File "/home/mhbourget/code/axondeepseg/AxonDeepSeg/segment.py", line 276, in main
    segment_image(
  File "/home/mhbourget/code/axondeepseg/AxonDeepSeg/segment.py", line 112, in segment_image
    axon_segmentation(path_acquisitions_folders=path_acquisition, acquisitions_filenames=[str(path_acquisition / acquisition_name)],
  File "/home/mhbourget/code/axondeepseg/AxonDeepSeg/apply_model.py", line 56, in axon_segmentation
    merge_masks(
  File "/home/mhbourget/code/axondeepseg/AxonDeepSeg/visualization/merge_masks.py", line 12, in merge_masks
    axon = ads.imread(path_axon)
  File "/home/mhbourget/code/axondeepseg/AxonDeepSeg/ads_utils.py", line 259, in imread
    raw_img = imageio.imread(filename, format='tiff-pil')
  File "/home/mhbourget/miniconda/envs/ads_venv_v4/lib/python3.8/site-packages/imageio/__init__.py", line 86, in imread
    return imread_v2(uri, format=format, **kwargs)
  File "/home/mhbourget/miniconda/envs/ads_venv_v4/lib/python3.8/site-packages/imageio/v2.py", line 159, in imread
    with imopen(uri, "ri", plugin=format) as file:
  File "/home/mhbourget/miniconda/envs/ads_venv_v4/lib/python3.8/site-packages/imageio/core/imopen.py", line 206, in imopen
    raise err_type(err_msg) from err_from
RuntimeError: `tiff-pil` can not handle the given uri.
```

mathieuboudreau:
## Checklist

## Description
Resolves #618, "tif in filename" imread bug & refactor valid image exensions

## Linked issues

#618


mathieuboudreau:
Hum, I think I may not branched out of the master branch, which is why there are so many commits; my bad.

Let me see if I can easily resolve it without having to recreate a PR.

mathieuboudreau:
Ok, so merging master fixed it thankfully; the changes reflected in this PR/branch start at commit 
8053f8a, and a squash merge will clean it when we merge to master. Sorry about that!

hermancollin:
LGTM. Only thing was that the plugin still used a local valid_extensions list (slightly different from the new one). Now it's all centralized with config.py's valid extensions.

hermancollin (approved):

AxonDeepSeg/ads_utils.py
```diff
@@ -254,7 +257,18 @@ def convert_path(object_path):
 def imread(filename, bitdepth=8):
     """ Read image and convert it to desired bitdepth without truncation.
     """
-    if 'tif' in str(filename):
+
+    # Convert to Path
+    filename = Path(filename)
+
+    file_ext = filename.suffix.lower()
```
mariehbourget, line 264: This will return ".tif" for both ".tif" files and ".ome.tif" files.
We currently don't support OME-TIF files in ADS (nor ivadomed) with imageio/pillow.
However OME-TIF is one of the standard format in BIDS so they are likely to be found in microscopy datasets.

To avoid having an OME-TIF mistaken for a TIF in ivadomed, we use an "ordered" list of extensions that returns the first extension that match the end of the filename [here](https://github.com/ivadomed/ivadomed/blob/7b76bf81a025cde3096fd1d686d6f3c0b8ce8f02/ivadomed/loader/utils.py#L404).

hermancollin: In that case we should use a function like in ivadomed because this pattern of suffix extraction is everywhere.

mathieuboudreau: > ".ome.tif"

I was not aware that we had to provide support for this file extension (I hate double file extensions! but I get it).

Are there any other ones that we need to consider that are not or have never been in our valid extensions list? "ome.tiff", "ome.png", etc?

mariehbourget: > I was not aware that we had to provide support for this file extension (I hate double file extensions! but I get it).

Yeah, double extensions are not fun.
The implementation in this PR works as intended with what we discussed before, I just thought about that corner case yesterday while reviewing it.

In ivadomed we don't support "ome" files either at the moment but having them in the list of extension is a way to identify them correctly before they enter the "traditional tif" pipeline. If the file is identified as an "ome", it currently raise an error [here](https://github.com/ivadomed/ivadomed/blob/7b76bf81a025cde3096fd1d686d6f3c0b8ce8f02/ivadomed/loader/segmentation_pair.py#L284).

> Are there any other ones that we need to consider that are not or have never been in our valid extensions list? ".ome.tiff", ".ome.png", etc?

We could use the same list as ivadomed [here](https://github.com/ivadomed/ivadomed/blob/7b76bf81a025cde3096fd1d686d6f3c0b8ce8f02/ivadomed/loader/utils.py#L35) except `.nii` and `.nii.gz`.

mathieuboudreau: @mariehbourget I've resolved this issue you raised in commits [2fe92b7](https://github.com/axondeepseg/axondeepseg/pull/629/commits/2fe92b723e9e45a531257643d05ac759ac904cf2)  [084a386](https://github.com/axondeepseg/axondeepseg/pull/629/commits/084a386f293fa83894a823fcfbe4af0c8ca3d025) [5cb822a](https://github.com/axondeepseg/axondeepseg/pull/629/commits/5cb822a41df3d66c3b4a938229d3b6162af08ec2)

In particular, here is a working example of the filename extension extraction I implemented in these lines

https://github.com/axondeepseg/axondeepseg/blob/5cb822a41df3d66c3b4a938229d3b6162af08ec2/AxonDeepSeg/ads_utils.py#L261-L266


```python
Python 3.9.12 (main, Mar 26 2022, 15:51:15) 
[Clang 13.1.6 (clang-1316.0.21.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from pathlib import Path
>>> filename = Path('foo.OME.tif')
>>> print(''.join(filename.suffixes).lower())
.ome.tif
>>> 
```

mathieuboudreau:
@hermancollin @mariehbourget I think this PR is ready for review again!

hermancollin:
Everything LGTM here. @mathieuboudreau Should we add a warning message of some sort if `ome` is in the suffix list saying the format is unsupported for now? Will ivadomed warn the user?

AxonDeepSeg/ads_utils.py
mariehbourget: I completely missed the notification 13 days ago 😬, my apologies for the delay.

At first glance, I really like this method compared to what I have done in ivadomed. Unfortunately, I think I found a bug here where:
if there are "dots" in the filename, they will be picked up as file extensions, ex:
```
>>> filename = Path('foo.bar.tif')
>>> print(''.join(filename.suffixes).lower())
.bar.tif
```
This unfortunately happens often with non-BIDS filenames containing all sorts of illegal characters like:
`mouse Nerve-1.4NA-63x.tif`.

Let me know your thoughts.

mathieuboudreau: >This unfortunately happens often with non-BIDS filenames containing all sorts of illegal characters like:
mouse Nerve-1.4NA-63x.tif.

>Let me know your thoughts.

My thoughts are we shouldn't support filenames with dots in them 😅 , what a mess that would cause 😬. Dots in filenames are meant to delineate the suffixes, so if a user came to me with a dataset like that, at most I'd suggest a script that renames it (replacing the non-suffix related dot with a dash).

But that's my view - do we have many datasets already like that? If it's an edge case (eg, one user), I don't think we should support it.

Otherwise, I'm open to a suggestion as to how to extract suffixes *multiple dots* (i.e. `.ome.tiff`) to check if they are in the filename, but not random non-suffix dots (i.e. your `.4NA-63x.tif` example above). We can't just parse the list of extensions and check if they are in the filenames (eg, `is 'ome.tiff' in filename`), because that's what caused the bug in the first place (e.g, `is tif in filename ` will be true for something like `tiffany.txt`)

mariehbourget: > But that's my view - do we have many datasets already like that? If it's an edge case (eg, one user), I don't think we should support it.

Yeah my example came from a real user. So it happened in the past but I don't know to which extent.

> Otherwise, I'm open to a suggestion as to how to extract suffixes _multiple dots_ (i.e. `.ome.tiff`) to check if they are in the filename, but not random non-suffix dots (i.e. your `.4NA-63x.tif` example above). We can't just parse the list of extensions and check if they are in the filenames (eg, `is 'ome.tiff' in filename`), because that's what caused the bug in the first place (e.g, `is tif in filename ` will be true for something like `tiffany.txt`)

Here is how we do it in ivadomed which seems to work: https://github.com/ivadomed/ivadomed/blob/b9a208d79254abce847ae9d522fea376057ce2b3/ivadomed/loader/utils.py#L408

But we would have to add an error, as @hermancollin suggested if the extension ends up being some sort of "ome" because we don't support it. (same for ivadomed: https://github.com/ivadomed/ivadomed/blob/b9a208d79254abce847ae9d522fea376057ce2b3/ivadomed/loader/segmentation_pair.py#L283)

mathieuboudreau: @mariehbourget thanks a lot for these resources! I've integrated them into our utils file and modified them slightly, and wrote tests for them as well. Please let me know if with these changes you think this PR is ready!

ads_plugin.py
```diff
@@ -313,7 +313,6 @@ def on_load_png_button(self, event):
 
         # Check if the image format is valid
         image_extension = in_file.suffix
-        valid_extensions = [".png", ".tif", ".jpg", ".jpeg"]
```
mariehbourget, line 316: I'm unsure about this.

If the file is `.ome.tif`, it won't raise an error in the plugin itself.
It will however raise an error later in `imread` but the error message will only be displayed in the terminal and not in the plugin,

Should we apply the same strategies than in `imread` here? Both for fetching the extension and for the error message? or is the error in the terminal good enough?

ads_plugin.py
```diff
@@ -346,7 +345,6 @@ def on_load_mask_button(self, event):
...
```
mariehbourget, line 349: Same as above

AxonDeepSeg/ads_utils.py
```diff
@@ -254,7 +257,21 @@ def convert_path(object_path):
...
+
+    # Convert to Path
+    filename = Path(filename)
+
+    # Get list of all suffixes in file, join them into a string, and then 
+    # lowercase to set the file extention to check against valid extension.
+    file_ext = get_file_extension(filename)
+
+    if (not file_ext) or ("ome" in file_ext):
+            raise IOError(f"The input file extension '{file_ext}' of '{Path(filename).stem}' is not "
```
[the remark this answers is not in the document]
mathieuboudreau: Agreed!

ads_plugin.py
mathieuboudreau: This sounds like a plugin issue (that it doesn't throw an error for ome.tif) and not an issue related to the specific aim of this PR, so I would say leave it the current way and open a new issue to handle the problem you raised about the plugin.

AxonDeepSeg/segment.py
```diff
@@ -448,7 +439,7 @@ def main(argv=None):
                     "pixel_size_in_micrometer.txt file containing the pixel size value."
         if not current_path_target.is_dir():
 
-            if current_path_target.suffix.lower() in validExtensions:
+            if current_path_target.suffix.lower() in valid_extensions:
```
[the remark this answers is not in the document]
mathieuboudreau: Done in [bd03c67](https://github.com/axondeepseg/axondeepseg/pull/629/commits/bd03c67d261eadc6a00ca535a9a7b1752c2fd9d1)

AxonDeepSeg/ads_utils.py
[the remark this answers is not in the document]
mariehbourget: Ok thanks, I'll take care of it in ivadomed as well.

mariehbourget (approved):
LGTM, thanks @mathieuboudreau for taking care of this!

PR #631, opened 2022-04-26
Auto-update ivadomed with `pip install -e .`

### LINKED ISSUE #623: Instructions on how to update ADS won't update the Ivadomed package
While testing #620, I encountered an error that was previously fixed on the Ivadomed side. This was because I didn't update the Ivadomed package in my environment.

Currently, in our docs, it says that we can update ADS by using these commands:
```
cd axondeepseg
git pull
pip install -e .
```

However, those instructions won't update the Ivadomed package. The user would need to do an additional command:
`pip install ivadomed==<The latest version supported by ADS>`.

Question: Will ADS always support the latest Ivadomed version? In this case, we would just need to add `pip install ivadomed` to the instructions. But if we insist on installing a fixed version, the user would have to check in the environment.yml file which version to install, which might be a bit cumbersome for the user.

hermancollin:
## Checklist

## Description

Extracts ivadomed's semantic version and runs an additionnal `pip install` command to update it at the end of the installation process. This way, when a user updates ADS using `pip install -e .`, the ivadomed version will be updated as well.

## Linked issues

Resolves #623

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,7 @@ import atexit
 
 from codecs import open
 from os import path
+import platform
 import re
 
 import AxonDeepSeg
@@ -20,8 +21,14 @@ def update_ivadomed():
     """Ivadomed is a pip package: needs manual update"""
     with open('environment.yml', 'r') as env_file:
         lines = env_file.read()
-        ivadomed_version = re.search("ivadomed==\d+.\d+.\d+", lines)
-        run(["pip", "install", ivadomed_version.group()])
+        ivadomed_version = re.search(r"ivadomed==\d+\.\d+\.\d+", lines).group()
+        if platform.system() == "Darwin":
+            torch_version = re.search(r"torch==\d+\.\d+\.\d+(?!\+)", lines).group()
+            torchvision_version = re.search(r"torchvision==\d+\.\d+\.\d+(?!\+)", lines).group()
+        else:
+            torch_version = re.search(r"torch==\d+\.\d+\.\d+\+cpu", lines).group()
+            torchvision_version = re.search(r"torchvision==\d+\.\d+\.\d+\+cpu", lines).group()
+        run(["pip", "install", ivadomed_version, torch_version, torchvision_version])
 
 class PostDevelopCommand(develop):
     """Post-installation for installation mode."""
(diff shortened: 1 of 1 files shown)


hermancollin:
> Should we do this for all our pip packages? Maybe the function(s) could generalize to do so?

The other pip dependencies are
```
    - opencv-contrib-python
    - opencv-python-headless
    - torch==1.8.1+cpu; sys_platform != "darwin"
    - torchvision==0.9.1+cpu; sys_platform != "darwin"
    - torch==1.8.1;sys_platform == "darwin"
    - torchvision==0.9.1; sys_platform == "darwin"
```
opencv packages aren't pinned so I guess we don't want to update them? Also IIUC the sys_platform check is because torch/torchvision on macOS don't have separate cpu/gpu versions?

mathieuboudreau:
> > Should we do this for all our pip packages? Maybe the function(s) could generalize to do so?
> 
> The other pip dependencies are
> 
> ```
>     - opencv-contrib-python
>     - opencv-python-headless
>     - torch==1.8.1+cpu; sys_platform != "darwin"
>     - torchvision==0.9.1+cpu; sys_platform != "darwin"
>     - torch==1.8.1;sys_platform == "darwin"
>     - torchvision==0.9.1; sys_platform == "darwin"
> ```
> 
> opencv packages aren't pinned so I guess we don't want to update them? Also IIUC the sys_platform check is because torch/torchvision on macOS don't have separate cpu/gpu versions?

I think one important note is that, automatically updating IVADOMED without updating torch/torch visions may introduce bugs, as it's possilbe that IVADOMED needs different versions of them in future updates, and this won't be communicated to us through your current PR (as they only document those requirements in their documentation, but they remain IVADOMED requirements nonetheless).

hermancollin:
In that case it makes sense to add the torch/torchvision auto update as well.

It works on my system but I'm curious to see if it also does for Darwin

Stoyan-I-A:
Command to update a conda environment:
(You need to deactivate the environment first)
`conda env update --name env_name --file environment.yml`

However, according to @hermancollin , this will not update the packages installed with pip

hermancollin:
Actually we should close this PR in favor of changing the doc to tell the user that if he wants to update ADS, he should reinstall the whole environment. [This comment](https://github.com/axondeepseg/axondeepseg/issues/623#issuecomment-1119118046) sums up why it's a bad idea to go forward with this solution.

PR #632, opened 2022-05-02
Zoom factor sweep

hermancollin:
## Checklist

## Description

The goal of this PR is to add a feature to repeatedly segment the same image using different zoom factors within a given range, as discussed in #610. This tool will be useful to evaluate an appropriate zoom factor for optimal segmentation visual aspect.

I added the feature in a new file but the sweep function will be called from the main `segment.py` CLI. For now, having this feature in a separate file does not make much sense because the code is pretty short but, eventually, when we add a metric to evaluate the masks produced, it will be helpful.

Current parameters: 
- `sweep_range` (e.g. [3, 5] for zoom factors between 3 and 5)
- `sweep_length`: number of equidistant zoom factors to sample in sweep_range

## Linked issues

#610

### DIFF
--- a/AxonDeepSeg/zoom_factor_sweep.py
+++ b/AxonDeepSeg/zoom_factor_sweep.py
@@ -1,0 +1,52 @@
+# Successively segments an image with zoom factors within a given range
+
+from pathlib import Path
+from AxonDeepSeg.segment import segment_image
+from config import axonmyelin_suffix, axon_suffix, myelin_suffix
+
+
+def sweep(
+    path_image,
+    path_model,
+    overlap_value,
+    sweep_range,
+    sweep_length,
+    acquired_resolution = None,
+    ):
+    """
+    Wrapper over segment_image to produce segmentations for zoom factor values within a given range.
+    :param path_image:          the path of the image to segment.
+    :param path_model:          where to access the model
+    :param overlap_value:       the number of pixels to be used for overlap when doing prediction. Higher
+                                value means less border effects but longer segmentation time.
+    :param sweep_range:         upper and lower bounds of the zoom factor range
+    :param sweep_length:        number of equidistant zoom factor values to sample from the range
+    :param acquired_resolution: isotropic pixel resolution of the acquired images.
+    :return: Nothing.
+    """
+
+    lower_bound, upper_bound = sweep_range
+    # create new directory to store segmentations
+    path_results = Path(path_image).parent.absolute() / 'sweep_results'
+    path_results.mkdir(parents=True, exist_ok=True)
+    path_seg_outputs = [
+        Path(path_image).stem + str(axon_suffix),
+        Path(path_image).stem + str(myelin_suffix),
+        Path(path_image).stem + str(axonmyelin_suffix),
+    ]
+
+    for i in range(sweep_length):
+        zoom_factor = lower_bound + i * (upper_bound - lower_bound) / sweep_length
+        segment_image(
+            path_image,
+            path_model,
+            overlap_value,
+            acquired_resolution,
+            zoom_factor,
+        )    
+        # move and rename segmentations
+        for path_seg in path_seg_outputs:
+            path = Path(path_seg)
+            path.rename(path_results / Path(path.stem + f'_zf-{zoom_factor}.png'))
+
+        print(f"Done with zoom factor {zoom_factor}.")


hermancollin:
(Just want to point out that I changed the look of the main segmentation CLI's `add_argument()` expressions to be a bit more pep8 compliant. Tell me which version you think is better. Personnally, the way it was layed out before hurt my eyes)

hermancollin:
Ok this feature now works in the CLI. Before updating the doc and adding tests, I would like some feedback.

The way it works is you specify zoom factor lower/upper bounds with the `-r` option and the number of points to sample with the `-l` option. For example, the following:
```
axondeepseg -t SEM -i image.png -s 0.37 -r 1 2 -l 5
```
will create a new folder called *image_sweep* and compute segmentations for 5 equidistant zoom factor values in the range **[ 1, 2 [** (i.e. ***1.0***-***1.2***-***1.4***-***1.6***-***1.8***).

AxonDeepSeg/zoom_factor_sweep.py
```diff
@@ -0,0 +1,53 @@
...
+            path_image,
+            path_model,
+            overlap_value,
+            acquired_resolution,
+            zoom_factor,
+        )    
+        # move and rename segmentations
+        for path_seg in path_seg_outputs:
+            path = Path(path_seg)
+            path.rename(path_results / Path(path.stem + f'_zf-{zoom_factor}.png'))
```
mariehbourget, line 50: This is not working for me.
I ran this CLI:
`axondeepseg -t SEM -i AxonDeepSeg/models/model_seg_rat_axon-myelin_sem/data_test/image.png -s 0.13 -r 1 2 -l 5`

and I get the following error:
```
FileNotFoundError: [Errno 2] No such file or directory: 'image_seg-axon.png' -> '/home/mhbourget/code/axondeepseg/AxonDeepSeg/models/model_seg_rat_axon-myelin_sem/data_test/image_sweep/image_seg-axon_zf-1.0.png'
```
The folder image_sweep is empty and the segmentations files from the first zoom factor are in the parent folder (data_test) with their original name (i.e. without the zf part).

mariehbourget (commented):
Thanks @hermancollin, the code looks good and the feature will be a great addition.
However, I was not able to test it properly, something about the "move and rename" is not working for me (see comment).

AxonDeepSeg/zoom_factor_sweep.py
hermancollin: Good catch. It should be fixed now.

AxonDeepSeg/segment.py
```diff
@@ -363,6 +410,15 @@ def main(argv=None):
     else:
         zoom_factor = 1.0
 
+    # check for sweep mode
+    sweep_mode = (args["sweeprange"] is not None) & (args["sweeplength"] is not None)
+    if sweep_mode:
+        sweep_range = [float(args["sweeprange"][0]), float(args["sweeprange"][1])]
+        sweep_length = int(args["sweeplength"])
+        if len(path_target_list) != 1:
+            print("ERROR: Please use a single image for zoom factor sweeps.")
```
mariehbourget, line 419:
```suggestion
            print("ERROR: Please use a single image for zoom factor sweep.")
```

AxonDeepSeg/segment.py
```diff
@@ -419,6 +484,9 @@ def main(argv=None):
                 break
 
         else:
+            if sweep_mode:
+                print("ERROR: Please use a single image for zoom factor sweeps.")
```
mariehbourget, line 488:
```suggestion
                print("ERROR: Please use a single image for zoom factor sweep.")
```

mariehbourget (commented):
Thanks @hermancollin, "move and rename" is working as expected now.
Everything looks good to me and I like the new version of the CLI's add_argument() expressions.
Great work!

mariehbourget (approved):
LGTM, thanks @hermancollin!

mathieuboudreau:
Wow, excited to see this issue get completed while I was away for ISMRM! Thanks a lot @hermancollin  and @mariehbourget !

Looking at this quickly, the only suggestion I'd have is to write in a check to ensure that the [min, max] zoom range does not exceed the limit set by the patch size (so maybe if we zoom in too much, reset the minimum to that limit. I calculate it elsewhere in the code already, here: https://github.com/axondeepseg/axondeepseg/blob/e7a9373836efa6afedf9393de8fc2109737fcfaa/AxonDeepSeg/segment.py#L269-L291)

hermancollin:
Thank you both @mariehbourget @mathieuboudreau for these comments.
 
> Looking at this quickly, the only suggestion I'd have is to write in a check to ensure that the [min, max] zoom range does not exceed the limit set by the patch size (so maybe if we zoom in too much, reset the minimum to that limit.

So IIUC we would check the minimum zoom factor for the given image before starting the zoom sweep processing, and then tell the user if one of the bounds is invalid? This sounds like a good idea.

For the record, if we try to segment with an invalid zoom factor during the sweep, the `segment_image` function will do the check anyway so there would be some redundance.

mathieuboudreau:
> So IIUC we would check the minimum zoom factor for the given image before starting the zoom sweep processing, and then tell the user if one of the bounds is invalid? This sounds like a good idea.

Yeah kinda. I would just inform the user and tell them that we internally reset the range so it doesn't go beyond it. Otherwise, there's a risk (unsure if it would happen) that the "optimal zoom value" falls beyond the valid range, and that would be a headache to explain.

mathieuboudreau:
>For the record, if we try to segment with an invalid zoom factor during the sweep, the segment_image function will do the check you wrote anyway so there would be some redundance, albeit not much resource consumption.

Ah gotcha. It's still better to automatically change the range so that it only contains valid zoom values though I think!

mathieuboudreau (approved):
Looks great to me - just tested a few ranges and it works well! Thanks a lot @hermancollin !

I just modified the example slightly to have a wider range, in order to see that the value 1.0 truly is optimal for this case (the other time, it only started at this value, so we didn't know if a lower zoom factor might improve it more or not).

PR #645, opened 2022-06-03
Add border_touching flag to morphometrics

### LINKED ISSUE #638: Add morphometrics flag for border touching axons
**Describe the problem**
As discussed recently in #314, it would be useful to have an additional column in the morphometrics file that tells the user if an axon is touching the border of the image, mainly for easier cleanup when selecting fibers to keep in the analysis. The issue had already been raised by previous collaborators that wanted to remove partial axons, often located on the image border.

**Proposed solution**
very straightforward using `sklearn.measures.regionprops.bbox` and the image dimensions 

We have to wait for #636 to be merged first

hermancollin:
## Checklist

## Description

This PR adds a flag in the morphometrics to tell the user if the axonmyelin object touches a border for faster manual filtering. Thanks to the morph. data structure refactoring, this was very quick to implement.

I tested the PR on the following test image:
test_img_axonmyelin_index
According to the new `border_touching` column, the elements touching the border are the following:
```
0 1 2 3 4 7 9 12 13
```
As we can see, most of these are unusable except for id=3 where the element touches the border but is still good, so it would still require the user to take a look at the result.

## Linked issues

Resolves #638

### DIFF
--- a/AxonDeepSeg/download_tests.py
+++ b/AxonDeepSeg/download_tests.py
@@ -10,7 +10,7 @@ def download_tests(destination=None):
         destination = convert_path(destination)
         test_files_destination = destination / "__test_files__"
 
-    url_tests = "https://github.com/axondeepseg/data-testing/archive/refs/tags/r20220607a.zip"  
+    url_tests = "https://github.com/axondeepseg/data-testing/archive/refs/tags/r20220522a.zip"  
     files_before = list(Path.cwd().iterdir())
 
     if (
--- a/AxonDeepSeg/download_tests.py
+++ b/AxonDeepSeg/download_tests.py
@@ -10,7 +10,7 @@ def download_tests(destination=None):
         destination = convert_path(destination)
         test_files_destination = destination / "__test_files__"
 
-    url_tests = "https://github.com/axondeepseg/data-testing/archive/refs/tags/r20220522a.zip"  
+    url_tests = "https://github.com/axondeepseg/data-testing/archive/refs/tags/r20220607a.zip"  
     files_before = list(Path.cwd().iterdir())
 
     if (
--- a/AxonDeepSeg/morphometrics/compute_morphometrics.py
+++ b/AxonDeepSeg/morphometrics/compute_morphometrics.py
@@ -77,6 +77,8 @@ def get_axon_morphometrics(im_axon, path_folder=None, im_myelin=None, pixel_size
     # Measure properties for each axon object
     axon_objects = measure.regionprops(im_axon_label)
 
+    im_shape = im_axon.shape
+
     # Deal with myelin mask
     if im_myelin is not None:
 
@@ -150,7 +152,8 @@ def get_axon_morphometrics(im_axon, path_folder=None, im_myelin=None, pixel_size
                 'myelin_thickness': np.nan,
                 'myelin_area': np.nan,
                 'axonmyelin_area': np.nan,
-                'axonmyelin_perimeter': np.nan
+                'axonmyelin_perimeter': np.nan,
+                'border_touching': False
             }
             stats.update(myelin_stats)
             # Find label of axonmyelin corresponding to axon centroid
@@ -183,6 +186,11 @@ def get_axon_morphometrics(im_axon, path_folder=None, im_myelin=None, pixel_size
                     stats['myelin_area'] = np.nan
                     stats['axonmyelin_area'] = np.nan
                     stats['axonmyelin_perimeter'] = np.nan
+
+                # check if bounding box touches a border (partial axonmyelin object)
+                bbox = prop_axonmyelin.bbox
+                if 0 in bbox[:2] or bbox[2] == im_shape[0] or bbox[3] == im_shape[1]:
+                    stats['border_touching'] = True
 
             else:
                 print(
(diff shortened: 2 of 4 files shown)


Stoyan-I-A:
Should we add another column for axon-touching border? (To determine if the axon is touching the border and not just the myelin)

hermancollin:
> Should we add another column for axon-touching border? (To determine if the axon is touching the border and not just the myelin)

While it's true that border touching *axons* are technically a separate case, I'm not sure it woud add any information. For a good segmentation mask (that would be worth computing morphometrics for and manually filtering the results), if an axon touches a border, its associated myelin object will also touch the border, since the axonmyelin object should contain the entire axon.

akraskovSMDB:
@hermancollin 
Would it be possible to output bounding box coordinates?

One reason is that it will make it easier to select axons if only a part of the image is suitable/used for analysis (can rerun morphometrics with a cropped images but it is a bit more cumbersome)
Second reason stems from stereology conventions (to avoid double counting) when axons touching lower and left bound are ignored but the one touching upper and right border are taken into account, see attached image with a analysis box and considered axons (indicated with diameters).

Having bounding box coordinates will allow to select axons according to a given rule (e.g. stereology rule)

image

akraskovSMDB:
@hermancollin thank you for the response.
Sure, you can make an additional png with a bounding box, it might be useful for visualisation. But for me it will be more useful to have coordinates of the bounding box in the morphometrics excel output file. I will write my own parsing script to select axons according to a rule, for example the one I described earlier, other people can use this information in a different way.


A separate issue. What is holding a merge of this change?

Stoyan-I-A:
> A separate issue. What is holding a merge of this change?

I need to review it. I will do so in the upcoming days.

hermancollin:
@akraskovSMDB hmm I see. In this case, I think this is what we should do:
- wait for this PR to be merged. In the meantime, I'm pretty sure you could access this feature by going in the *axondeepseg* directory and checking out this branch:
```
git checkout ac/638_morph_flag
```
- I will open a separate PR to add more features to this. I'm thinking of making it an optional argument for the morphometrics script. For example, specifying the `--border-info` flag would do the following:
--  add border_touching flag
-- add bounding box coords
-- potentially save bounding boxes overlay image if it doesn't introduce too much overhead

mathieuboudreau:
Ok, tried it with another script and it works:

<details>

```python

# coding: utf-8

from pathlib import Path
import imageio
import numpy as np
import os
from skimage.transform import resize
import pytest

from AxonDeepSeg.visualization.simulate_axons import SimulateAxons, calc_myelin_thickness
from AxonDeepSeg.visualization.get_masks import get_masks

# Axon 1
x_pos = 100
y_pos = 100

axon_radius = 100
gratio = 0.6
myelin_thickness = calc_myelin_thickness(axon_radius, gratio)

obj = SimulateAxons()

obj.generate_axon(
    axon_radius=axon_radius, center=[x_pos, y_pos], gratio=gratio, plane_angle=0
)

# Axon 2
x_pos = 500
y_pos = 500

axon_radius = 40
gratio = 0.6
myelin_thickness = calc_myelin_thickness(axon_radius, gratio)

obj.generate_axon(
    axon_radius=axon_radius, center=[x_pos, y_pos], gratio=gratio, plane_angle=0
)

# Save the mask (which is what we really simulated)
obj.save('image_seg-axonmyelin.png')

# Save the axon and myelin masks separately
get_masks('image_seg-axonmyelin.png') # the file will be saved

# Save a file called "image", for the pupose of calling the axondeepseg_morphometrics command (shouldn't be needed, but it is)
obj.save('image.png')



```

</details>

image_seg-axonmyelin

`axondeepseg_morphometrics -i image.png -s 0.1`
Screen Shot 2022-06-09 at 4 05 48 PM

mathieuboudreau (changes requested):
Feature looks good, please

* Change the name of the metric to "image_border_touching" instead
* Add a test(s) to test the feature, you can reuse my simulation code if you like

hermancollin:
Btw for the tests I made a new release of `data-testing` with an additionnal column in the morphometrics consistency check. However, we can disregard this release because when I'll work on the next PR I'll revert it back to the previous one and add more dedicated tests. This is because the column added in the current PR will be optional, triggered by the `--border-info` flag, in the next iteration.

mathieuboudreau:
>I can add the image you provided in the next design iteration because we'll have more things to test on it.

You don't need to add the image to `data-testing`, you can actually just run the simulation code within the test, so it can be done in this PR without having to upload anything.

mathieuboudreau:
woops misclick, sorry

mathieuboudreau:
>You don't need to add the image to data-testing, you can actually just run the simulation code within the test, so it can be done in this PR without having to upload anything.

If you're not sure how this would be done, just let me know and I'll do it.

hermancollin:
> > You don't need to add the image to data-testing, you can actually just run the simulation code within the test, so it can be done in this PR without having to upload anything.
> 
> If you're not sure how this would be done, just let me know and I'll do it.

Ahhh I didn't see your image was simulated. Ok I'll add it directly then.

hermancollin:
Ok I had to revert the test data release so that the consistency check pass. As previously mentioned, we'll add more tests in the next PR. @mathieuboudreau now the changes you suggested are added.

mathieuboudreau (approved):
Excellent - thank you @hermancollin for the quick turnaround!

Issue #314, opened 2020-06-25
Stereology Measures


amoebacodes:
I have noticed that in the AxonDeepSeg paper, a grid is applied to small 512 x 512 pixels patches to overcome "border issues", meaning some axons land on the edges and thus are not properly analyzed. However, it is an issue still for axons on the edge of an entire image. 

Stereology measures could be utilized to circumvent this issue. As shown in the picture in the link below, a central area of the image can be carved out for analysis, whereas the crust can be used to assist analysis of axons that are on the border.
http://www.stereology.info/counting-rules/

Thank you for reading this and for considering this suggestion!

mariehbourget:
Hi @amoebacodes,
Thanks for bringing that method up to our attention.

In AxonDeepSeg, we use an overlap to properly analyse axons landing on the edges of patches. That overlap acts as the crust from your example with stereology measures, and assist analysis on the border of the central area of patches.
For example, with a patch of 512x512 and an overlap of 25, the segmentation is performed on the entire patch of 512x512 for context, but only the 462x462 central area is output without the overlap. When all patches are analyzed, the 462x462 patches are then stitches together in the final segmentation (the overlap being used only for patches at the very edges of the image).

Let us know if you have more questions about patches behavior or if I misinterpreted your suggestion.

PR #650, opened 2022-06-13
`--border-info` option for morphometrics (optical fractionator technique)

### LINKED ISSUE #647: Add `--border-info` morphometrics flag
Essentially, this comes from user feedback while working on #645. This flag would trigger a *border-info* mode which would do this:
- add the `border_touching` boolean column in the morph. file implemented in #645
- add the axonmyelin's bounding box coordinates
- save the original image with bounding boxes overlayed onto it (if it doesn't impact the morphometrics runtime too much). This would mostly be useful because it would give the user more insight on the way the morphometrics script "sees" the axons.

The only issue is that we'll need to add a parameter to `get_axon_morphometrics`, which will slightly change the API so we'll have to go over the notebooks too.

hermancollin:
## Checklist

## Description
This PR expands on a recent feature request to add more information to the morphometrics output files. These are the new columns:
- `image_border_touching`: True if axonmyelin object touches any of the image border
- `bbox_min_row` + `bbox_min_col` + `bbox_max_row` + `bbox_max_col`: coordinates of the axonmyelin's bounding box corners

We could also add a column that directly tells the user if an axonmyelin object should be included in a count, assuming the whole image is a counting frame. It would simply require to compare the axonmyelin bounding box against the exclusion line (e.g. red line below)
count 

For example, in the following example image, this column would mark every axons as True except for 7, 12 and 13
example
Maybe @akraskovSMDB has some thoughts about this?
Also, if we added this "stereology count", I would still keep the previous columns so the user can perform further processing (e.g. in 3D, the exclusion line will not always be the same depending on which slice of the optical fractionator box we are on).

More info about this method [here](https://www.jove.com/t/55737/the-optical-fractionator-technique-to-estimate-cell-numbers-rat-model)

## Linked issues
Resolves #647 
Originally suggested in #314


hermancollin:
Initially, I wanted this morphometrics option to also output an image showing the bounding boxes, but I feel like the overhead it would introduce (mostly for huge datasets) is not necessary. The user can always process the morphometrics file to create himself this visualization. For example, this tiny 22-lines script:
```python
from AxonDeepSeg import ads_utils
from AxonDeepSeg.morphometrics.compute_morphometrics import load_axon_morphometrics
import numpy as np

img = ads_utils.imread('test_img_seg-axonmyelin.png')
morph_df = load_axon_morphometrics('test_img_axon_morphometrics.xlsx')
for i, row in morph_df.iterrows():
    bbox = [
        row["bbox_min_row"],
        row["bbox_min_col"],
        row["bbox_max_row"],
        row["bbox_max_col"],
    ]
    img[bbox[0]:bbox[2]-1, bbox[1]] = 255
    img[bbox[0]:bbox[2]-1, bbox[3]-1] = 255
    img[bbox[0], bbox[1]:bbox[3]-1] = 255
    img[bbox[2]-1, bbox[1]:bbox[3]-1] = 255

ads_utils.imwrite('test_img_bbox.png', img)

```
would output this:
test_img_bbox

hermancollin:
In the absence of more input from the user who suggested this, I think this feature is ready. Post-processing a segmentation to account for the counting rule should be fairly easy with these morphometrics columns (i.e. including every axons where `image_border_touching==True` as well as those who are not touching the exclusion line.

I'll add tests and update the doc.

akraskovSMDB:
Hi guys, could you please review and approve this change, thank you

mathieuboudreau (approved):
Tested the feature and works for me! Code looks good too, thanks @hermancollin !

PR #667, opened 2022-07-14
Save instance segmentation image

### LINKED ISSUE #662: Instance segmentation feature request
Per this discussion

### Discussed in https://github.com/axondeepseg/axondeepseg/discussions/661

<div type='discussions-op-text'>

<sup>Originally posted by **mcib3d** July  6, 2022</sup>
Hi and thanks for this wonderful tool,

I would eventually perform more analysis on the detected objects axons and myelins. I can  easily get a labelled image with the axons since they are separated. Sometimes two myelins touch, so my questions are : 

- how are they separated in the analysis ? It is a simple watershed ?
- can we get access to the _hidden_ separated myelins ?

Many thanks.

Thomas</div>

hermancollin:
## Checklist

## Description
This PR adds a functionality to save the instance segmentation obtained at the post-processing step (morphometrics stage, after the watershed segmentation). My goal is to have a color generator that takes as input a list of colors and colors the instance segmentation map such that 2 adjacent objects are never assigned the same color. This flexibility will allow us to change the color palette easily later on if we want to output other types of visualization that use the instance segmentation as a canvas for the heatmap.

As we know, for such a colorization problem, we need more than 4 colors (https://en.wikipedia.org/wiki/Four_color_theorem). Using less than 3 might be possible in some cases, but asserting if a map can be colored with less than 4 colors is a NP problem that we definitely don't want to deal with.

## Linked issues
Resolves #662


hermancollin:
Hm I tried implementing the colorization algorithm but it's getting unnecessarily complex for this purpose. Also, it's very much computer intensive, much more than any other part of the morphometrics process. So, what I'll do is use a semi-random method to assign the colors, just like @Stoyan-I-A suggested. It should be much more efficient.

hermancollin:
I am almost done with the implementation of this feature. This is the output of the current version as of today:
sub-nyuMouse36_sample-0001_TEM_axonmyelin_index
instance_seg
The semi-random nature of the colorization algorithm isn't ideal because some touching instances almost have the same color. Currently, the color generator saves every color internally and only outputs unique values. However, sometimes it gives very similar colors that happen to be assigned to adjacent regions.

hermancollin:
Hm I made a minor logic adjustment and it's already much better:

instance_seg

Maybe we can change the default color palette later on to have prettier visualizations, but for now this should be enough to colorize both small and large images.

hermancollin:
I'll mark this as ready for review because I'd like some feedback before writing tests and updating the documentation. @mathieuboudreau @Stoyan-I-A if you have a minute could you test the feature? It can be used with the `-c` flag when computing morphometrics.

mathieuboudreau:
> I'll mark this as ready for review because I'd like some feedback before writing tests and updating the documentation. @mathieuboudreau @Stoyan-I-A @Nish228 if you have a minute could you test the feature? It can be used with the `-c` flag when computing morphometrics on the `ac/662_instance_seg` branch.

Yup! I'll start taking a look in about a half hour or so

mathieuboudreau:
Tests are passing but a check is failing; Coveralls.io is currently having some service issues, so no worries about needing to resolve that: https://status.coveralls.io

AxonDeepSeg/visualization/colorization.py
```diff
@@ -0,0 +1,102 @@
...
+from random import randint
+from PIL import Image, ImageDraw
+
+from loguru import logger
+
+
+class color_generator(object):
+    '''Generator that yields the next color given a color palette.'''
+
+    def __init__(self, colors, mem_length=50, tolerance=30):
```
mathieuboudreau, line 17: Would we maybe want to set an optional random.seed number somewhere in the class? In case we want to rerun it on a mask that's been manually editied slightly, maybe some users would want the same colors for each axons again?

mathieuboudreau (changes requested):
Overall it works well for me ! Just missing some documentation (in RTD) and a test or two. Also see my comment about potentially adding a seed option.

Really cool feature, with the SEM test image, I could clearly see how much (!) some random axon pixels next to a large axon were erroding into the myelin, decreasing the myelin estimate of that large axon. See gif: 
gif

AxonDeepSeg/visualization/colorization.py
hermancollin: Hm yes this is an excellent idea. I'll look for a way to control the seed.

hermancollin:
> Overall it works well for me ! Just missing some documentation (in RTD) and a test or two. Also see my comment about potentially adding a seed option.

Thank you @mathieuboudreau for quickly reviewing this. Indeed, I think this will pave the way for future enhancements, because we weren't familiar with this part of the processing and this visualization basically tells us what we should work on.

AxonDeepSeg/visualization/colorization.py
hermancollin: Ok the random seed control is now implemented. I kinda want to keep the usage of this feature as it is (e.g. no value to pass when using the CLI, just adding the `-c` flag to activate colorization), but the seed control will be available for anyone using the API. I'll edit the morphometrics notebook to mention this. What do you think?

mathieuboudreau: >I kinda want to keep the usage of this feature as it is (e.g. no value to pass when using the CLI, just adding the -c flag to activate colorization), but the seed control will be available for anyone using the API. I'll edit the morphometrics notebook to mention this. What do you think?

I think this makes sense - agreed!

hermancollin:
Ok so the feature is pretty much good to go. What's left to do is
- add tests
- mention the colorization API in the morphometrics notebook

One thing I noticed is that the colorization method is pretty slow. That's the disadvantage of implementing this part ourselves instead of leveraging a well-optimized python module. The reason why I did it from scratch is because I wanted to have all the flexibility that the `color_generator` gives us for future visualization work. Downside is that an image with 8112 instances takes **a lot of time** to colorize (see example below). On my laptop, it took 31 minutes!!! Not only is it slow, but it's pretty memory-intensive (2 Gb during the colorization step and it kept on growing as the dataframe got bigger).

sub-17269_sample-17269_chunk-3_BF_instance-map

I think I know where the performance issue comes from: the data structure I'm using to store the previous 50 generated colors (so that colors are not too similar for adjacent elements) should be a FIFO queue, but I'm using a pandas DataFrame. I should change this to a better-suited data structure (the current method is O(n) which is very bad for large images with thousands of axons). Python's built-in `deque` should do the job alright but I need to refactor my code a bit.

mathieuboudreau:
>On my laptop, it took 31 minutes!!! Not only is it slow, but it's pretty memory-intensive (2 Gb during the colorization step and it kept on growing as the dataframe got bigger).

I would say this is ok as a first iteration of this feature, but we should open an issue to work on speeding this up if possible. 

Once you've got a test written, I'll do a quick do-over of your recent changes and approve!

hermancollin:
Before changing the memory data structure (on a small image with 65 instances):
```
axondeepseg_morphometrics -i . -s 0.211 -c  20,29s user 4,90s system 134% cpu 18,771 total
```
axondeepseg_morphometrics-2022-07-21T14:46:54-04:00
A little bit hard to read but the bottleneck is definitely the colorization algorithm.

---

After changing to an optimized FIFO queue:
```
axondeepseg_morphometrics -i . -s 0.211 -c  22,96s user 5,30s system 133% cpu 21,226 total
```
Actually slightly slower :sweat: 
However, I prefer this implementation (more readable). Looking at the flamegraph, I think the `color_generator` isn't really the bottleneck. PIL's `ImageDraw` functions seem responsible for the performance issues. 

Now, I'm not really sure this is the right way to do this. Also, the instance segmentation I showed in #661 is way prettier than the ones this PR currently outputs (and it was generated using ***a single matplotlib line***...) Maybe I went too deep with this one...

Thoughts?

hermancollin:
@mathieuboudreau sorry I didn't see your comment and went ahead to change the data structure anyway. Unfortunately, I think most of what I wrote will be deleted (i.e. color_generator iterator) in a future design iteration...

Stoyan-I-A:
I used this command:
`axondeepseg_morphometrics -i image.png -c`

With this image from `AxonDeepSeg/models/model_seg_rat_axon-myelin_sem/data_test`:
image_seg-axonmyelin

and obtained this result:
image_instance-map

Since we are going to change the algorithm a bit, this result is kind of irrelevant, but it does feel like something happened and the default color palette didn't create a nice image for me.

I'm attaching the logs if you ever want to take a look at it.

[axondeepseg.log](https://github.com/axondeepseg/axondeepseg/files/9176337/axondeepseg.log)

hermancollin:
@Stoyan-I-A thank you for finding this. In commit [150e01b](https://github.com/axondeepseg/axondeepseg/pull/667/commits/150e01b478e9e8ebff4f100ac1c2a69b18c72aa8), I changed the logic and it caused this bug on any image. I just fixed it, so it now works on your test image:
stoyan_test_instance-map

hermancollin:
I tried some color palettes to have a prettier output because I really didn't like the previous one.
<details><summary>green-ish</summary>

            (0, 75, 35),
            (0, 114, 0),
            (56, 176, 0),
            (204, 255, 51),

stoyan_test_instance-map

</details>

<details><summary>green-blue-ish</summary>

            (216, 243, 220),
            (149, 213, 178),
            (82, 183, 136),
            (27, 67, 50),

stoyan_test_instance-map

</details>

<details><summary>blue-ish</summary>

            (217, 237, 146),
            (118, 200, 147),
            (22, 138, 173),
            (24, 78, 119),

stoyan_test_instance-map

And I think I'm going with the last one. Time to write some tests...

mathieuboudreau:
@hermancollin I'll re-review this right now!

mathieuboudreau (approved):
The feature works as expected for me. The documentation reads clear. The tests appear sufficient, I just added a teardown step to remove a file that was generated and removes some lines that appear to have been unused.

Thanks a lot for working on this @hermancollin , I think this will be very useful! Merging now.

PR #696, opened 2022-11-09
Segment images without patches

### LINKED ISSUE #676: Leverage "no patch" option soon to be released in IVADOMED
IVADOMED currently has a new feature (though not released in a pip version yet) that allows segmenting an image without patching during this step, which could resolve issues we occasionally encounter related to cutting the images in patches (eg straight lines that cut through the image masks).

There remains questions about what we'll want the default behaviour to be, because this is a memory intensive feature (not all computers will have suffiient RAM to work on images larger than a certain size, and this will vary between users).

We'll also need to test this feature vs our current behaviour on a range of test images (eg diffferent models, different axon sizes).

mariehbourget:
## Checklist

## Description
This PR adds the option to segment without patches with the `--no-patch` flag in the CLI (as implemented in ivadomed v2.9.7).

The PR is still a draft, TODO:
- [x] Fix default overlap when using the no-patch option [here](https://github.com/axondeepseg/axondeepseg/blob/86a93308842a61175e0e75ddb8aeaae7fb2d6034/AxonDeepSeg/apply_model.py#L36)
- [x] Update documentation
- [x] Update docstrings
- [x] Add test

## Linked issues
Fixes #676

### DIFF
--- a/AxonDeepSeg/apply_model.py
+++ b/AxonDeepSeg/apply_model.py
@@ -21,9 +21,11 @@ def axon_segmentation(
     :param path_acquisitions_folders: the directory containing the images to segment.
     :param acquisitions_filenames: filenames of the images to segment.
     :param path_model_folder: path to the folder of the IVADOMED-trained model.
-    :param overlap_value: the number of pixels to be used for overlap when doing prediction. Higher value means less
-    border effects but more time to perform the segmentation.
+    :param overlap_value: the number of pixels to be used for overlap between patches when doing prediction.
+    Higher value means less border effects but more time to perform the segmentation.
     :param acquired_resolution: isotropic pixel size of the acquired images.
+    :param no_patch: If True, the image is segmented without using patches. Default: False. This parameter supersedes
+    the "overlap_value" parameter. It may not be suitable with large images depending on computer RAM capacity.
     :param verbosity_level: Level of verbosity. The higher, the more information is given about the segmentation
     process.
     :return: Nothing.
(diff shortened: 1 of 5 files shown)


mariehbourget:
This PR is now ready for review.
The new "no-patch" feature only applies to CLI and not to fsleyes where the behavior is unchanged,
@Stoyan-I-A, do you think we should integrate it in fsleyes or wait to do it in Napari?

Stoyan-I-A:
It should be pretty easy to add this option in FSLeyes, so I'd say we should do it.
Now, should we do it in this PR or in another one?

mariehbourget:
> It should be pretty easy to add this option in FSLeyes, so I'd say we should do it.
> Now, should we do it in this PR or in another one?

I would say probably separately because there is a small catch:
the "no-patch" option is supposed to be used without the "overlap" option. If both are used on the CLI, currently the "no-patch" option supersedes the "overlap" option and a warning is shown in the terminal from ivadomed.
So to make it clean in fsleyes, we would need to somehow disable the "overlap" option in the GUI when using the "no-patch" option. I don't know enough about the plugin to estimate if that is complicated or not.

Stoyan-I-A:
> I would say probably separately

Ok, let's do it this way.

> we would need to somehow disable the "overlap" option in the GUI when using the "no-patch" option

For the user, I'm thinking of greying out (disabling) the "overlap" box in the settings menu when a no-patch checkbox is selected.

mathieuboudreau:
Tests pass successfully on my local machine. I had to not only update my conda env (expected) but also re-do `pip install -e .` (thought for some reason that I wouldn't have to, but no-patch option wasn't working until I did so).

mathieuboudreau:
Tested functionality with and without no patch option. I feel like just from this first test, the no-patch option results in better segmentation quality for several axons, in particular it fixes "irregular artifacts" in myelin.

## Without patching

image_seg-axonmyelin


## Without patching

image_seg-axonmyelin 2

## GIF


out

mathieuboudreau:
Documentation looks good

mathieuboudreau:
Thanks for adding a test [🎉](https://emojipedia.org/party-popper/)

mathieuboudreau (approved):
Looks great to me as is - thanks a lot @mariehbourget !

PR #707, opened 2023-01-03
Change pixel size to reduce RAM usage during no-patch test

### LINKED ISSUE #706: Scheduled CI fails irregularly on Windows (no-patch test)
Over the holiday break I've received irregular emails stating that the CI has failed.

Here is the latest example: https://github.com/axondeepseg/axondeepseg/actions/runs/3826925769/jobs/6511100185#step:8:275

It occurs during the Windows matrix, for the test test/test_segment.py::TestCore::test_main_cli_runs_succesfully_with_nopatch_flag, which was merged from PR #696 

It doesn't occur consistently, i.e. I've rerun the tests and they sometimes pass.

The error is a segfault.

We've had this issue with Windows tests failing "randomly" before, they can be difficult to debug (maybe it's GitHub machine dependent?).

Nonetheless, having failing scheduled CI tests is not favourable, because we'll start ignoring them and the whole purpose of the scheduled CI tests are for us to be notified when something unexpected occurs with our code/dependencies on master, and so we might miss these.

@mariehbourget @kousu thoughts?

mathieuboudreau:
## Checklist

## Description

Because the Windows and Linux machines that GitHub action uses only have 7GB of ram, occasionally the CI will fail the no-patch tests that were recently added when the memory runs out (it uses approx 6GB). This PR reduces the image sizes via the pixel size for these tests, so that they (hopefully) won't fail anymore during the nightly scheduled CI runs. In addition, this will hopefully prevent some users (not happened yet) from getting fails from running the test suite after installation, also due to insufficient RAM.

## Linked issues

Resolves #706

### DIFF
--- a/test/test_segment.py
+++ b/test/test_segment.py
@@ -401,7 +401,7 @@ class TestCore(object):
     def test_main_cli_runs_succesfully_with_nopatch_and_overlap_flags(self):
 
         with pytest.raises(SystemExit) as pytest_wrapped_e:
-            # Note that the pixel size set here, 0.1, differs from the true size, 0.37, in order to reduce RAM burden GitHub Actions CIs and users computers
+            # Note that the pixel size set here, 0.1, differs from the true size, 0.37, in order to reduce RAM burden on GitHub Actions CIs and users computers
             AxonDeepSeg.segment.main(["-t", "SEM", "-i", str(self.imagePath), "-v", "1", "-s", "0.1", "--no-patch", "--overlap", "48"])
 
         assert (pytest_wrapped_e.type == SystemExit) and (pytest_wrapped_e.value.code == 0)
--- a/test/test_segment.py
+++ b/test/test_segment.py
@@ -392,7 +392,8 @@ class TestCore(object):
     def test_main_cli_runs_succesfully_with_nopatch_flag(self):
 
         with pytest.raises(SystemExit) as pytest_wrapped_e:
-            AxonDeepSeg.segment.main(["-t", "SEM", "-i", str(self.imagePath), "-v", "1", "-s", "0.37", "--no-patch"])
+            # Note that the pixel size set here, 0.1, differs from the true size, 0.37, in order to reduce RAM burden GitHub Actions CIs and users computers
+            AxonDeepSeg.segment.main(["-t", "SEM", "-i", str(self.imagePath), "-v", "1", "-s", "0.1", "--no-patch"])
 
         assert (pytest_wrapped_e.type == SystemExit) and (pytest_wrapped_e.value.code == 0)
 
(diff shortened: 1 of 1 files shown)


test/test_segment.py
```diff
@@ -392,15 +392,17 @@ def test_main_cli_zoom_factor_sweep_throws_error_folder_with_multiple_images(sel
     def test_main_cli_runs_succesfully_with_nopatch_flag(self):
 
         with pytest.raises(SystemExit) as pytest_wrapped_e:
-            AxonDeepSeg.segment.main(["-t", "SEM", "-i", str(self.imagePath), "-v", "1", "-s", "0.37", "--no-patch"])
+            # Note that the pixel size set here, 0.1, differs from the true size, 0.37, in order to reduce RAM burden GitHub Actions CIs and users computers
```
mariehbourget, line 395:
```suggestion
            # Note that the pixel size set here, 0.1, differs from the true size, 0.37, in order to reduce RAM burden on GitHub Actions CIs and users computers
```

test/test_segment.py
```diff
@@ -392,15 +392,17 @@ def test_main_cli_zoom_factor_sweep_throws_error_folder_with_multiple_images(sel
...
+            AxonDeepSeg.segment.main(["-t", "SEM", "-i", str(self.imagePath), "-v", "1", "-s", "0.1", "--no-patch"])
 
         assert (pytest_wrapped_e.type == SystemExit) and (pytest_wrapped_e.value.code == 0)
 
     @pytest.mark.integration
     def test_main_cli_runs_succesfully_with_nopatch_and_overlap_flags(self):
 
         with pytest.raises(SystemExit) as pytest_wrapped_e:
-            AxonDeepSeg.segment.main(["-t", "SEM", "-i", str(self.imagePath), "-v", "1", "-s", "0.37", "--no-patch", "--overlap", "48"])
+            # Note that the pixel size set here, 0.1, differs from the true size, 0.37, in order to reduce RAM burden GitHub Actions CIs and users computers
```
mariehbourget, line 404:
```suggestion
            # Note that the pixel size set here, 0.1, differs from the true size, 0.37, in order to reduce RAM burden on GitHub Actions CIs and users computers
```

mariehbourget (dismissed):
LGTM, thanks @mathieuboudreau!

For reference, the max RAM used is now :
- about 4 GB for all tests in `test_segment.py`.
- about 0.9 GB only for the 2 `no-patch` tests (instead of 6 GB).

mathieuboudreau:
Great thanks @mariehbourget ! I've committed your suggested change, plz re-approve so that I can merge after the CI tests are done.

mariehbourget (approved):
Thanks @mathieuboudreau!

There was 2 (identical) suggestions to commit after my first review. You committed the first one, so I took the liberty of committing the second one before approving the PR.

PR #728, opened 2023-03-09
[Patch] changed colormap used in the FSLeyes plugin

Stoyan-I-A:
## Checklist

## Description
The aim of this change is to try to fix an error that happened to a user. I am not 100% sure it will resolve it because I can't reproduce the problem on my end.
This change also seems to have the side effect that we can't change the opacity of the overlay. And if we try to change it in one of the menu settings of FSLeyes it will throw an error. 

To be clear, I don't think this is a good change and I would rather not merge it but it may just solve an issue for a user.

## Linked issues
Resolves #727


Stoyan-I-A:
I can probably think of something better, but it is getting late right now. I can continue investigating in the upcoming days.

PR #669, opened 2022-07-26
Improve and force imread/imwrite conversion to 8bit int

mathieuboudreau:
## Checklist

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://www.neuro.polymtl.ca/software/contributing#pr_labels) to this PR
- [x] I've added relevant tests for my contribution
- [ ] I've updated the documentation and/or added correct docstrings
- [x] I've assigned a reviewer
- [x] I've consulted [ADS's internal developer documentation](https://github.com/neuropoly/axondeepseg/wiki/Guidelines-for-contribution) to ensure my contribution is in line with any relevant design decisions

## Description

Changes how bit 8-bit depth conversion is done in ADS's imread, removes the optional bit depth arg for this function (it appeared to have been unused for a long time, always using the default value of 8), and add test verifying that the same image saved at different int and float precision loaded with ads.imread outputs the same 8bit image array.

## Linked issues

Resolves #175

PR #742, opened 2023-05-23
Save raw instance segmentation

### LINKED ISSUE #668: Save 16-bit instance segmentation image
**Describe the problem**
In #667, a first iteration of the instance segmentation output feature was completed. To make this change even more useful for users, we would also like to save the instance segmentation as is, without any colorization, so that they can use it efficiently in a pipeline.

**Proposed solution**
A 16-bit image should do the job for any real-case scenario.

### LINKED ISSUE #697: Add 16 bit image compatibility for colorization of more than 256 axons
See the discussion in #669 for more details, specifically this comment: https://github.com/axondeepseg/axondeepseg/pull/669#issuecomment-1202690091

hermancollin:
## Checklist

## Description

This PR adds a feature discussed last year in #668 after the instance colorization feature was merged. When using the `-c` flag, the raw instance map is now also saved alongside the colorized image.

Note 2 things: 
- I changed the colorized image suffix from `_instance-map.png` to `_colorized.png` as the former was more appropriate for the raw instance map file.
- I added back the *bitdepth* argument in imread which was removed in #669 (I don't recall why). For now it's useless because the raw instance map (saved in 16bit PNG) is never read anywhere. However, if we want to read a 16-bit instance map, we will be able to use our imread function with `bitdepth=16`.
https://github.com/axondeepseg/axondeepseg/blob/7e042332e7bbf296c0d25ee9cabce377cc6e4275/AxonDeepSeg/ads_utils.py#L258

## Linked issues

Resolves #668
Resolves #697

### DIFF
--- a/AxonDeepSeg/ads_utils.py
+++ b/AxonDeepSeg/ads_utils.py
@@ -255,7 +255,7 @@ def convert_path(object_path):
         else:
             raise TypeError('Paths, folder names, and filenames must be either strings or pathlib.Path objects. object_path was type: ' + str(type(object_path)))
 
-def imread(filename):
+def imread(filename, bitdepth=8):
     """ Read image and convert it to desired bitdepth without truncation.
     """
 
@@ -280,7 +280,7 @@ def imread(filename):
         raw_img = imageio.v2.imread(filename)
         if len(raw_img.shape) > 2:
             raw_img = imageio.v2.imread(filename, as_gray=True)
-    img = imageio.core.image_as_uint(raw_img, bitdepth=8)
+    img = imageio.core.image_as_uint(raw_img, bitdepth=bitdepth)
     return img
 
 def imwrite(filename, img, format='png'):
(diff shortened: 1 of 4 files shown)


hermancollin:
By using this feature I found a bug. When the number of axons in the image is inferior to 256, even though we force the datatype to `np.uint16`, the image is saved as an 8-bit PNG. In that case, I'm not sure if we should try to force the conversion to 16 bits.

On one hand, we would have the guarantee that the instance map is correctly handled by `ads_util.imread()` if `bitdepth=16` is used (predictable interface for the myelin maps). On the other hand, it could be highly inefficient memory-wise.
If we keep the flexibility of using 8b or 16b depending on the number of axons in the image, we could make sure that myelin maps are correctly loaded by catching the myelin map suffix in the filename and using a try-except block to catch conversions from uint16 to uint8 and vice-versa.

mathieuboudreau:
@hermancollin this PR is still a draft, but is 1+ year old. Is this something we want to revisit?

PR #756, opened 2023-09-18
Remove conda-forge channel for speed and refactor environment 

mathieuboudreau:
## Checklist

## Description

Scheduled tests started to fail. After some digging, I believe that it was due to some conda conflict resolution issue that was using up more memory than GitHub allocates. To resolve this, I've removed the conda-forge channel per this suggestion (https://github.com/conda/conda/issues/5003#issuecomment-292006267), and moved two of the conda dependencies to pip because they were only avalailable with conda through conda-forge. These changes substantially sped up the CI env installation, and reduced memory usage resulting in the tests now completing.


mathieuboudreau:
The first dependency to trigger the memory error in the Linux/Windows Github Actions tests is imageio - https://github.com/axondeepseg/axondeepseg/actions/runs/6226857304/job/16900267199?pr=756

I'm going to remove it, and add the remaining ones to see if this is a singular problematic dependency, or if still happens when there are too many / others

mathieuboudreau:
The same error happens after I remove imageio and add pytest instead, indicating the problem may not be specific to any package but just the sum of all of them (maybe there's a a subdependency that was updated that's causing havoc in the conflict resolution step).

Going to try out this https://github.com/runforesight/workflow-telemetry-action and also make conda install the env with more verbosity

Stoyan-I-A (approved):
Looks good to me

PR #767, opened 2023-10-27
Add A. Wheeler testimonial

mathieuboudreau:
Just a small change in the documentation to add one testimonial, so I'll force merge to main after the build passes and have looked at the RTD page to make sure it formatted correctly.


mathieuboudreau:
Huh - seems like GitHub just recently removed the options for admins to force merge PRs without an approving review. I'll assign some ppl then

jcohenadad (approved):

Issue #640, opened 2022-05-24
Add quality control


hermancollin:
I've been thinking about this for a while. It would be nice to have a minimal app to quickly do quality control on single segmented image or more importantly for lengthy folders. I remember using SCT's [quality control script](https://spinalcordtoolbox.com/user_section/command-line.html#sct-qc) a year ago, which implemented a browser-based solution if I recall correctly. This is ideal from a cross-platform point of view.

This also ties in with previous discussions where we concluded that it could be interesting to offer an alternative to `fsleyes` as a GUI, so if it works well, it could be used for a graphical interface to ADS as well.

My suggestion is to implement a browser-based solution. The user would call a `axondeepseg_qc` script which would start a server, then prompt to open a localhost:portno in the browser. 
I've been researching this a bit and I think web-assembly (WASM) would be a great candidate. WASM was conceived to complete javascript's functionalities in an attempt to bring high-performance applications to the web. The concept is that you write your code in a low-level language (like C/C++ or, more interestingly, in Rust) and chose wasm as a compilation target. The browser acts as a level of abstraction over the operating system. Nowadays, wasm is supported by practically all browsers. The fact that it allows for high-performance web apps is what makes me think it could also work as an alternative for ADS (also, browsers can use GPUs).

I will do some testing on my side to check how easy it would be to implement and interface with python code. One question I have is this: for maintainability, would it cause a problem to use another language than python for this? I will say I have my eyes on Rust (pretty much the fastest growing language out there right now), but maybe C++ would be preferable for students in the future. For Rust, the `yew` project (https://yew.rs/) grabbed my attention. You write your application in Rust and then compile it for WASM. I'm not exactly sure but I think we could essentially release a wasm binary executable, then call it through a python script like the other scripts. This executable would work across every OS since its executed by the browser.

jcohenadad:
related to https://github.com/axondeepseg/axondeepseg/issues/404

hermancollin:
I tried `yew.rs` and its pretty good but it's not production ready. We'll see how it goes.

Obviously, for QC alone (not ADS GUI), the simpler way would be a script producing a html file, just like SCT: https://spinalcordtoolbox.com/overview/concepts/inspecting-results-qc-fsleyes.html. We could have a new `axondeepseg_qc`  script to generate an interactive html file from the current directory. It makes sense to me to have a separate interface for these kinds of "reports", although it would be nice to have it consolidated with the ADS GUI. Also, much simpler to implement.

[...]

PR #781, opened 2024-01-12
Morphmetrics QA - export histogram & stats

mathieuboudreau:
## Checklist

## Linked issues

#779


mathieuboudreau:
Create an initial QA class for metrics, it saves histograms with some statistics next to them. Did an initial integration into Napari using the pre-existing Morphmetrics button (all metrics histograms are saved in a folder called QA/ where the morphometrics file is saved), but would likely be best to add it's own button. 

Current status: Feature works, code is minimally documented, no tests yet, some hanging files (eg notebook being used during development).

Here's some examples of outputs using the sample SEM dataset:

gratio
axon_diam (um)

myelin_thickness (um)
solidity

mathieuboudreau:
One additional output I'd like to include are some suspected wrongly segmented axon IDs / locations, eg gratio = 0 or 1, axon_diam = 0, solidity = 1, etc. Maybe even a user-input

mathieuboudreau:
Adding on to this feature, I just made a proof-of-concept implementation of a feature that takes some pre-determined metric conditions to flag potential image objects that have issus (eg g-ratio = 1), and then combines with the instance segmentation to create masks of image objects that should be QA-d, 

eg,
Screenshot 2024-01-14 at 2 14 12 PM

Zoomed in,


Screenshot 2024-01-14 at 2 14 18 PM

Which the user can then decide what to do in to fix the image, for example here by looking at just the axon masks, we see that some spurious axon pixels were hidden behind the myelin values (something we should fix in another issue!)
Screenshot 2024-01-14 at 2 14 44 PM

mathieuboudreau:
From our 2025-01-17 meeting, here's @hermancollin's example QA workflow in a colab notebook: https://colab.research.google.com/drive/1CPbA1wMrRrpgC5hc-RbcikHs6jQSGvLv?usp=sharing

AxonDeepSeg/qa/metrics_qa.py
```diff
@@ -0,0 +1,126 @@
...
+        flagged_objects = np.unique(flagged_objects)
+
+        mask = np.zeros_like(morph_out[2])
+
+        # go through each element in arr
+        for id in flagged_objects:
+            locations = np.where(np.isclose(morph_out[2], id+1))
+            mask[locations] = 1
+
+        ads.imwrite(Path(save_folder) / 'flagged_objjects.png', mask)
```
hermancollin, line 126:
```suggestion
        ads.imwrite(Path(save_folder) / 'flagged_objects.png', mask)
```

hermancollin:
So far this feature looks very neat! Especially this "auto-flagging" function. Now #864 and even #640 will be easier to implement. I'm thinking we can input a list of axon IDs and the histograms can be updated.

However, what was the rational behind the rules for the `get_flagged_objects` method? I don't understand the last 2 with axon and myelin areas.

https://github.com/axondeepseg/axondeepseg/blob/1a41aebfddcecb3d1c855ded7b1a63a311090800/AxonDeepSeg/qa/metrics_qa.py#L112-L115

Eventually, we could make the user choose his own rules

mathieuboudreau:
> So far this feature looks very neat! Especially this "auto-flagging" function. Now #864 and even #640 will be easier to implement. I'm thinking we can input a list of axon IDs and the histograms can be updated.
> 
> However, what was the rational behind the rules for the `get_flagged_objects` method? I don't understand the last 2 with axon and myelin areas.
> 
> https://github.com/axondeepseg/axondeepseg/blob/1a41aebfddcecb3d1c855ded7b1a63a311090800/AxonDeepSeg/qa/metrics_qa.py#L112-L115
> 
> Eventually, we could make the user choose his own rules


It was more for development purposes / see what good default values might be, but yes they should be controllable!

mathieuboudreau:
Maybe @hermancollin now would be a good time for you to test out the feature. Here's the instructions (I havent made documentation changes yet).

1. Open napari
2. Open the plugin
3. Open the image
4. Load or generate the segmentation
5. Generate a morphometrics file
6. Click the QA Report button, and wait until the HTML gets loaded into a website (I'll probably need a message to the user, as this can take a minute if there are a lot of axons since there are PNGs made for each axons)

hermancollin:
> So, that was kind of just some proof of concepts I think. I was trying to maybe find some way to flag to users some suspect axons that they could then check and manually correct; at the time, I think I was flagging g-ratios >0.99, and the bottom 5% axon diameter and myelin thickness axons. That was working with the test image at the time, but when I ran it on this most recent test image, a lot of axons that were completely fine got flagged, so I think if I were to to this again, maybe we would need a smarter algorithm.

Hmm I feel like the g-ratios >0.99 is pretty safe to flag as a bad seg. Bottom 5% axon diam was most likely the problem because if all axons are well segmented, it doesn't make sense to remove the smaller ones.

> I think a better solution now that I have the single axon tab would be, instead of browsing through them in order of axon index (so top right to top left, top from bottom), adding a dropdown menu to change how they are sorted, so that maybe a user can just look at the smallest axons in ascending order and flag those accordingly. It should be fairly easy to implement, but maybe it's could be done in a subsequent PR in order to move this one ahead quicker.

Agreed, I'll open an issue for this. This sounds like a really good solution and elegant solution to this problem.


> I haven't made the QA report available through CLI, that would need some refactoring to make it possible since I was using the napari GUI object properties to gain access to a lot of needed data (image/seg data, filenames, morphometrics path, etc)

Oh that's a shame! I feel this feature would be really cool to review batches of images, but I don't know how to accomodate this in the GUI.

hermancollin:
> So, I havent tested out if there's a way to dynamically load the main image and then display close ups, that could be another issue to resolve the one you just mentionned. I think it technically won't run any slower if the image is very large my way; actually the opposite, if I force the browser to load a 1000 mb image into the memory and then manipulate that to display the closeups, I think there's a higher likelyhood that it will crash the browser / slow down

Ah but in your QA you are also loading the full image (in the summary tab) so the 1000 Mb image still needs to be loaded.

I think I can grasp how impractical javascript is for this, from my limited experience.

hermancollin:
In terms of UE this feature is really neat and super user friendly. I really like the workflow. I can definitely imagine recommending this to users, and they would very quickly understand how this works.

Like you mentioned earlier, it would be great to display a message when the `QA Report` button is pressed to give some feedback to the user (i.e. please wait this might take a minute).

mathieuboudreau:
> > So, I havent tested out if there's a way to dynamically load the main image and then display close ups, that could be another issue to resolve the one you just mentionned. I think it technically won't run any slower if the image is very large my way; actually the opposite, if I force the browser to load a 1000 mb image into the memory and then manipulate that to display the closeups, I think there's a higher likelyhood that it will crash the browser / slow down
> 
> Ah but in your QA you are also loading the full image (in the summary tab) so the 1000 Mb image still needs to be loaded.
> 
> I think I can grasp how impractical javascript is for this, from my limited experience.

So, the QA HTML is intended to be run on your local computer. I think in the case of the large image, it's only meant to be displayed, and it points to a local image, so I don't think the browser adds it to the memory (but maybe). For the other case I was talking about, javascript would have to load each pixel value and assign that to an array which will likely go to the RAM, and then further manipulate this memory to create the cropped versions, so I think the memory usage would definitely be more.

But yes, we should definitely stress test this and inform the users accordingly

mathieuboudreau:
@hermancollin I added documentation, but not figures. Once the documentation looks good, then I'll have a better idea which figures to include and where. Feel free to suggest the ones you feel are needed to better understand the text through comments in github, and I'll add those

mathieuboudreau:
Hmm my fill holes fix impacted the tests on some expected data assert tests ; I'm exploring possible ways of getting around it since we don't want to change the behavious for prior images, just in the context of the QA assurance for now

hermancollin:
ok I'll take a look in a minute.

btw I have a project which will be perfect to try this feature in a real scenario. The images will be fairly large, so it will also be a stress test.

I talked with the stanford group recently. If this feature ends up being really useful, it might be of interest to extend to unmyelinated axons.

PR #788, opened 2024-02-16
Add reinstallation documentation

mathieuboudreau:
## Checklist

## Description

Our old update/reinstall instructions would occasionally lead to users having broken installs (see: https://github.com/axondeepseg/axondeepseg/discussions/784), and so asking the users (note: not devs) to simply delete the repo and conda env and start over is likely the cleanest instructions we can provide.

I've also updated the Training section with two notices: one that IVADOMED is being discontinued, and another informing people that ADS is able to train models with their data (with conditions)


mathieuboudreau:
Link to documentation that was added: https://axondeepseg--788.org.readthedocs.build/en/788/documentation.html#updating-axondeepseg

docs/source/documentation.rst
```diff
@@ -183,6 +173,28 @@ Below is a short tutorial describing how to use the AxonDeepSeg plugin for Napar
...
+
+To uninstall a conda environment:::
+
+    conda env remove -n ads_venv
+
+In addition, we recommend that user (developers: see note below) delete the axondeepseg/ folder and redownload it as described in our installation instructions to ensure that your software will be the latest version.
+
+From the parent folder of the AxonDeepSeg directory:::
+
+    rm -rf axondeepseg
```
Stoyan-I-A, line 191: For Windows users, the command will be `rmdir axondeepseg`

mathieuboudreau: Is that on the powershell? Would the command be the same on the Anaconda terminal?

Stoyan-I-A: > Would the command be the same on the Anaconda terminal?

Yes

docs/source/documentation.rst
```diff
@@ -183,6 +173,28 @@ Below is a short tutorial describing how to use the AxonDeepSeg plugin for Napar
...
```
Stoyan-I-A, line 191: Here's the instruction box I suggest using

```
    # Mac and Linux
    rm -rf axondeepseg

    # Windows
    rmdir axondeepseg /s
```

Stoyan-I-A (dismissed):

mathieuboudreau:
I think the update/reinstall instructions in this PR are stale, we've moved to an SCT-style installation.

I \*think\* now the user would just need to update the git repo and call `./install_ads` again, @joshuacwnewton can you confirm?

joshuacwnewton:
> I *think* now the user would just need to update the git repo and call `./install_ads` again, @joshuacwnewton can you confirm?

It depends what needs to be updated?

- If only the ADS code needs to be updated, then updating the git repo is sufficient. (This is because ADS gets installed in "editable mode", so it will adopt any updates to the code automatically.)
- If only a specific package needs to be updated, then it's possible to run `pip install -U {package_name}` (with the venv activated) to just upgrade that single package.
- To reinstall all of the dependencies from scratch, you can run this short-cut command: `pip freeze --exclude-editable | xargs pip uninstall -y && pip install -r requirements.txt`. I keep an alias for this!
- If you want to be safe and update everything (including the venv itself, any data/model files, etc.) then indeed, the safest option is to run `./install_ads`.

I think I would consider the first three options to be "dev only", and the last option to be the safest for all users, since it will result in an installation that is guaranteed to match the setup used in automated testing. 

But, the other options are still useful when developing locally to avoid having to recreate everything.

docs/source/documentation.rst
mathieuboudreau: Stale - entire installation instructions have changed, so this comment isn't relevant

docs/source/documentation.rst
mathieuboudreau: Stale

docs/source/documentation.rst
```diff
@@ -74,6 +74,15 @@ Then, to use ADS's command-line scripts in Command Prompt, please follow these i
 4. Click 'New', then copy and paste the (ADS_PATH)\bin directory:
 5. Click 'OK' three times. You can now access ADS's scripts in the Command Prompt.
 
+Updating AxonDeepSeg
+--------------------
+
+To update the AxonDeepSeg to the latest version, run::
+
+    git checkout master && git fetch && git pull
```
joshuacwnewton, line 82: Technically `git fetch && git pull` is redundant, since `git pull` already contains `git fetch`?

joshuacwnewton (approved):
This is a straightforward fixup of this old PR! LGTM :)

docs/source/documentation.rst
mathieuboudreau: True! I think it's a habit I've kept too long from when I first was taught git haha. Git pull only it is

PR #792, opened 2024-03-11
Check IVADOMED PR #1308 compatibility

mathieuboudreau:
My MBP has issues resolving the ADS environment file today, so I'm resorting to GitHub acctions to check this PR change to the dependencies. Will close after IVADOMED PR is resolved

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -26,5 +26,5 @@ dependencies:
     - raven
     - napari
     - protobuf==3.20.1
-    - ivadomed @ git+https://github.com/ivadomed/ivadomed.git@c52689d087de28f9a9a814b6eb99bf2093d96c5a
+    - ivadomed @ git+https://github.com/ivadomed/ivadomed.git@8f87bca335ea3de60bac98224e13ea6a3e1b91a6
     - onnxruntime<1.16.0


mathieuboudreau:
nibabel pinned to 3.2.* on our end is in conflict with ivadomed, but the only reason we had it pinned to that version was to be on the same wavelength as IVADOMED (https://github.com/axondeepseg/axondeepseg/pull/655), so it's fair to update it if they're doing it too

Stoyan-I-A:
Is this still a draft?

mathieuboudreau:
> Is this still a draft?

Yes, I’m going convert it to point to the newest release of ivadomed now that it was made today!

hermancollin:
@mathieuboudreau what is the status of this PR? I think we might want to merge this before the change to nnunet to make sure the last commit using ivadomed still works.

Note that I am removing both ivadomed and nibabel from the dependencies.

mathieuboudreau:
@hermancollin I think we just need your approval to merge this PR - the tests all passed - although, this was pre-numpy 2.x, so I'm going to rerun them just to make sure

mathieuboudreau:
Ha glad I rechecked, will look into this

mathieuboudreau:
Hum, looks like we had this issue recently in this PR: https://github.com/axondeepseg/axondeepseg/pull/777

For that one, it seems to have been related to memory consumption, but here the tests pass so it doesn't feel related.

mathieuboudreau:
Seems like the issues stems from using two different Coverals services; the pip coveralls for each parallel test running, and then the Github Actions coveralls service to combine them all and send the final report. https://github.com/lemurheavy/coveralls-public/issues/1665#issuecomment-1322763874

The fix should be to just use the GitHub actions coveralls service for both steps, but it's not clear right away to me on how to do this using their example - I'll look into it: https://github.com/marketplace/actions/coveralls-github-action

hermancollin (approved):

mathieuboudreau:
The coveralls issue can be handled in it's own PR, I'll force approve this one

PR #816, opened 2024-07-16
Update sphinx to fix doc build

hermancollin:
RTD fails on all PRs at the moment. It complains that sphinx should be at least v5.0.0


jcohenadad (approved):
Fix works! Thank you @hermancollin 😊

PR #800, opened 2024-04-23
`ivadomed` to `nnunetv2` backend migration

hermancollin:
## Checklist
- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://intranet.neuro.polymtl.ca/geek-tips/contributing#pr-labels-a-href-pr-labels-id-pr-labels-a) to this PR
- [x] I've added relevant tests for my contribution
- [x] I've updated the documentation and/or added correct docstrings
- [x] I've assigned a reviewer
- [x] I've consulted [ADS's internal developer documentation](https://github.com/neuropoly/axondeepseg/wiki/Guidelines-for-contribution) to ensure my contribution is in line with any relevant design decisions

## Description
This PR replaces the `ivadomed` dependency with `nnunetv2`. The old models will unfortunately no longer be supported directly, but they can always be used by reverting to an earlier commit. The main changes are
- addition of model cards to list all available models with download links, intended use, training data details
- refactoring `download_models.py` where URLs to the 3 standard ivadomed models were hardcoded; this script now only downloads one model at a time (generalist model by default)
- refactoring of `segment.py` where some ivadomed-related options are no longer available. Input resampling is also applied on our side now (it was previously done in ivadomed itself based on the input px size VS training px size)
- refactoring of `apply_model.py`; here, we now use the nnunet inference function instead of `ivadomed.inference.imed_inference()`

## To-do
* [x] Refactor tests - @mathieuboudreau 
* [x] Refactor documentation

## Linked issues
Resolves #765 (master issue)
Addresses #797 (download_model refactoring)
Addresses #798 (segmentation CLI refactoring)
Addresses #799 (model cards)

PR #821, opened 2024-08-19
Add `install_ads` scripts to automate conda env creation and dependency installation

### LINKED ISSUE #820: Change installation procedure to SCT/ST-style shell script install
@joshuacwnewton and I had several discussions, and think ADS would greatly benefit from the type of installation that both SCT and ShimmingToolbox uses. This would go one step towards consolidating how we install all the major NeuroPoly softwares that are python-base.

In addition, Joshua had a few more suggestions (eg. using pip-only in the conda install, freezing versions at releases, etc).

@joshuacwnewton indicated that he has the bandwidth to start working on this, and has already setup a proof of concept:

https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4598
https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/4598/files
https://github.com/spinalcordtoolbox/spinalcordtoolbox/actions/runs/10295156783/job/28494386192?pr=4598

I indicated that I'll follow along so if at any point he doesn't have the time to work on it, I'll pick it up.

Ideally, we'll add this to the v5.0 release of ADS, along with PR #800 but in it's own PR.

It will however be developed using the dependencies in PR #800 and not main, so if it's done before PR #800 is merge, we can merge it directly into that branch instead of master.

@joshuacwnewton I'd suggest you make a branch and then a draft PR so that 1) the CI can run while changes are pushed, and 2) so that I can get alerts when you push commits. We can have the technical discussions there instead of this issue.

joshuacwnewton:
## Description

This PR copies the SCT installation scripts, then modifies them so that they work for ADS. This is possible because SCT and ADS have [very similar installation steps](https://gist.github.com/joshuacwnewton/3522221596d79a2affa78a76e53153f8).

> [!IMPORTANT]
> The SCT installation scripts contain a lot of little tweaks and fixes that have accumulated over the years (even before I joined NeuroPoly). Given the age of the script, some of these fixes may be outdated/unnecessary. 
>
> However, the reason I decided to copy the SCT installation scripts (as opposed to starting from scratch) is because I figured that we could interrogate the tweaks/fixes together, then propagate any updates we make back to the SCT repository. So, if there are any lines of code in `install_ads` that look suspicious/unnecessary, please feel free to point them out and I will be happy to research and/or explain them. (Thank you!)

> [!NOTE]
> This branch is built off of the `ac/nnunet_migration` branch. Once that branch is merged, the base branch for this PR will be automatically updated by GitHub to `master` instead. So, feel free to prioritize that PR before reviewing this one, since the CI runs will only get triggered if this PR is targeting `master`.

## Linked issues

Fixes #820.


joshuacwnewton:
I've changed the base branch to `master` temporarily just to be able to trigger the CI workflows. (This is why you see all of these extra commits from `ac/nnunet_migration`.)

As soon as I get the tests passing (or as soon as `ac/nnunet_migration` is merged), I can fixup the base branch to remove all of the duplicated commits.

AxonDeepSeg/__init__.py
```diff
@@ -1 +1,7 @@
+import os
+import sys
+
+# make sure repo directory (two directories up) is on path (to access config.py)
+sys.path.append(os.path.abspath(os.path.join(__file__, os.pardir, os.pardir)))
```
joshuacwnewton, line 5: Note: We try to access a file outside of the `AxonDeepSeg` package directory (`config.py`), but there is no root-level `__init__.py`. 

On Windows, the root of the repo doesn't get added to the PATH, yet on Linux/macOS it does. This caused test failures, hence me adding this line.

But, I'm surprised! I wonder why these errors didn't occur with the old installation method.

install_ads
```diff
@@ -0,0 +1,712 @@
...
+    # Switch between shell
+    if [[ "$THE_RC" == "bash" ]]; then
+      echo "export ADS_DIR=$ADS_DIR"
+    elif [[ "$THE_RC" == "csh" ]]; then
+      echo "setenv ADS_DIR $ADS_DIR"
+    fi
+    # add line
+    echo ""
+  ) >> "$RC_FILE_PATH"
+}
```
joshuacwnewton, line 299: Note that one of the bigger changes in this script is how the CLI scripts are accessed:

- Before: Necessary to activate the conda environment
- After: Specific launchers (`axondeepseg_*`, `download_*`) are copied to a separate `bin/` folder, which is then made available via the `.{bash,zsh}rc` file in the user's home directory.

There are some caveats with this approach, namely the pollution of said `.rc` file with repeated installs + it is tricky to have multiple concurrent versions (as the most recently installed version supersedes the other).

I'd love to chat about the caveats of SCT's installation procedure, as it could retroactively help SCT as well. :)

test/morphometrics/test_compute_morphometrics.py
```diff
@@ -744,7 +744,7 @@ def test_morphometrics_consistency(self):
         for column in reference_stats_dataframe:
             column_ref_vals = reference_stats_dataframe[column].to_numpy()
             column_new_vals = new_stats_dataframe[column].to_numpy()
-            assert np.allclose(column_ref_vals, column_new_vals, rtol=0, atol=1e-11, equal_nan=True)
+            assert np.allclose(np.abs(column_ref_vals), np.abs(column_new_vals), rtol=0, atol=1e-11, equal_nan=True)
```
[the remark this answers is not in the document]
mathieuboudreau: Do you recall which metrics?

.github/workflows/run_tests.yaml
```diff
@@ -12,11 +12,12 @@ on:
   push:
     branches:
       - master
+      - ac/nnunet_migration
 
   # Trigger current workflow on PR from ANY branch
   pull_request:
     branches:
-      - '*'
+      - '**'
```
joshuacwnewton, line 20: I had been wondering about this! (Thank you for finding the right syntax to get the run to trigger. I'll need to add this to SCT, too.)

mathieuboudreau:
Attempting to open napari on my mac, I get this error:

```
(venv_ads) mathieuboudreau@Mathieus-MacBook-Pro axondeepseg % napari
  File "/Users/mathieuboudreau/neuropoly/github/axondeepseg/ads_conda/envs/venv_ads/lib/python3.11/site-packages/napari/_qt/__init__.py", line 9, in <module>
    from qtpy import API_NAME, QT_VERSION, QtCore
  File "/Users/mathieuboudreau/neuropoly/github/axondeepseg/ads_conda/envs/venv_ads/lib/python3.11/site-packages/qtpy/__init__.py", line 287, in <module>
    raise QtBindingsNotFoundError from None
qtpy.QtBindingsNotFoundError: No Qt bindings could be found

  File "/Users/mathieuboudreau/neuropoly/github/axondeepseg/ads_conda/envs/venv_ads/bin/napari", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/mathieuboudreau/neuropoly/github/axondeepseg/ads_conda/envs/venv_ads/lib/python3.11/site-packages/napari/__main__.py", line 574, in main
    _maybe_rerun_with_macos_fixes()
  File "/Users/mathieuboudreau/neuropoly/github/axondeepseg/ads_conda/envs/venv_ads/lib/python3.11/site-packages/napari/__main__.py", line 452, in _maybe_rerun_with_macos_fixes
    from napari._qt import API_NAME
  File "/Users/mathieuboudreau/neuropoly/github/axondeepseg/ads_conda/envs/venv_ads/lib/python3.11/site-packages/napari/_qt/__init__.py", line 18, in <module>
    raise ImportError(
ImportError: No Qt bindings could be found.

napari requires either PyQt5 (default) or PySide2 to be installed in the environment.

With pip, you can install either with:
  $ pip install -U 'napari[all]'  # default choice
  $ pip install -U 'napari[pyqt5]'
  $ pip install -U 'napari[pyside2]'

With conda, you need to do:
  $ conda install -c conda-forge pyqt
  $ conda install -c conda-forge pyside2

Our heuristics suggest you are using 'pip' to manage your packages.
(venv_ads) mathieuboudreau@Mathieus-MacBook-Pro axondeepseg % 
```

Our tests don't check on opening the GUI, but I'm curious why this would be thrown now since napari was being installed through pip (via conda) before anyways.

test/morphometrics/test_compute_morphometrics.py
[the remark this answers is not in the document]
mathieuboudreau: This only happens for one single axon, ID 262, and for the "orientation" parameter. 

It's a weird chape, where it's symmetrical but somewhat horshoe shaped 

Screenshot 2024-08-20 at 12 45 04 PM

I'm think it's related to my initial hunch, but maybe just for this sort of edge case?

I think the best solution is to just adopt this new behaviour from one of our depedencies and update our reference morphometric file to use this value now. It didn't change for any of the other "real" axons, so I'd rather do that than use `abs()` for the test condition

joshuacwnewton:
> I'm curious why this would be thrown now since napari was being installed through pip (via conda) before anyways.

Ah, this is my mistake, I think! I realize that I specified `pip install -e .` rather than `pip install -e . plugins/`. And, the `plugins/setup.cfg` file is what specifies `pyqt`:

https://github.com/axondeepseg/axondeepseg/blob/master/plugins/setup.cfg#L12

I presume that adding `plugins/` will install the required dependency?

mathieuboudreau:
> I presume that adding `plugins/` will install the required dependency?

Ah yes that's likely the issue! I forgot that the plugin had its own depedencies too in that folder!

mathieuboudreau:
Do you know what's the syntax here that would work for Windows? 

https://github.com/axondeepseg/axondeepseg/blob/e8e3b31c17d7cc424a2a064453f7bbb2d9d845c9/install_ads.bat#L185

Would it be `ads_conda\envs\venv_ads\Scripts\pip install -e . plugins\ --use-pep517` or still `ads_conda\envs\venv_ads\Scripts\pip install -e . plugins/ --use-pep517`?

.github/workflows/run_tests.yaml
mathieuboudreau: Through some digging I think what was said was that `*` won't work when the branch name has a `/` in it? But `**` will? I'll dig up the discussion again if I can

joshuacwnewton:
I think `plugins\` or even just `plugins` should be okay? I would imagine that the trailing slash is optional.

test/morphometrics/test_compute_morphometrics.py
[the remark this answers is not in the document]
joshuacwnewton: Absolutely! `abs()` was just a hack until we could identify the real cause. :)

.github/workflows/run_tests.yaml
joshuacwnewton: That makes sense! And, we use `/` a _lot_ in SCT branch names, which would explain why PRs haven't been triggering against non-`master` branches.

test/morphometrics/test_compute_morphometrics.py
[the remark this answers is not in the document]
mathieuboudreau: Resolved with [2664221](https://github.com/axondeepseg/axondeepseg/pull/821/commits/2664221b60a52f15254f5ac83429ffa90189fc00) !

mathieuboudreau:
This [35abc3a](https://github.com/axondeepseg/axondeepseg/pull/821/commits/35abc3a7366b39204d2f29827cd2c0a2e0b98731) hasn't resolved the issue. And, I notice that the last raw logs on master (https://github.com/axondeepseg/axondeepseg/actions/runs/10201388130/job/28223085307) has `pyqt` installed, but not the ones for https://github.com/axondeepseg/axondeepseg/actions/runs/10476158607/job/29014463599

But I can't seem to see in the raw logs where/when `pyqt` was getting installed before.

mathieuboudreau:
This 

https://github.com/axondeepseg/axondeepseg/blob/16ccf62ab5bf284257a06f972c46bb3398ae9517/plugins/setup.cfg#L11

in [16ccf62](https://github.com/axondeepseg/axondeepseg/pull/821/commits/16ccf62ab5bf284257a06f972c46bb3398ae9517) got napari working for me after a fresh install

.github/workflows/run_tests.yaml
```diff
@@ -12,11 +12,12 @@ on:
   push:
     branches:
       - master
+      - ac/nnunet_migration
```
joshuacwnewton, line 15: TODO: Revert this change once we're done testing this branch. (Note to self to avoid accidentally merging, hehe.)

mathieuboudreau (changes requested):
I think this PR is pretty close for merging, the real main issue at the moment I think is that:

1) At the end of the install, although it instructs the user to source their bash/zsh script to add the CLI lines to their path, we want to point most of our users to open napari and use our plugin directly; the CLI is moreso for advanced usage
2) After installation, napari can't be called from the CLI, as it's not in the path and/or the venv isn't activated
3) There are no instructions to the user as to how to activate their CLI, and actually doing so is somewhat of a longer command that I'd like as you need to give the path down multiple folders.

So I think for me, I'd like that:

- [ ] At the end of the install, we inform the user to open napari to use ADS's plugin, and that more advanced users have the option of the CLI commands.
- [ ] I'd like maybe an alias to the venv's `napari` so that a quick command can open it, something like `alias ads_napari='$(PATH_TO_ADS)/ads_conda/envs/venv_ads/bin/napari`, and add this to the user's bash/zsh script
- [x] I'd like to maybe add an alias to the venv itself and give it a short name, something likse `alias venv_ads='$alias activate_ads="source activate $(PATH_TO_ADS)/ads_conda/envs/venv_ads/"`, that way users could easily just `source activate venv_ads` and `source deactivate` when wanting to run test or develop.

Those are my main thoughts anyways, open to discussions/alternative solutions!

mathieuboudreau:
@joshuacwnewton I drafted the changes I requested  in [3923de2](https://github.com/axondeepseg/axondeepseg/pull/821/commits/3923de2e70cd9a6362277213c19fb94d1bf873f1), but only for the Mac/Linux install script. Could you try doing the same for the Windows one (I don't have access to a Windows computer to test them)

mathieuboudreau:
Another review item I just noticed is updated the readthedocs, it could either be done here or in PR 800 where we've also been updating them for v5

install_ads
```diff
@@ -0,0 +1,717 @@
...
+    echo "# AXONDEEPSEG (installed on $(date +%Y-%m-%d\ %H:%M:%S))"
+    echo "$DISPLAY_UPDATE_PATH"
+    # Switch between shell
+    if [[ "$THE_RC" == "bash" ]]; then
+      echo "export ADS_DIR=$ADS_DIR"
+    elif [[ "$THE_RC" == "csh" ]]; then
+      echo "setenv ADS_DIR $ADS_DIR"
+    fi
+    echo "alias ads_activate=\"source $ADS_DIR/ads_conda/bin/activate $ADS_DIR/ads_conda/envs/venv_ads/\""
+    echo "alias ads_napari=$ADS_DIR/ads_conda/envs/venv_ads/bin/napari"
```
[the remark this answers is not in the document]
joshuacwnewton: Hrm. This approach works OK for `ads_napari` on Windows+Linux, and `ads_activate` works on Linux, too. But, I can't for the life of me get `ads_activate` to work on my Windows machine. :thinking: 

It calls the `activate.bat` script, and the path to `venv_ads` is indeed correct, but the activation doesn't seem to persist outside of the Python subprocess.

Any of these commands work from the command line, though:

```
C:\Users\Joshua\repos\axondeepseg\ads_conda\Scripts\activate.bat venv_ads
C:\Users\Joshua\repos\axondeepseg\ads_conda\condabin\conda.bat activate venv_ads
```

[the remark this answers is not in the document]
joshuacwnewton: Ah! It's simple enough to just create a file called `ads_activate.bat` on Windows, put it in `bin/`, then have the batch script run the above command.

I'll push a second draft with this approach.

install_ads
```diff
@@ -657,6 +655,9 @@ done
 for file in "$ADS_DIR/$CONDA_DIR"/envs/venv_ads/bin/*download*; do
   cp "$file" "$ADS_DIR/$BIN_DIR/" || die "Problem creating launchers!"
 done
+cp "$ADS_DIR/$CONDA_DIR/envs/venv_ads/bin/napari" "$ADS_DIR/$BIN_DIR/ads_napari" || die "Problem creating launchers!"
```
mathieuboudreau, line 658: Is it best practice to copy a binary from inside the venv like this ? I would think a symbolic link would work just as well without the duplication - but open to either way if this does in fact work

mathieuboudreau: Or is this copy required for the Windows compatibility?

mathieuboudreau:
I've updated the documentation and built them on ReadTheDocs: https://axondeepseg.readthedocs.io/en/jn-install-script/

Open to additional information that needs to be added.

hermancollin (commented):
Wow, I just tested this on my linux machine and it works great! This is so practical, especially for less experienced users. I have to admit I don't 100% understand how this works. When we decide to _add the axondeepseg* scripts to your PATH environment_

install_ads
joshuacwnewton: This is a great question! 

Looking at the history of the original SCT install script, the `cp` lines were added in https://github.com/spinalcordtoolbox/spinalcordtoolbox/pull/1996 (specifically https://github.com/spinalcordtoolbox/spinalcordtoolbox/commit/213881eb101efad48c7ae39234ee6b94afb6dee1). And, even those changes were replacing an even older script called [`update_bin.sh`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/91745bbc385d3bb27cb27de247bccf6adb64d970/install/update_bin.sh). Despite this history, though, I can't seem to find any concrete rationale for why we chose to use `cp` over `ln -s`. And, because the decision was made prior to my joining, I don't have any first-hand knowledge, heh.

----

To try and figure out if it's worth it to switch to softlinks, I've spent the past little while doing some reading, trying to understand what possible risks there could be to using a softlink here, but... I couldn't really find any? Softlinks seem perfectly safe and sensible to use here.

That said, I don't think there's too much harm in using a copy, either, as the launcher files are very light files without much actual logic in them besides importing the entry point function. For example, `ads_napari` looks like this:

```python
#!/home/joshua/repos/axondeepseg/ads_conda/envs/venv_ads/bin/python
# -*- coding: utf-8 -*-
import re
import sys
from napari.__main__ import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

Still, that logic comes with some risk? (For example, napari could change its package structure to no longer use [`__main__.py :: main()`](https://github.com/napari/napari/blob/main/napari/__main__.py#L571-L584), along with a user updating ADS using `ads_activate` + `pip install -U`.)

But, because that risk seems relatively small, I'm thinking that it might be better to stick with the "known unknowns" (`cp`) vs. venturing into ["unknown unknowns"](https://en.wikipedia.org/wiki/There_are_unknown_unknowns) (`ln -s`). Maybe there are some risks to softlinks that I've missed (e.g. weird filesystem quirks, network drives, Windows compatibility, etc.) that wouldn't be present with simple copies.

mathieuboudreau: >But, because that risk seems relatively small, I'm thinking that it might be better to stick with the "known unknowns" (cp) vs. venturing into ["unknown unknowns"](https://en.wikipedia.org/wiki/There_are_unknown_unknowns) (ln -s). Maybe there are some risks to softlinks that I've missed (e.g. weird filesystem quirks, network drives, weird ComputeCanada restrictions, Windows compatibility, etc.) that wouldn't be present with simple copies.

Ok sounds good! Let's do that then!

mathieuboudreau (approved):

mathieuboudreau:
To me things look good to merge into our other branch now - especially since it worked well for @hermancollin.

We'll have additional users testing the installation when reviewing PR #800, so unless @joshuacwnewton you have any additional changes you want to make, I'd say let's go and merge!

docs/source/documentation.rst
```diff
@@ -89,7 +40,13 @@ To install AxonDeepSeg, in a terminal window (macOS or Linux) or Anaconda Prompt
     git clone https://github.com/neuropoly/axondeepseg.git
     cd axondeepseg
 
-Virtual environments are a tool to separate the Python environment and packages used between Python projects. They allow for different versions of Python packages to be installed and managed for the specific needs of your projects. There are several virtual environment managers available, but the one we recommend and will use in our installation guide is `conda <https://conda.io/docs/>`_, which is installed by default with Miniconda. We strongly recommend you install into a virtual environment.
+Then, for Linux or macOS systems, run::
+
+    ./install_ads
+
+and follow the prompts. Or, if you want to isntall AxonDeepSeg with default settings, run::
```
joshuacwnewton, line 47:
```suggestion
and follow the prompts. Or, if you want to install AxonDeepSeg with default settings, run::
```

docs/source/documentation.rst
```diff
@@ -105,34 +62,9 @@ Virtual environments are a tool to separate the Python environment and packages
...
-
-.. NOTE :: To update an already cloned AxonDeepSeg package, pull the latest version of the project from GitHub and re-install the application:
-   ::
-
-        cd axondeepseg
-        git pull
-        pip install -e . plugins/
-
-.. WARNING :: When re-installing the application, the model folders in ``AxonDeepSeg/models`` will be deleted and re-downloaded. Please do not store valuable data in these folders.
+    install_ads.bat
```
joshuacwnewton, line 67: I think it might be good to mention the additional steps required in order to add the CLI scripts to the PATH on Windows. We do this for SCT [here](https://spinalcordtoolbox.com/user_section/installation/windows.html#installing-sct).

docs/source/documentation.rst
```diff
@@ -105,34 +62,9 @@ Virtual environments are a tool to separate the Python environment and packages
...
 
 
 Testing the installation
```
joshuacwnewton, line 70: I can't highlight this line (it's outside the diff) but on line 72, there is a warning that the virtual environment needs to be activated to run the quick test.

This is no longer necessary, since `axondeepseg_test` is now on the PATH after installation.

docs/source/documentation.rst
```diff
@@ -153,14 +85,14 @@ This integrity test automatically performs the axon and myelin segmentation of a
...
 ~~~~~~~~~~~~~~~~~~
 
-To run the entire testing suite (more code coverage), go to your AxonDeepSeg project directory on the terminal and run ``py.test``::
+To run the entire testing suite (more code coverage), go to your AxonDeepSeg project directory on the terminal, activate the virtual environment, and then run ``py.test``::
 
     cd axondeepseg
+    ads_activate
     py.test --cov AxonDeepSeg/ --cov-report term-missing
 
 If all tests pass, AxonDeepSeg was installed succesfully.
```
joshuacwnewton, line 94: This is probably out of scope for this PR, but something to note: SCT has a dedicated CLI script for exposing `pytest` called `sct_testing`.

Maybe ADS could do something similar? For example, in `axondeepseg_test`, you could add a `--full` option that will run all of the the pytest tests. That way, it isn't necessary to `cd` and activate the environment?

docs/source/documentation.rst
```diff
@@ -181,6 +113,19 @@ Below is a short tutorial describing how to use the AxonDeepSeg plugin for Napar
...
+
+To use AxonDeepSeg in a Python shell, first activate your virtual environment::
+
+    ads_activate
+
+.. NOTE :: To switch back to your default environment, run:
+  ::
+
+       conda deactivate
+
```
joshuacwnewton, line 128: I'm wondering how useful this section is on its own. Since we expose the `axondeepseg_*` scripts already, is there a reason for users to activate the conda venv besides updating ADS?

(I might be missing some additional usage of ADS here...)

docs/source/documentation.rst
```diff
@@ -181,6 +113,19 @@ Below is a short tutorial describing how to use the AxonDeepSeg plugin for Napar
```
joshuacwnewton, line 113: I think because of the new section "Python Shell use", the sentence on line 110 makes a little less sense.

> Below is a short tutorial describing how to use the AxonDeepSeg plugin for Napari.

Maybe it could be good to directly link to the short tutorial, rather than saying "Below"?

```markdown
In the [Manual Correction](#manual-correction-of-segmentation-masks) section of the documentation, we provide a short tutorial describing how to use the AxonDeepSeg plugin for Napari.
```

docs/source/documentation.rst
```diff
@@ -32,55 +32,6 @@ Installation
...
-
-To install Miniconda, go to the `Miniconda installation website <https://conda.io/miniconda.html>`_ and click on the Python 3.x version installer compatible with your Windows system (64 bit recommended). After the download is complete, execute the downloaded file, and follow the instructions. If you are unsure about any of the installation options, we recommend you use the default settings.
-
-git
----
-``git`` is a software version control system. Because AxonDeepSeg is hosted on GitHub, a service that hosts ``git`` repositories, having ``git`` installed on your system allows you to download the most up-to-date development version of AxonDeepSeg from a terminal, and also allows you to contribute to the project if you wish to do so.
-
-To install ``git``, please follow instructions for your operating system on the `git website <https://git-scm.com/downloads>`_
-
 Install AxonDeepSeg
```
joshuacwnewton, line 35: The sentence below this header mentions "Anaconda prompt" for Windows, but we can change this to say "Command Prompt" now, since users no longer have to install Anaconda. :)

joshuacwnewton (commented):
I think given that we're not necessarily committing to ADS 5.0.0 yet, and that we have more time to test between now and then, I agree this is a good place to merge.

I just had a quick look over the documentation changes, and had some small comments. But apart from this, I think we're good to go?

Do note, though, that if you ever have any troubles with the `install_ads` script whatsoever in the coming weeks/months/years, please feel free to ping me on any issue or PR. Any issues that ADS faces with this script will likely affect SCT's install scripts too, so I'm very interested in any updates ADS makes to the script. I'm happy to provide any information I can from having maintained this script for several years now. :)

I'm happy to reach out to ADS whenever SCT encounters any issues/improvements to our scripts, too!

docs/source/documentation.rst
mathieuboudreau: Good catch!

docs/source/documentation.rst
mathieuboudreau: Change made - thanks!

docs/source/documentation.rst
mathieuboudreau: I guess it's more for devs, specifically internally. Our wiki is pretty out of date - I'll update the title of this section to make it more explicit that its for developers

docs/source/documentation.rst
mathieuboudreau: Interesting - I'll open an issue and we'll get that done later!

mathieuboudreau: https://github.com/axondeepseg/axondeepseg/issues/822

docs/source/documentation.rst
mathieuboudreau: Deleted!

docs/source/documentation.rst
mathieuboudreau: Thanks!

docs/source/documentation.rst
mathieuboudreau: Thanks - done!

mathieuboudreau:
LGTM! I'm going to merge, and we can continue any further discussions in PR #800 or in a new issue. Thank you SO MUCH @joshuacwnewton 🙏🙏🙏

PR #830, opened 2024-09-26
Change "download_models" to "download_model"

mathieuboudreau:
## Checklist

## Description

The documentation currently says to the users to call `download_model`, however our actual CLI command is `download_models`. As a simple fix, we can simply update the doc; however, it is a bit inconsistent that in actuality, our python **module** name is `download_model` (and makes more sense, since we can only download one model at a time I believe). So maybe changing the CLI command would be a better solution considering we're releasing 5.0 soon (this would require more changes though in the tests, installation, and CI files). Open to opinions/discussion

## Linked issues

#829

### DIFF
--- a/docs/source/documentation.rst
+++ b/docs/source/documentation.rst
@@ -153,11 +153,11 @@ Here are the details of all the models currently supported by AxonDeepSeg:
 
 To download these models, you must first have AxonDeepSeg installed. Afterwards, run::
 
-    download_model -m <model name> -t <model type>
+    download_models -m <model name> -t <model type>
 
 where <model name> is the full name (e.g. *model_seg_rat_axon-myelin_SEM*) and <model type> is either *light* or *ensemble*. To view available models and their details, run::
 
-    download_model --list
+    download_models --list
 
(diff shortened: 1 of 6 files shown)


joshuacwnewton (dismissed):
For comparison, SCT has a similar command called `sct_download_data`, so I think singular (`model`) does make sense.

But, as far as a documentation change goes, this LGTM!

hermancollin:
Yeah I agree it should be the other way around. 

> So maybe changing the CLI command would be a better solution considering we're releasing 5.0 soon (this would require more changes though in the tests, installation, and CI files). Open to opinions/discussion.

Should we do this right now or merge this as a temporary fix and then open a new PR?

mathieuboudreau:
>Should we do this right now or merge this as a temporary fix and then open a new PR?

I think we should do this now - let me see if I can catch all instances quickly

mathieuboudreau:
@hermancollin I went that route and updated all the files (did a replace all in VS code), the tests pass so I think it would be safe to merge.

hermancollin (approved):
thank you for this fix

PR #834, opened 2024-11-09
Pin acvl_utils!=0.2.1

mathieuboudreau:
## Checklist

## Description

nnUnetV2 depends on acvl_utils, but didn't pin a version for it. Two days ago acvl_utils == 0.2.1 was released, but that one imports a package that is never installed (blosc2) which has lead to failing overnight tests, https://github.com/axondeepseg/axondeepseg/actions/runs/11753375787

I propose to pin the version to <0.2.1, because if acvl_utils didn't catch this missing dependency with tests or a CI, then maybe it's better to trust that they didn't miss something else in this most recent push and download an earlier version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,7 +16,7 @@ jupyter
 openpyxl
 qtconsole<5.4.2
 napari
-acvl_utils<0.2.1
+acvl_utils!=0.2.1
 nnunetv2==2.2.1
 loguru
 torch<2.4.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,6 +16,7 @@ jupyter
 openpyxl
 qtconsole<5.4.2
 napari
+acvl_utils<0.2.1
 nnunetv2==2.2.1
 loguru
 torch<2.4.0


mathieuboudreau:
https://github.com/MIC-DKFZ/nnUNet/issues/2589

mathieuboudreau:
@joshuacwnewton maybe you could approve, since I saw you dealt with this similarly in SCT?

requirements.txt
```diff
@@ -16,6 +16,7 @@ jupyter
 openpyxl
 qtconsole<5.4.2
 napari
+acvl_utils<0.2.1
```
joshuacwnewton, line 19: I chose to do `acvl_utils!=0.2.1`, since I'm assuming that if `acvl_utils==0.2.2` will be released, it will include a bugfix? Maybe this is overly confident, but it would save us the hassle of having to update this once more in the future.

(It might also be good to add a comment explaining the limitation? SCT's requirements.txt is peppered with comments, though I guess `git blame` functions similarly to find out the reasons for each limitation.)

joshuacwnewton (approved):
Approving since my changes are minor and not a huge deal.

requirements.txt
joshuacwnewton: Ah, funnily enough, the fix is already out: https://github.com/MIC-DKFZ/acvl_utils/issues/2#issuecomment-2470346349

mathieuboudreau: Ah great! In that case, your solution is likely best

hermancollin (approved):
LGTM

PR #838, opened 2024-12-01
Class value check - Change error to warning

### LINKED ISSUE #837: `extract_from_nnunet_prediction()` error is annoying
https://github.com/axondeepseg/axondeepseg/blob/a7d16e1b4b9a502462557c027b68fa090df37677/AxonDeepSeg/apply_model.py#L76-L77

When the segmentations are saved as separate files in `extract_from_nnunet_prediction`, the first thing we do is check if the class value exists in the masks, and otherwise throw an error. This shouldn't be an error because
1. sometimes a mask could be empty (i.e. no axon in the image). This happens often with my multiclass model, where some classes may not be present in the image.
2. when we segment many images using the CLI, the model will first be applied to every image (which could potentially take a long time), and then all outputs are divided in separate masks. This error interrupts the second phase, so many masks could be missing

Simple fix: it should be a warning instead

mathieuboudreau:
Resolves #837

## Checklist

### DIFF
--- a/AxonDeepSeg/apply_model.py
+++ b/AxonDeepSeg/apply_model.py
@@ -4,6 +4,7 @@ import numpy as np
 import torch
 from loguru import logger
 from typing import List, Literal, NoReturn
+import warnings
 
 # AxonDeepSeg imports
 from AxonDeepSeg.visualization.merge_masks import merge_masks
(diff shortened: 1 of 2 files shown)


AxonDeepSeg/apply_model.py
```diff
@@ -74,7 +75,7 @@ def extract_from_nnunet_prediction(pred, pred_path, class_name, class_value) ->
     pred_path = ads_utils.convert_path(pred_path)
 
     if not np.any(pred == class_value):
-        raise ValueError(f'Class value {class_value} not found in the raw prediction.')
+        warnings.warn(f'Class value {class_value} not found in the raw prediction.')
```
hermancollin, line 78:
```suggestion
        logger.warning(f'Class value {class_value} not found in the raw prediction.')
```

Maybe we can use logger instead for prettier output in the console.

hermancollin (commented):
wow that was fast! Thank you for helping with this.

AxonDeepSeg/apply_model.py
```diff
@@ -4,6 +4,7 @@
 import torch
 from loguru import logger
 from typing import List, Literal, NoReturn
+import warnings
```
hermancollin, line 7:
```suggestion
```

hermancollin (approved):

PR #844, opened 2024-12-12
Refactor config.py into ADS/params.py

### LINKED ISSUE #724: Consider moving the config.py file inside the AxonDeepSeg directory.
Currently, the [config.py](https://github.com/axondeepseg/axondeepseg/blob/master/config.py) file is located outside the AxonDeepSeg directory.

This makes it so when we use it in a script, we will import it by typing
`import config` or `from config import <insert parameter you are trying to import>`.

The problem with this is that it makes it look like config file comes from outside of the ADS project.
An issue recently opened on the napari-plugin page (https://github.com/axondeepseg/napari-plugin/issues/4) is an example of how missleading this can be. If config.py was inside of the AxonDeepSeg directory, the error would have been something like `No module named 'AxonDeepSeg'` which would have been more clear that AxonDeepSeg might not be properly installed.

Moreover, we have the [AxonDeepSeg/params.py](https://github.com/axondeepseg/axondeepseg/blob/master/AxonDeepSeg/params.py) file which existed before `config.py` and serves a similar purpose.

Therefore, I suggest moving the content of `config.py` inside `params.py`

mathieuboudreau:
## Checklist

## Linked issues

Resolves #724


hermancollin (dismissed):

AxonDeepSeg/apply_model.py
```diff
@@ -8,7 +8,7 @@
 # AxonDeepSeg imports
 from AxonDeepSeg.visualization.merge_masks import merge_masks
 from AxonDeepSeg import ads_utils
-from config import nnunet_suffix
+from AxonDeepSeg.params import nnunet_suffix
 from AxonDeepSeg.params import intensity
```
Stoyan-I-A, line 12: In this file, you could group the two `AxonDeepSeg.params` import together

Stoyan-I-A (changes requested):

AxonDeepSeg/apply_model.py
Stoyan-I-A: In some other files, it is also imported twice, but it's a bit different. They will have `from AxonDeepSeg.params import ...` and another line will have `import AxonDeepSeg.params`.

In those cases, it might be easier to keep it the way they are.

mathieuboudreau: Done in [34f8ca8](https://github.com/axondeepseg/axondeepseg/pull/844/commits/34f8ca8551e6b6843b6ceb8afbaf6b46a7ddaec3) !

Stoyan-I-A (approved):
LGTM

PR #845, opened 2024-12-12
Add CLI option to run all tests

### LINKED ISSUE #822: Create CLI for pytest-ing in ADS
See comment from @joshuacwnewton in PR #821:

> This is probably out of scope for this PR, but something to note: SCT has a dedicated CLI script for exposing `pytest` called `sct_testing`.
>
> Maybe ADS could do something similar? For example, in `axondeepseg_test`, you could add a `--full` option that will run all of the the pytest tests. That way, it isn't necessary to `cd` and activate the environment?

_Originally posted by @joshuacwnewton in https://github.com/axondeepseg/axondeepseg/pull/821#discussion_r1736801956_

mathieuboudreau:
## Checklist

## Notes

I'm not adding a new test for this, because it would cause an infinite loop 😃

## Linked issues

Resolves #822

### DIFF
--- a/AxonDeepSeg/integrity_test.py
+++ b/AxonDeepSeg/integrity_test.py
@@ -8,6 +8,11 @@ from pathlib import Path
 import numpy as np
 import tempfile
 import shutil
+import sys
+import pytest
+from loguru import logger
+
+import argparse
 
 # AxonDeepSeg imports
 from AxonDeepSeg.testing.segmentation_scoring import pw_dice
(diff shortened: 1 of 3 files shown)


AxonDeepSeg/integrity_test.py
```diff
@@ -17,48 +22,70 @@
...
+    parser = argparse.ArgumentParser(description='Run AxonDeepSeg tests.')
 
-            # Generate separate axon and myelin masks of the segmentation output
-            print('Generating axon and myelin segmentation masks and saving.')
-            gt_axon = mask > 200 # Generate binary image with the axons for the ground truth (myelin=127, axon=255)
-            gt_myelin = np.logical_and(mask >= 50, mask <= 200) # Generate binary image with the myelin for the ground truth (myelin=127, axon=255)
-            pred_axon = pred > 200 # Generate binary image with the axons for the segmentation (myelin=127, axon=255)
-            pred_myelin = np.logical_and(pred >= 50, pred <= 200) # Generate binary image with the myelin for the segmentation (myelin=127, axon=255)
+    parser.add_argument('--integrity', action='store_true', help='Run the integrity test. (default)')
+    parser.add_argument('--full', action='store_true', help='Run the the full test suite.')
```
joshuacwnewton, line 76: If I'm understanding correctly, `--full` is the only argument that has any effect (i.e. whether `--integrity` is passed has no bearing)?

To avoid confusion, maybe it's best to just remove `--integrity`, and explain what happens in the description for `--full`:

```suggestion
    parser.add_argument('--full', action='store_true', help='Run the the full test suite.'
                        'If not specified, then only the integrity tests will be run.')
```

AxonDeepSeg/integrity_test.py
```diff
@@ -17,48 +22,70 @@
...
+    else:
+        # Run the integrity test
+        integrity_test()
+    return 0
 
-        # Else, there is a problem in the installation
-        print("Integrity test failed... ")
-        return -1
+if __name__ == "__main__":
+    with logger.catch():
```
joshuacwnewton, line 90: I had never seen [`loguru.catch()`](https://loguru.readthedocs.io/en/stable/api/logger.html#loguru._logger.Logger.catch) before... neat!!

> Return a decorator to automatically log possibly caught error in wrapped function.
>
> This is useful to ensure unexpected exceptions are logged, the entire program can be wrapped by this method.

AxonDeepSeg/integrity_test.py
```diff
@@ -17,48 +22,70 @@
...
-        return 0
+    if args.full:
+        # Run the full test suite
+        sys.exit(pytest.main(["-x", Path(__file__).resolve().parent.parent / "test"]))
 
-    except IOError:
+    else:
+        # Run the integrity test
+        integrity_test()
+    return 0
```
joshuacwnewton, line 87: Rather than `return 0` for `main()`, maybe it would be useful to preserve the result of `integrity_test()` and return that instead?

AxonDeepSeg/integrity_test.py
```diff
@@ -17,48 +22,70 @@
 from config import axonmyelin_suffix
 
 def integrity_test():
+        try:
+
+            # get path of directory where AxonDeepSeg was installed
+            dir_path = Path(__file__).resolve().parent
```
joshuacwnewton, line 28: I think the reason why the GitHub diff is so weird is because `integrity_test` now has an indentation of 8 spaces?

I think if you revert this back to 4 spaces, the GitHub diff should end up looking much neater.

joshuacwnewton (changes requested):
A very nice improvement! :tada:

AxonDeepSeg/integrity_test.py
mathieuboudreau: Yeah! I think @hermancollin might have been the one to first introduce it?

AxonDeepSeg/integrity_test.py
mathieuboudreau: How could that output be leveraged ? (it's already all printed for the user)

AxonDeepSeg/integrity_test.py
mathieuboudreau: Ahh woops - I had an added if/else while testing things out and didn't reset the tabs - thanks!

AxonDeepSeg/integrity_test.py
joshuacwnewton: Honestly, I don't know, hehe. I just figured it would be more accurate than `return 0` even when the tests fail? 

But maybe `return 0` is meant to signify whether the tests ran successfully, rather than whether the tests passed? If so, then ignore this comment!

AxonDeepSeg/integrity_test.py
```diff
@@ -62,3 +66,26 @@ def integrity_test():
...
+    parser = argparse.ArgumentParser(description='Run AxonDeepSeg tests.')
+
+    parser.add_argument('--full', action='store_true', help='Run the the full test suite.'
+                        'If not specified, then only the integrity tests will be run.')
+
+    args = parser.parse_args()
+
+    if args.full:
+        # Run the full test suite
+        sys.exit(pytest.main(["-x", Path(__file__).resolve().parent.parent / "test"]))
```
joshuacwnewton, line 82: Just out of curiosity, why have we specified the `-x` option here?

> ```
>   -x, --exitfirst       Exit instantly on first error or failed test
> ```

If a user runs `axondeepseg_test --full` and there are multiple errors, then it might end up wasting a lot of time when troubleshooting with the users, as the errors would have to be solved iteratively, 1 by 1, since only one error would be shown at a time.

joshuacwnewton (dismissed):
Apart from my one nitpick about `-x`, this LGTM!

AxonDeepSeg/integrity_test.py
mathieuboudreau: Yep I think its the latter, if other errors in the software arose, they’d produce different integer values

AxonDeepSeg/integrity_test.py
mathieuboudreau: Nevermind my last comment, i didnt fully read yours. It came from an example for using pytest within python, i thought it was used for another reason. Ill change it!

mathieuboudreau: Done!

joshuacwnewton:
Sorry, I was a bit unclear, hehe. No need to dismiss my stale review and then re-request a review, as the PR LGTM!

((Often when reviewing PRs for SCT, if the PR is 99% good to go, and there's just a small nitpick with a clear/obvious change, I'll pre-approve the PR, signaling that I think it's good to merge as soon as the nitpick is addressed without any further review.))

joshuacwnewton (approved):

PR #847, opened 2024-12-13
CLI: Log/print commit hash, and --version option

### LINKED ISSUE #56: Displayed version of ADS should reflect current commit
Currently, when running ADS from github master (or any other branch) repository, the version displayed is something like that:
~~~
axondeepseg
AxonDeepSeg v.0.3
usage: axondeepseg [-h] -t {SEM,TEM} -i IMGPATH [IMGPATH ...] [-m MODEL]
                   [-s SIZEPIXEL] [-v {0,1,2,3}] [-o OVERLAP]
~~~

However, the problem is that the displayed version does not reflect the current state of ADS. For example, two users can run ADS with the apparent "v.0.3" version, but with two different commits, resulting in potentially two different behaviours, hence this might cause debugging problems in the future.

The solution is to display the commit # for github installations of ADS, and version # for stable release installation.

mathieuboudreau:
## Checklist

## Linked issues

Resolves #56 (!)


mathieuboudreau:
Outputs:

### Normal CLI run (with image)

```
(venv_ads) mathieuboudreau@arm64-apple-darwin20 axondeepseg % axondeepseg -i test/__test_files__/__test_demo_files__/image.png
2024-12-13 12:21:15.049 | INFO     | AxonDeepSeg.segment:main:212 - AxonDeepSeg v.5.0.0 (58db2631bdba38e7a7dcda7a21bb164ea331c47b)
perform_everything_on_gpu=True is only supported for cuda devices! Setting this to False
2024-12-13 12:21:15.064 | INFO     | AxonDeepSeg.apply_model:axon_segmentation:148 - Running inference on device: cpu
2024-12-13 12:21:17.593 | INFO     | AxonDeepSeg.apply_model:axon_segmentation:158 - Model successfully loaded.
```

etc.

### With --version flag

```

(venv_ads) mathieuboudreau@arm64-apple-darwin20 axondeepseg % axondeepseg --version
AxonDeepSeg v.5.0.0 (58db2631bdba38e7a7dcda7a21bb164ea331c47b)
(venv_ads) mathieuboudreau@arm64-apple-darwin20 axondeepseg %


```

AxonDeepSeg/segment.py
```diff
@@ -247,6 +255,13 @@ def main(argv=None):
 
     # Processing the arguments
     args = vars(ap.parse_args(argv))
+
+    if "version" in args:
+        print(f"AxonDeepSeg v.{AxonDeepSeg.__version__} ({git.Repo(search_parent_directories=True).head.object.hexsha})")
+        sys.exit(0)
+
+    logger.info(f"AxonDeepSeg v.{AxonDeepSeg.__version__} ({git.Repo(search_parent_directories=True).head.object.hexsha})")
```
joshuacwnewton, line 263: It might be good to make this string a constant instead? Maybe inside `__init__.py`, since we already define `__version__` there?

Something like:

```
import git

__version__ = "5.0.0"
__git_version__ = git.Repo(search_parent_directories=True).head.object.hexsha
__version_string__ = f"AxonDeepSeg v{__version__} ({__git_version__})"
```

That way, it's easy to access `AxonDeepSeg.__version_string__` not only in `segment.py`, but anywhere else we may want it too.

joshuacwnewton: Also, tiny nitpick, but how do you feel about `v{__version__}` (i.e. `v5.0.0` rather than `v.5.0.0`)?

joshuacwnewton (changes requested):
TIL argparse has a built-in option for specifying the version: https://stackoverflow.com/questions/15405636/pythons-argparse-to-show-programs-version-with-prog-and-version-string-formatt

This way, we can keep `-i` as mandatory, and we won't have to check `args` for `version`. :)

AxonDeepSeg/segment.py
mathieuboudreau: good point for both! 

I'm also fine with the second re point, I don't recall any reason we set it that way in the first place but I agree the second appears better and is more consistent with othter softwares styles

mathieuboudreau: @joshuacwnewton all done! Here's the outputs:

(venv_ads) mathieuboudreau@arm64-apple-darwin20 axondeepseg % axondeepseg -V
AxonDeepSeg v5.0.0 (3bddcb6eaa201da86e43074500bb3b4eae24f787)
(venv_ads) mathieuboudreau@arm64-apple-darwin20 axondeepseg % axondeepseg --version
AxonDeepSeg v5.0.0 (3bddcb6eaa201da86e43074500bb3b4eae24f787)
(venv_ads) mathieuboudreau@arm64-apple-darwin20 axondeepseg % axondeepseg -i test/__test_files__/__test_demo_files__/image.png
2024-12-13 13:27:18.198 | INFO     | AxonDeepSeg.segment:main:260 - AxonDeepSeg v5.0.0 (3bddcb6eaa201da86e43074500bb3b4eae24f787)
perform_everything_on_gpu=True is only supported for cuda devices! Setting this to False
2024-12-13 13:27:18.214 | INFO     | AxonDeepSeg.apply_model:axon_segmentation:148 - Running inference on device: cpu

mathieuboudreau:
These changes broke something for windows only, I see one issue but it's not an error, 

https://github.com/axondeepseg/axondeepseg/blob/df8184cf83e6578869e78fea9f5f32f0336bf9a4/install_ads.bat#L64-L67

with the proposed additions to __init__.py, it's not possible for the version number to be the last line anymore.

Besides that, github actions complains about `git` missing during an import, despite both the ADS and RTD requirements containing `gitpython`

mathieuboudreau:
Weiiiiird https://stackoverflow.com/questions/67459422/modulenotfound-no-module-named-git

AxonDeepSeg/segment.py
joshuacwnewton: Looks great! Though, I notice that the `logger.info` call could be updated to use  `AxonDeepSeg.__version_string__` as well?

mathieuboudreau:
@joshuacwnewton I feel like I've exhausted my windows-error debugging capabilities for the day - if you get a chance I think maybe a fresh set of eyes (or maybe with more windows/setup experience) could help!

AxonDeepSeg/segment.py
mathieuboudreau: Done!

joshuacwnewton:
Taking a look now! :hearts:

joshuacwnewton:
Looking at `GitPython`'s README, I saw a note for [Windows support](https://github.com/gitpython-developers/GitPython?tab=readme-ov-file#windows-support), which led to https://github.com/gitpython-developers/GitPython/issues/525 and subsequently https://github.com/godaddy/tartufo/issues/66.

Given that our git usage is so slight, I would maybe try switching over to `pygit2`? The changes involved seem very straightforward.

AxonDeepSeg/__init__.py
```diff
@@ -1,7 +1,15 @@
 import os
 import sys
+import pygit2
```
joshuacwnewton, line 3: Welp, it looks like the error was _not_ related to `GitPython`: 

```
  ModuleNotFoundError: No module named 'pygit2'
```

It seems like it instead has something to do with the build process of the `axondeepseg` package, and the fact that we have a non-stdlib import in our `__init__.py`. 

I'll keep debugging!

mathieuboudreau: Well, in that case maybe we could just move all of the git version stuff out of init and into ads_utils.py and then import from there - that should be a quick fix?

joshuacwnewton: Okay, I think the reason this is failing is because in `setup.py`, we import `AxonDeepSeg.__version__`, and thus `import pygit2`/`import GitPython` gets hit when installing our package in editable mode.

While looking up this pattern, I found an [old post](https://stackoverflow.com/a/2058872) with a little edit at the bottom describing something similar to our dilemma:

> UPDATE (2017): coverage.py no longer imports itself to get the version. Importing your own code can make it uninstallable, because you product code will try to import dependencies, which aren't installed yet, because setup.py is what installs them.

But, the dependencies _should_ be installed? They're separate lines! Weird.

Anyway, I think maybe we should fix this by not importing from `AxonDeepSeg` to get the version. I'll give it another shot...

mathieuboudreau: >Okay, I think the reason this is failing is because in setup.py, we import AxonDeepSeg.__version__, and thus import pygit2/import GitPython gets hit when installing our package in editable mode

Ahh I missed that!

>But, the dependencies should be installed? They're separate lines! Weird.

ikr! installed but maybe the env/PATH isn't "seen" yet during setup for some reason?

joshuacwnewton: It looks like the import of `AxonDeepSeg` into `setup.py` was indeed the problem! :tada:

mathieuboudreau: Fantastic! Thank you so much!

joshuacwnewton (approved):
LGTM! :tada:

Issue #747, opened 2023-06-21
How to change pixel size in napari


hermancollin:
**Describe the problem**
When you load an image in napari and click on `Apply ADS model`, the user is prompted for the pixel size of the image. However, after this initial prompt, I don't think there is a way to change it. For example, if you load a new image and would like to use a different pixel size, there is no way to change it.

**Proposed solution**
Maybe some field where the user can change this value.

mathieuboudreau:
How did we handle this in FSLeyes? I think if we changed images, it prompted us again, but I'm not 100% certain

Stoyan-I-A:
> How did we handle this in FSLeyes?

We asked again.

The reason why this is different with napari is because in the napari plugin, it is added in the image's metadata. 

But if a new image is loaded and the pixel size file is not present, I expect it to prompt the user again. Otherwise it might be a bug. I'll have to check

Stoyan-I-A:
>  if a new image is loaded and the pixel size file is not present, I expect it to prompt the user again. Otherwise it might be a bug.

I checked and the feature currently works as intended:
- If an image is loaded and the pixel size file is not present, a prompt will ask the user for the pixel size.
- If a new different image is loaded and the pixel size file is not present, it will prompt the user again.

Now, we can discuss if we want a way to change the pixel size but ideally the user would enter the correct pixel size and then adjust the zoom factor if needed.

mathieuboudreau:
Is this still needed now @hermancollin, or has it become obsolete with the new models?

Issue #698, opened 2022-11-16
Prepare support for 3-class segmentation


hermancollin:
In order to support 3-class segmentation (context: unmyelinated fibers), we will need to change some stuff on the ADS side. For ivadomed, nothing really changes: we will use 3 ground-truths in the BIDS derivatives and change the training config accordingly. However, in ADS we will need to add some flexibility, notably:
- For the segmentation process, `axon_segmentation(...)` will need to also save the third prediction. Maybe also add flexibility to the `merge_masks(...)` if we want to support it: https://github.com/axondeepseg/axondeepseg/blob/821074c2c8b539bcec69686cce72304656124d51/AxonDeepSeg/apply_model.py#L46-L50
Not exactly sure how we would handle the 3rd class using grayscale format in the combined prediction image though
- Most of the work will probably be on the morphometrics process. Thanks to @Stoyan-I-A's refactoring, this should be easier to do because I think we will need to add columns for the 3rd class metrics (e.g. area, etc.). Fortunately, processing unmyelinated axons should be the exact same as processing axons. I'm thinking of adding a parameter to `get_axon_morphometrics(...)` indicating if we want 3-class morphometrics, and if so will load the 3rd segmentation and run the usual axon metrics on it. Not sure how we could merge this data in the morphometrics file (because the "myelin" columns will not apply to unmyelinated axons). In that case, maybe we would need 2 separate morphometrics file, but I don't really like that idea.

mariehbourget:
> * Not exactly sure how we would handle the 3rd class using grayscale format in the combined prediction image though

That's a great point, we could add flexibility in merge mask by assigning the value based on the number of class, ex:
* 2 class = 127 and 255
* 3 class = 85, 170 and 255

Also, we have to take in consideration how fsleyes loads the mask overlays, as discussed in the ADS meeting last week, I believe it is hardcoded with fixed values and only for 2 class segmentation at the moment.

mathieuboudreau:
>3 class = 85, 170 and 255

Why don't we just use the old values (127 and 255) for consistency between myelin and axon (and because these will be generated by ADS in the "first draft" segmentation), and add an intermediate value (eg 65 or 200) for the third axon class (whichever is closer to axon, I can't remember).

mathieuboudreau:
The values don't really matter as far as I understand it except for them to be visually "contrasting"; they get separated to 0s and 1s later on in the code anyways right when we "separate" the masks.

[...]

PR #855, opened 2025-01-09
Update imread/imwrite functionality and tests

### LINKED ISSUE #806: our `imread` and `imwrite` are barely tested
**Describe the problem**
The `ads_utils.imread()` and `imwrite()` functions are barely unit tested. In _test_ads_utils.py_, we only have 2 tests on imread.

**Proposed solution**
Given how extensively we use these functions everywhere, I think we should add a few more unit tests. This is even more important now that ivadomed is no longer used, because we are responsible for more image validation and preprocessing on the ADS side.

mathieuboudreau:
## Checklist

## Linked issues

Resolves #806

### DIFF
--- a/AxonDeepSeg/ads_utils.py
+++ b/AxonDeepSeg/ads_utils.py
@@ -107,23 +107,64 @@ def imread(filename):
                                f"supported. AxonDeepSeg supports the following "
                                f"file extensions:  '.png', '.tif', '.tiff', '.jpg' and '.jpeg'.")
 
-    # Load image
-    if str(file_ext) in ['.tif', '.tiff']:
-        raw_img = imageio.v2.imread(filename, format='tiff-pil')
-        if len(raw_img.shape) > 2:
-            raw_img = imageio.v2.imread(filename, format='tiff-pil', mode='L')
-    else:
-        raw_img = imageio.v2.imread(filename)
-        if len(raw_img.shape) > 2:
-            raw_img = imageio.v2.imread(filename, mode='L')
-    img = imageio.core.image_as_uint(raw_img, bitdepth=8)
+    props = imageio.v3.improps(filename) # new in v3 - standardized metadata
+    _img = imageio.v3.imread(filename)
+
+    if _img.dtype==np.float32 or _img.dtype==np.float64:
+        if len(_img.shape) > 2:
+            raise IOError(f"Multichannel 32bit and 64bit float images are not supported. Please convert the image to 8bit or 16bit, or make a single channel image.")
+
+    # brute force fall back to support backward compatibility 
+    if '.tif' in file_ext:
+        img = np.expand_dims(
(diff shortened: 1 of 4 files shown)


AxonDeepSeg/ads_utils.py
```diff
@@ -107,23 +107,64 @@ def imread(filename):
...
+       img = img[:, :, 0]
+
     return img
 
 def imwrite(filename, img, format='png'):
     """ Write image.
     """
     # check datatype:
-    if img.dtype == 'float64':
+    conversion_list = ['float64', 'float32', 'float16', 'uint16']
```
hermancollin, line 165: hmm just a heads up that we use `ads_utils.imwrite()` to write the instance segmentations in #742. If there are more than 255 axons in the image, I think it will use 16 bits instead of 8 (we need a unique ID for every axon). In this case, this conversion would be incompatible.

Since we considered revisiting #742, there would be 2 options. Remove this `uint16` to `uint8` conversion, or save the instance seg without this imwrite function.

hermancollin (commented):
Thank you for this improvement. Apart from this new `imwrite` conversion, everything lgtm

AxonDeepSeg/ads_utils.py
mathieuboudreau: Yeah I recalled #742, but didn't want to mix both of these issues up.

When I was going through this issue, I found that several of our images were not the expected bit depth, and annoyingly sometimes an image might get saved as 16bit using Photoshop but it really is just 8bit depending on the file type (eg png, it's possible to get a 16 bit B&W image saved but if converted to RGB, it will always be 8bit).

So I really just wanted to nail this issue solidely first, and then refactor as needed to resolve #747 and add the required tests at that time

mathieuboudreau: Also, if you check the history of the file, because also we've gone through so many iterations of imread of forcing it to 8 bit when read to then allowing flexible  bit depths to then rolling it back, etc.

These tests in this issue will kinda set the boundary of what's possible in terms of png or tiff and bit depth (as far as I could get them to work through manny iterations with saving using photoshop and/or gimp and then reading it in python using different tools)

mathieuboudreau: Also note that before #742 can be merged, we need to handle #668 and #698 (unless, were these done already inside #742? I'll have to check)

hermancollin: #742 solves #668 and #698 is somewhat unrelated and already implemented (the stanford model even outputs 5 classes)

hermancollin: > So I really just wanted to nail this issue solidely first, and then refactor as needed to resolve https://github.com/axondeepseg/axondeepseg/pull/742 and add the required tests at that time

makes sense!

hermancollin (approved):

PR #867, opened 2025-01-28
Set border morphometrics true by default

### LINKED ISSUE #863: Add border touching morphometrics information to GUI output
Is there a reason why we even have this as a flag, and not produced by default? I think it should just be that way personally, but happy to hear opposition if there's a reason why we didn't before that I don't recall

mathieuboudreau:
## Checklist

## Description

I also removed two debugging print statements from a previous PR that were accidentally left there.

## Linked issues

Resolves #863

### DIFF
--- a/AxonDeepSeg/ads_utils.py
+++ b/AxonDeepSeg/ads_utils.py
@@ -151,8 +151,6 @@ def imread(filename):
     img = imageio.core.image_as_uint(_img, bitdepth=8)
 
     # Remove singleton dimension
-    print(img.ndim)
-    print(img.shape)
     if img.ndim == 3 and img.shape[-1] == 1:
        img = img[:, :, 0]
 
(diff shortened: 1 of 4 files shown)


hermancollin:
So IIUC the `-b` option has no use anymore. Shouldn't we remove it?

mathieuboudreau:
>So IIUC the -b option has no use anymore. Shouldn't we remove it?

It's use is to turn it off instead of on, in case your concern for big images happens. I'm indifferent - have you ever encountered such a case?

hermancollin:
It can't be used to turn it off because it is currently in `store_true` mode.

https://github.com/axondeepseg/axondeepseg/blob/c946d8d7fc420b5515b5a859b892dac03cdb9141/AxonDeepSeg/morphometrics/launch_morphometrics_computation.py#L140-L146

mathieuboudreau:
> It can't be used to turn it off because it is currently in `store_true` mode.
> 
> https://github.com/axondeepseg/axondeepseg/blob/c946d8d7fc420b5515b5a859b892dac03cdb9141/AxonDeepSeg/morphometrics/launch_morphometrics_computation.py#L140-L146

Ahh I missed that part for the CLI, I changed the default only at the function level. If you want to keep the CLI option I can edit that one, otherwise as you said we can maybe just remove it altogether

hermancollin:
Sure let's remove it. The simpler, the better I guess.

hermancollin (approved):

Issue #513, opened 2021-04-24
Determine how to test the GUI


Stoyan-I-A:
Currently, AxonDeepSeg uses tests to validate if the installation is working and if the functions we wrote are behaving as expected.

However, we can't use the same tools to test the GUI (the FSLeyes plugin) because it requires user interaction and visual observations.

In order to test the GUI, one option would be to write a procedure that a collaborator would follow in order to verify if the features of the GUI are working correctly.

Stoyan-I-A:
@dyt811 In one of our meetings, Julien suggested we reach out to you for some insights on how we could test a GUI.

dyt811:
Hey, @Stoyan-I-A, if this is this web/browser based GUI, Selenium is pretty much the de facto tool to use. If it is an application like FSLEye, then I need to dig into what is currently out there.

jcohenadad:
@ahill187 has already looked into this for shimming-toolbox so i'm tagging her here

mathieuboudreau:
Not a high priority, but this could be revisited now that we're using Napari, there's better documentation/examples on how to do it: https://napari.org/dev/developers/contributing/testing.html

PR #868, opened 2025-01-28
Add click to delete "axon" objects and click to view morphometrics

### LINKED ISSUE #864: Add click to delete "axon" objects in Napari plugin
From [Stoyan's 20241219 meeting report](https://docs.google.com/presentation/d/11_oAFwHBA8p8DGUb9SP1yFSoIxv095-qOMgZuhRXhKs/edit?usp=sharing), here's a mockup of how to do it:

```python
import numpy as np
import imageio.v2 as iio
import napari
import napari.utils.notifications

viewer = napari.Viewer()
axon_image = iio.imread(r"C:\Users\Stoyan\Desktop\ADS\axondeepseg\AxonDeepSeg\models\model_seg_rat_axon-myelin_sem"
                        r"\data_test\image.png", mode="L")
image_layer = viewer.add_image(axon_image, name='axon_image')

@image_layer.mouse_drag_callbacks.append
def get_connected_component_shape(layer, event):
    data_coordinates = layer.world_to_data(event.position)
    cords = np.round(data_coordinates).astype(int)
    napari.utils.notifications.show_info(f"Clicked at {cords}")


if __name__ == '__main__':
    napari.run()


```

mathieuboudreau:
## Checklist

## Description

Current WIP working workflow
* Change to this branch
* Open napari
* Open the plugin
* Open an image
* Segment or load a mask
* Compute the morphometrics
* Select the initial image on the layer list
* Press and hold the command key (on Mac, on windows it might be the control key?) and then click on axon objects in the image to delete them in the axon and myelin masks.

## Linked issues

Resolves  #864 and #513


mathieuboudreau:
Some general thoughts:

* Do we want the option to delete axons to be available by default once proper conditions / layers are met, or do we want a button to "activate" it
* I binded the control key + click to do delete axons, 
* Do we want to add a log of which axons were deleted (likely yes)
* Unsure if possible (very likely tho), but do we want an "undo" feature?
* Do we want to bind a different key + click to maybe 1) show some morphometrics of that axon prior to deleting it and/or 2) to highlight the voxels that would be deleted if the proper binding / click happened?
* Several conditions must occur for click deletion to be possible: 1) there must be a mask loaded/generated, there must be an instance segmentation saved in the background, and the user must go back and select the raw image prior to clicking for deletion. It's a bit of a roundabout way, but I don't see any way around it, so maybe have a dedicated button to "activate" the feature could be nice as then we could check that all conditions are met and inform the user otherwise
* When axons are deleted, the indices for all the other axons aren't regenerated. Do we want that, or instruct the user to re-do the morphometrics button to generate this (likely).
* I'm not currently deleting the "number" of the axon when the axon is deleted - might want to add that.

I might have some more thoughts, and welcome @Stoyan-I-A for ideas on how to improve/clean this feature both in the back and front end.

Stoyan-I-A:
It looks very good so far from what I've seen in your video. I haven't looked at the code or tried it myself, I will do so in the upcoming days.
Here's my response to some of your thoughts:


> * Do we want the option to delete axons to be available by default once proper conditions / layers are met, or do we want a button to "activate" it

We would need the watershed to be computed at some point, which we do during the morphometrics but we could also do it separately. But yeah, I don't know what would be the most user-friendly approach. It might be easier for a user to understand the possibilities with a separate button (to which we could add a tooltip).

> * Do we want to add a log of which axons were deleted (likely yes)

If we can easily add it to the axondeepseg.log file, it would be nice. That said, you would have to retrive the index of the removed axon, which adds extra steps.

> * Unsure if possible (very likely tho), but do we want an "undo" feature?

I think it is BUT if you look at this part of the source code of napari ([link](https://github.com/napari/napari/blob/5432508b01e5dafe25889cf61f4eaffdc0e1b4c3/napari/layers/labels/labels.py#L990)), you will se that the undo/redo history is stored in private variables, so documentation for those will be limited and subject to change.

> * Do we want to bind a different key + click to maybe 1) show some morphometrics of that axon prior to deleting it and/or 2) to highlight the voxels that would be deleted if the proper binding / click happened?

This might be difficult to implement nicely with napari

> * When axons are deleted, the indices for all the other axons aren't regenerated. Do we want that, or instruct the user to re-do the morphometrics button to generate this (likely).
> * I'm not currently deleting the "number" of the axon when the axon is deleted - might want to add that.

If we choose to update the "numbers" layer after each modification, we would need to change the way it's generated because right now we just write the numbers on the image, but the individual indexes are not separate "objects" on the image

mathieuboudreau:
>Do we want the option to delete axons to be available by default once proper conditions / layers are met, or do we want a button to "activate" it
>We would need the watershed to be computed at some point, which we do during the morphometrics but we could also do it separately. But yeah, I don't know what would be the most user-friendly approach. It might be easier for a user to understand the possibilities with a separate button (to which we could add a tooltip).

I just made a button to toggle this feature on/off in 835cf0d

mathieuboudreau:
>Unsure if possible (very likely tho), but do we want an "undo" feature?
I think it is BUT if you look at this part of the source code of napari ([link](https://github.com/napari/napari/blob/5432508b01e5dafe25889cf61f4eaffdc0e1b4c3/napari/layers/labels/labels.py#L990)), you will se that the undo/redo history is stored in private variables, so documentation for those will be limited and subject to change.

So by default you can "undo" the axons that were removed via my "click to remove axon" feature, *but* you can only restore them for one layer at a time, so users need to go to the myelin layer and press control-z to restore those labels, and then switch to the axon and do the same to restore those labels for the same axon. Like you hinted, there could be a way to store the pixels in a private variable and make our own "undo" command via a different button / click combo, but then this would mess with the native napari history flow. I think keeping the native feature is fine for now, as long as we instruct the users on how to do it. If we hear enough annoyance or feel it ourselves, then we could explore introducing a custon undo feature that would do entire axons (axon+myelin) per undo action

mathieuboudreau:
>Do we want to bind a different key + click to maybe 1) show some morphometrics of that axon prior to deleting it and/or 2) to highlight the voxels that would be deleted if the proper binding / click happened?
This might be difficult to implement nicely with napari

I just mocked up this feature in the latest commit, once the feature is activated, ctrl-click on an axon pops up a show_info with some key morphometrics,

Screenshot 2025-01-29 at 3 47 48 PM

mathieuboudreau:
I think this feature might be at a point where it could welcome testers and feedback

mathieuboudreau:
The documentation preview can be accessed here: https://axondeepseg.readthedocs.io/en/mb-864/documentation.html#manual-correction-of-segmentation-masks

There's also pop up messages, information messages in the image viewer, and tooltips.

hermancollin:
napari was crashing, I had to reinstall to test the PR.

The axon removal feature is great. I tested it on a small image and it is very nice to use, this will definitely be a useful tool for fast manual corrections.

Only problem: with a large image, the feature is harder to use. IIUC, when the user clicks on the _axon removal toggle_ button, it launches the watershed segmentation computation. My prediction has 5974 axons, so clicking on the button freezes the software during the computation (I was not able to use the removal tool with this image). It would be a good idea to warn the user that this part can take some time. On this subject, I can see this in the console:

```
2025-01-31 10:10:33.382 | INFO     | AxonDeepSeg.visualization.colorization:colorize_instance_segmentation:100 - Colorizing 5974 instances.
```
Do we really need to colorize the instance seg? We are not using it (and it can be very slow).

mathieuboudreau:
> napari was crashing, I had to reinstall to test the PR.
> 
> The axon removal feature is great. I tested it on a small image and it is very nice to use, this will definitely be a useful tool for fast manual corrections.
> 
> Only problem: with a large image, the feature is harder to use. IIUC, when the user clicks on the _axon removal toggle_ button, it launches the watershed segmentation computation. My prediction has 5974 axons, so clicking on the button freezes the software during the computation (I was not able to use the removal tool with this image). It would be a good idea to warn the user that this part can take some time. On this subject, I can see this in the console:
> 
> ```
> 2025-01-31 10:10:33.382 | INFO     | AxonDeepSeg.visualization.colorization:colorize_instance_segmentation:100 - Colorizing 5974 instances.
> ```
> 
> Do we really need to colorize the instance seg? We are not using it (and it can be very slow).

Good catch and good point! Will look into an alternative implementation, and also add a test for this case

mathieuboudreau:
@hermancollin my latest push should have resolved the issue you raised re: refactor the use of colourized labels when identifying the axons in my feature

plugins/napari_ADS/_widget.py
```diff
@@ -402,6 +564,95 @@ def _on_load_mask_button_click(self):
...
+        """Handles the click event of the 'Toggle axon removal' button.
+
+        Switches the state of the remove_axon_state attribute, which is used to determine whether the user can click axons to remove
+
+        Returns:
+            None
+        """
+
+        if not self.image_loaded_after_plugin_start:
+            self.show_info_message("Please load an image first. If you loaded an image and are seeing this, you loaded the image prior to the plugin. Please remove and relopen the image and masks.")
```
Stoyan-I-A, line 577: Do you have to restrict to images loaded after the plugin has been loaded? Maybe we can add a check when the `Axon Removal` or `show axon morphometrics` features are toggled and add the callback at that moment

plugins/napari_ADS/_widget.py
```diff
@@ -402,6 +564,95 @@ def _on_load_mask_button_click(self):
...
+        if (axon_layer is None) or (myelin_layer is None):
+            self.show_info_message(f"To use this feature, the image layer must be selected and the myelin and axon masks must have been loaded or segmented via Apply ADS model.\nPlease load the masks or segment the image via Apply ADS model, and ensure that the image is selected as the active layer.")
+            # Uncheck the button
+            self.show_axon_metrics_state = False
+            self.show_axon_metrics_button.setChecked(False)
+            return
+        else:
+            if self.stats_dataframe is None:
+                self.show_info_message(f"Morphometrics for this image hasn't been computed yet - starting the Compute morphometrics process.")
+                self._on_compute_morphometrics_button()
```
Stoyan-I-A, line 646: There is an unhandled error in this scenario:
1. The morphometrics have not been computed yet, thus `_on_compute_morphometrics_button` is called.
2. No pixel size file is found, so the user is prompted to enter the pixel size value
3. The user changed his mind and presses "Cancel" on the pixel size dialog.
4. The `show_axon_metrics` button is still highlighted and the user can use ALT + click on an axon -> Error

plugins/test/test_widget.py
```diff
@@ -0,0 +1,61 @@
...
+from napari_ADS._widget import ADSplugin
+import numpy as np
+from unittest.mock import patch
+from pathlib import Path
+from AxonDeepSeg.ads_utils import imread
+class ImageLoadedEvent(object):
+    def __init__(self, data):
+        self.value = napari.layers.Image(data)
+
+class TestCore(object):
```
Stoyan-I-A, line 14: I like that you added some tests for the plugin, nice!

plugins/napari_ADS/_widget.py
```diff
@@ -181,6 +187,16 @@ def __init__(self, napari_viewer):
...
         fill_axons_button.clicked.connect(self._on_fill_axons_click)
 
+        remove_axons_button = QPushButton("Axon removal toggle")
+        remove_axons_button.clicked.connect(self._on_remove_axons_click)
+
+        remove_axons_button.setCheckable(True)
+        remove_axons_button.setStyleSheet(
+            "QPushButton:checked{background-color:blue;}"
+        )
+        remove_axons_button.setToolTip("Hold CONTROL/COMMAND and click on an axon to remove it in the axon and myelin masks.")
```
Stoyan-I-A, line 197: Maybe we should also mention in the tooltip that the image layer must be selected? I didn't read the changes in the documentation first and I was a bit confused when I tried this at first.

mathieuboudreau: I thought about that yeah, but was concerned with having too long of a message. I have a few other failsafes if the user doesn't do it themselves, but I'll add a short note here too thanks!

plugins/napari_ADS/_widget.py
mathieuboudreau: Ahh good catch, thanks! I'll handle it, and add a test for it too!

plugins/napari_ADS/_widget.py
mathieuboudreau: >Do you have to restrict to images loaded after the plugin has been loaded?

In my initial implementation, this is because our custom method `_on_layer_added` activates `layer.mouse_drag_callbacks.append(self._on_image_click)`, but now that I think of it, you're right that we can maybe add it to the buttons I added. I'll test it out!

plugins/napari_ADS/_widget.py
mathieuboudreau: Resolve in 34d4824

mathieuboudreau:
I've loosened the coverage % limits since this PR adds all of the GUI code in the coverage tests, but only adds tests targetting the features made in this PR. I'm going to open an issue to write tests for the rest of the GUI features, at which time I'll bring the coverage % limits back (per job, total, % change)

mathieuboudreau:
I added a new tests (axon removal button - click background pixel) and it fails. The reason being that when a background pixel is clicked, it removes all myelin that is in the image but isn't associated with an axon (which, why are we keeping in the first place?). These are just random bits of myelin pixels with no neighbouring axon pixels, so the morphometrics doesn't assign it an "axonmyelin" object.

I don't know why my test is failing in the code, or why it behaves this way. Even once I do, it's something I'll have to reflect on how to resolve in the feature/test

mathieuboudreau:
Actually, the solution in [67b5688](https://github.com/axondeepseg/axondeepseg/pull/868/commits/67b568896c347a2268f8e97a80e7dc8622ffb12a) was easy and quick

plugins/napari_ADS/_widget.py
mathieuboudreau: Resolved in [d9cc40a](https://github.com/axondeepseg/axondeepseg/pull/868/commits/d9cc40a045076ca680933bb1c81bd5b6b286927d)

plugins/napari_ADS/_widget.py
mathieuboudreau: I think I prefer to keep it this way for now - I like a single handler for when images are loaded, and we can setup / do checks there for all of our functions later. Maybe later when we tackle #871 we could have this handler initiated when the plugin is loaded (i.e. go through available layers that were loaded prior to the plugin and call this handler for each of them), but I'll have to write more extensive tests for that.

mathieuboudreau:
@Stoyan-I-A @hermancollin I've fixed a few more bugs and written all the tests I think I want to for now - let me know if there's any more changes to be made for this PR

hermancollin:
Now, when I try to remove an axon I get this error:
```python
AttributeError                            Traceback (most recent call last)
File ~/Documents/NEUROPOLY_21/axondeepseg/ads_conda/envs/venv_ads/lib/python3.11/site-packages/vispy/app/backends/_qt.py:496, in QtBaseCanvasBackend.mousePressEvent(self=<vispy.app.backends._qt.CanvasBackendDesktop object>, ev=<PyQt5.QtGui.QMouseEvent object>)
    494 if self._vispy_canvas is None:
    495     return
--> 496 self._vispy_mouse_press(
        self = <vispy.app.backends._qt.CanvasBackendDesktop object at 0x7f26ca7b3640>
        ev = <PyQt5.QtGui.QMouseEvent object at 0x7f25bddd4dc0>
        BUTTONMAP = {0: 0, 1: 1, 2: 2, 4: 3, 8: 4, 16: 5}
    497     native=ev,
    498     pos=_get_event_xy(ev),
    499     button=BUTTONMAP.get(ev.button(), 0),
    500     modifiers=self._modifiers(ev),
    501 )

[...]

File ~/Documents/NEUROPOLY_21/axondeepseg/ads_conda/envs/venv_ads/lib/python3.11/site-packages/napari_ADS/_widget.py:286, in ADSplugin._on_image_click(self=<napari_ADS._widget.ADSplugin object>, layer=<Image layer 'sub-rat1_sample-data1_SEM'>, event=<ReadOnlyWrapper at 0x7f25b4e34cc0 for MouseEvent>)
    283 cords = np.round(data_coordinates).astype(int)
    285 # Ensure the coordinates are within the bounds of the image
--> 286 if 0 <= cords[0] < self.im_instance_seg.shape[0] and 0 <= cords[1] < self.im_instance_seg.shape[1]:
        self = <napari_ADS._widget.ADSplugin object at 0x7f25be086ef0>
        cords = array([719, 451])
        self.im_instance_seg = None
    287     # Get the RGB value at the clicked position
    288     rgb_value = self.im_instance_seg[cords[0], cords[1]]
    290     # Get the indices for each region with the same RGB value

AttributeError: 'NoneType' object has no attribute 'shape'

```

mathieuboudreau:
Thanks @hermancollin  - i think this is related to one of my changes when making the tests last night. Odd that the tsat that evaluate this action doesnt trigger an error - i’ll try to reproduce your issue on my end. Sorry about that!

hermancollin:
@mathieuboudreau actually the problem was the plugin wasn't installed properly
. `pip install -e plugins/` did not work (I had the error I mentioned above), but when I re-install ADS from scratch it works. My bad.

I tested on a large image and it works now! It's not too slow either. To me everything looks good to go. Only thing left I see is related to this:
> Do we want to add a log of which axons were deleted (likely yes)

What was the conclusion for logging?

hermancollin:
In that case everything lgtm at this point. I'll let Stoyan give the final approval since he knows more about the plugin. Thank you for this great tool!

Stoyan-I-A:
I'm currently looking at this. I think we should update the version of the plugin in `plugins/setup.cfg`

mathieuboudreau:
> I'm currently looking at this. I think we should update the version of the plugin in `plugins/setup.cfg`

#873 may render obsolete, just a heads up

plugins/napari_ADS/_widget.py
```diff
@@ -183,6 +183,7 @@ def __init__(self, napari_viewer):
 
         load_mask_button = QPushButton("Load mask")
         load_mask_button.clicked.connect(self._on_load_mask_button_click)
+        self.load_mask_button = load_mask_button
```
Stoyan-I-A, line 186: This attribute doesn't seem to be used anywhere else in the plugin. Can we remove this line?

Stoyan-I-A (commented):
Just a small nitpick, the rest looks good to me

plugins/napari_ADS/_widget.py
mathieuboudreau: Good catch - fine with me! Don't recall why I added that

mathieuboudreau:
@Stoyan-I-A @hermancollin will need one of you to approve again now that I pushed that line deletion

mathieuboudreau:
@Stoyan-I-A ahhh, it's not used in the widget, but I needed it to access it through my tests haha.

```
=========================== short test summary info ============================
FAILED plugins/test/test_widget.py::TestCore::test_on_load_mask_button_click - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
FAILED plugins/test/test_widget.py::TestCore::test_on_remove_axons_click_no_morphometrics_computed - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
FAILED plugins/test/test_widget.py::TestCore::test_on_remove_axons_click_axon_pixel - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
FAILED plugins/test/test_widget.py::TestCore::test_on_remove_axons_click_myelin_pixel - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
FAILED plugins/test/test_widget.py::TestCore::test_remove_axon_undo - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
FAILED plugins/test/test_widget.py::TestCore::test_on_remove_axons_click_background_pixel - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
FAILED plugins/test/test_widget.py::TestCore::test_on_remove_axons_click_layer_changed_to_label_prior_click - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
FAILED plugins/test/test_widget.py::TestCore::test_on_show_axon_metrics_click_no_morphometrics_computed - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
FAILED plugins/test/test_widget.py::TestCore::test_on_show_axon_metrics_click_no_morphometrics_computed_user_cancels_pixel - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
FAILED plugins/test/test_widget.py::TestCore::test_on_show_axon_metrics_click_axon_pixel - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
FAILED plugins/test/test_widget.py::TestCore::test_on_show_axon_metrics_click_background_pixel - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
FAILED plugins/test/test_widget.py::TestCore::test_on_remove_axons_performance_large_image - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
FAILED plugins/test/test_widget.py::TestCore::test_on_show_axon_metrics_performance_large_image - AttributeError: 'ADSplugin' object has no attribute 'load_mask_button'
================== 13 failed, 5 passed, 11 warnings in 29.35s ==================
```

I'll reintroduce it; yeah to be able to test the widget, I needed to provide mode api access to some of the buttons

mathieuboudreau:
@Stoyan-I-A @hermancollin could you re-approve? The tests pass now

Stoyan-I-A (approved):

PR #873, opened 2025-02-06
pipy release / publish plugin in Napari

mathieuboudreau:
## Checklist


mathieuboudreau:
Changes in this PR:

* Remove git version logging; pygit2 can't be imported in the projects __init__ during install, as it runs prior to installing the packages.
* Switching from the deprecated setup.py+requirements.txt to pyproject.toml
* Several bug fixes
  * get_existing_models_list would not report None with no models were downloaded yet. Fixed.
  * Download models and download tests would not download to the correct locations if 1) no destination was provided and 2) the commands were run anywhere but axondeepseg/
  * Download tests and Download models can no longer be installed automatically during `pip install -e .` of the package (they do with `./install_ads` though), due to using pyproject.toml and because it's no longer recommended. Thus, there are now checks in integration test, the napari plugin, and segmentation API's to check if there are models downloaded, and if not download the default model. This doesn't impact users installing via github, but were necessary if installed via `pip` and via napari's plugin manager

mathieuboudreau:
One major change this PR is proposing is to merge the plugin project files into AxonDeepSeg/ . The reason for this is because to publish our plugin to Napari's napari-hub, we need to push a `pypi` package. If AxonDeepSeg and the plugin are seperate, this means we would need to publish two pypi projects: one for the plugin-only, and then another one for AxonDeepSeg that the plugin would have as a requirement. 

After some initial tests (https://pypi.org/project/ads-base/ and https://pypi.org/project/ads-napari/)., I rapidly found it annoying and cumbersome to handle both and update both,  and to have nearly duplicate things such as duplicate pyproject.toml for both projects and duplicate readme's. 

Merging both was quite straightforward; the widget now lives in AxonDeepSeg/ads_napari (https://github.com/axondeepseg/axondeepseg/tree/ads/pipy/AxonDeepSeg)/. There are no conflicting installation steps, and now we can simply publish to our old "axondeepseg" pypi package with `napari` in the frameworks list, and in napari our package would simply show up as AxonDeepSeg instead of something more cryptic like "ads_base", which doesn't inform the users really what our software does (also, maybe we should have called this AxonMyelinDeepSeg, but that's cumbersome too =) ).

This is only a proposal, so I'm open to discussions and feedback, particularly from @Stoyan-I-A but also @hermancollin and @joshuacwnewton.

mathieuboudreau:
Also, the base branch currently points to mb/864, as I built it off the latest plugin. Once that PR is merged, we can change the base branch for this PR to master

Stoyan-I-A:
Thank you @mathieuboudreau for looking into this. If the project gets listed in the napari-hub along with the other plugins, it will definitely be easier to install within napari and this will make the software more accessible overall.
More documentation might be required in order to guide users who would like to use the CLI commands, which I don't know how we would tackle. Would we still have installation instructions to install ADS outside of napari?
Another thing to keep in mind is that the PyPI package will need to be updated frequently as we will likely add bug fixes, especially early on.

mathieuboudreau:
>, but it's a little tricky at the moment given that there are some tangential changes in this PR, and the commits are a little experimental/messy ("Cleanup" etc.).

> Would it make sense to also do this for:
> 
> > * Switching from the deprecated setup.py+requirements.txt to pyproject.toml
> > * get_existing_models_list would not report None with no models were downloaded yet.
> > * Download models and download tests would not download to the correct locations if 1) no destination was provided and 2) the commands were run anywhere but axondeepseg/
> > * Download tests and Download models can no longer be installed automatically during pip install -e . of the package (they do with ./install_ads though), due to using pyproject.toml and because it's no longer recommended. Thus, there are now checks in integration test, the napari plugin, and segmentation API's to check if there are models downloaded, and if not download the default model. This doesn't impact users installing via github, but were necessary if installed via pip and via napari's plugin manager
> 
> Assuming that these changes would still be valuable even outside of a PyPI release?

Great point - I agree @joshuacwnewton . It'll also save me from this annoying Coverall's failing check in this PR I can't seem to reset. I'll open a new issue with all of these as sub-issues, and push PRs for each individual  one.

PR #876, opened 2025-02-07
Remove pygit2 & commit logging - breaks when ADS not downloaded as a git repo

### LINKED ISSUE #875: Installation error: "Repository not found" when downloading model files and test data for AxonDeepSeg
### Discussed in https://github.com/axondeepseg/axondeepseg/discussions/874

<div type='discussions-op-text'>

<sup>Originally posted by **liuyun016** February  7, 2025</sup>
Hello, 
while I was installing version 5.0.0 (I had previously installed Miniconda and Git, and followed the documentation on axondeepseg.readthedocs.io), I encountered the issue shown in the image below. The installation process had been normal up to that point. During the installation, I changed the installation directory and had my VPN turned on. I'm not sure if these actions are related to the problem. I would like to know what steps I should take to resolve this issue. Thank you for your assistance.
微信截图_20250208002744
</div>

mathieuboudreau:
## Checklist

## Description

Remove pygit2 & commit logging - breaks when ADS not downloaded as a git repo (i.e. when users download as zip from GitHub). See user discussion in #874 

## Linked issues

Resolves #875


mathieuboudreau:
@joshuacwnewton @hermancollin @Stoyan-I-A if one of you could approve quickly this bug fix, that would be greatly appreciated! It just removes all related to pygit2; a user downloaed ADS via *.zip (thus not a git repo) and thus breaks when trying to fetch the git info from the repo.

AxonDeepSeg/__init__.py
```diff
@@ -11,8 +10,3 @@
 # make sure repo directory (two directories up) is on path (to access config.py)
```
Stoyan-I-A, line 10: This comment is no longer accurate since we removed `config.py`

AxonDeepSeg/segment.py
```diff
@@ -260,8 +253,6 @@ def main(argv=None):
     with open("axondeepseg.log", "a") as f:
         f.write("===================================================================================\n")
 
-    logger.info(AxonDeepSeg.__version_string__)
```
Stoyan-I-A, line 263: Would it still be useful to have the version logged in another way? To be fair, I don't think so because we would always encourage users to use the master branch

Stoyan-I-A (approved):
I'm approving this if it can help a user. That said, the user in the discussion did mention that he used git clone. We'll see if this helps

AxonDeepSeg/__init__.py
```diff
@@ -11,8 +10,3 @@
...
 __repo_dir__ = os.path.abspath(os.path.join(__file__, os.pardir, os.pardir))
 sys.path.append(__repo_dir__)
-
-repo = pygit2.Repository(__repo_dir__)
-__git_version__ = str(repo.head.target)
-__git_name__ = repo.head.shorthand
-__version_string__ = f"{__name__} v{__version__} ({__git_name__}: {__git_version__})"
```
joshuacwnewton, line 18: Couldn't we just `try: except:` here if the call fails? Something like:

```python
# handle case where ADS isn't downloaded via git
# see: https://github.com/axondeepseg/axondeepseg/issues/875
try:
    repo = pygit2.Repository(__repo_dir__)
    __git_version__ = str(repo.head.target)
    __git_name__ = repo.head.shorthand
    git_string = " ({__git_name__}: {__git_version__})"
except pygit2.GitError:  # https://www.pygit2.org/general.html#pygit2.GitError
    git_string = ""
__version_string__ = f"{__name__} v{__version__}{git_string}"
```

mathieuboudreau: There is a secondary issue, which is with the new installation proposed in #873 , it throws an import error even before these lines, as the `pyproject.toml` way of installation load the project init prior to installing the dependencies. I think I just need to start from scratch and find an alternative solution to logging the version that doesn't depend on using pygit2 in __init__.py

AxonDeepSeg/segment.py
mathieuboudreau: Yes I agree - I think we should open a new issue to handle this and find an alternative solution, I'll sdo that now

AxonDeepSeg/__init__.py
joshuacwnewton: Ahhhhhh, woof!! Thanks for clarifying. That's quite a pain. :sweat:

Issue #881, opened 2025-02-10
Switching from the deprecated setup.py+requirements.txt to pyproject.toml


mathieuboudreau:
See #873

Issue #872, opened 2025-02-04
Publish AxonDeepSeg on Napari-Hub


mathieuboudreau:
I've revisited the suggestion made by @psobolewskiPhD in #808 re: restructuring the code/install such that ADS could be published on [napari-hub](https://www.napari-hub.org), which would require pipy releases of ADS and our plugin. A few major benefits of this would include greater exposure to potential users, better discoverability, and easier installation

After tinkering around in a branch ([ads/pipy](https://github.com/axondeepseg/axondeepseg/tree/ads/pipy)), I'm fairly certain we can do this. Well. because I just did it.

<img width="946" alt="Image" src="https://github.com/user-attachments/assets/1616e3b1-7699-4e7f-9933-99a45e0540e1" />

https://pypi.org/project/ads-base/
https://pypi.org/project/ads-napari/

Here's me installing both AxonDeepSeg and the Napari plugin directly from Napari (meaning, not using our currently documented `./install_ads` installation), using the napari-hub search:

https://github.com/user-attachments/assets/a961fefd-fabf-484d-8650-5782dee4e013

I found some bugs while working towards this, and I think some structural decisions may need to be discussed as there's a few ways we can do this. But, I think it's realistic in the short term to get a version out.

PR #883, opened 2025-02-10
Make get_existing_models_list returns None when no models exist

### LINKED ISSUE #878: get_existing_models_list would not report None with no models were downloaded yet.
See #873

mathieuboudreau:
## Checklist

## Description

This PR exists to solve a problem that isn't here yet, i.e. I only discovered it after trying the pypi route. Basically, currently the installation procedure is guaranteed to always download the models otherwise the installation fails, due to some embedded checks (running the unit tests) and explicit downloading of the data.

When we later switch to offer a pipy install, this won't be the case. And, due to the switch from setup.py to pyproject.toml, there's not really a way to do it either (and it's mainly discouraged to try and get around it); usually they recommend to tell the user to run some kind of axondeepseg init to do all that data-related stuff.

## Linked issues

Resolves #878

### DIFF
--- a/AxonDeepSeg/ads_utils.py
+++ b/AxonDeepSeg/ads_utils.py
@@ -192,6 +192,11 @@ def get_existing_models_list():
     """
     ADS_path = os.path.dirname(os.path.realpath(__file__))
     models_path = os.path.join(ADS_path, "models")
+
+    # If models folder doesn't exist or it's empty, return None
+    if not os.path.exists(models_path) or not os.listdir(models_path):
+        return None
+
     models_list = next(os.walk(models_path))[1]
     if "__pycache__" in models_list:
         models_list.remove("__pycache__")
--- a/test/test_ads_utils.py
+++ b/test/test_ads_utils.py
@@ -4,9 +4,11 @@ from pathlib import Path
 import shutil
 import numpy as np
 import imageio
+import os
 
 import pytest
 
+import AxonDeepSeg
 from AxonDeepSeg.ads_utils import download_data, convert_path, get_existing_models_list, extract_axon_and_myelin_masks_from_image_data, imread, imwrite, get_file_extension, check_available_gpus
 from AxonDeepSeg.model_cards import get_supported_models
 from AxonDeepSeg import params
@@ -20,7 +22,11 @@ class TestCore(object):
 
         self.precision_path = Path('test/__test_files__/__test_precision_files__')
 
-        self.tmp_folder = Path('test/__test_files__/tmp_folder')
+        # Get models folder
+        self.package_dir = Path(AxonDeepSeg.__file__).parent  # Get AxonDeepSeg installation path
+        self.model_dir = self.package_dir / "models"
+
+        self.tmp_folder =  self.package_dir.parent /'test/__test_files__/tmp_folder'
 
         if not self.tmp_folder.exists():
             self.tmp_folder.mkdir()
@@ -30,6 +36,12 @@ class TestCore(object):
         if output_path.exists():
             shutil.rmtree(output_path)
         
+        # If tmp/models folder isn't empty, move it back
+        if (self.tmp_folder / "models").exists():
+            for file in (self.tmp_folder / "models").iterdir():
+                shutil.move(file, self.model_dir)
+                shutil.rmtree(self.tmp_folder / "models")
+
         if self.tmp_folder.exists():
             shutil.rmtree(self.tmp_folder)
 
@@ -101,6 +113,26 @@ class TestCore(object):
 
         for downloaded_model in get_existing_models_list():
             assert downloaded_model in known_models
+
+    @pytest.mark.unit
+    def test_get_existing_models_list_returns_none_for_no_existing_models(self):
+        # Move content inside model folder to test/__test_files__/tmp_folder, without moving the folder itself
+        if not (self.tmp_folder / "models").exists():
+            (self.tmp_folder / "models").mkdir()
+
+        for file in self.model_dir.iterdir():
+            print(file)
+            # Make temp dir
+            shutil.move(file, self.tmp_folder / "models")
+        
+        # Main test
+        assert get_existing_models_list() == None
+
+        # Could leave it to teardown, but might as well try now in case other tests in this file use the model
+        if (self.tmp_folder / "models").exists:
+            for file in (self.tmp_folder / "models").iterdir():
+                shutil.move(file, self.model_dir)
+                shutil.rmtree(self.tmp_folder / "models")
 
     # --------------imread tests-------------- #
     @pytest.mark.unit
--- a/test/test_download_tests.py
+++ b/test/test_download_tests.py
@@ -58,7 +58,7 @@ class TestCore(object):
 
         download_tests(self.tmpPath)
 
-    @pytest.mark.single
+    @pytest.mark.unit
     def test_precision_images_expected_types(self):
 
         precision_path = Path('test/__test_files__/__test_precision_files__')


AxonDeepSeg/ads_utils.py
```diff
@@ -192,6 +192,11 @@ def get_existing_models_list():
     """
     ADS_path = os.path.dirname(os.path.realpath(__file__))
     models_path = os.path.join(ADS_path, "models")
+
+    # If models folder doesn't exist or it's empty, return None
+    if not os.path.exists(models_path) or not os.listdir(models_path):
+        return None
+
     models_list = next(os.walk(models_path))[1]
```
joshuacwnewton, line 200: Just to make sure I understand the previous error -- if the folder didn't exist, we'd get a `StopIteration` error:

```python
>>> next(os.walk("blahblah"))[1]
  File "/opt/pycharm-2024.2.3/plugins/python-ce/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
StopIteration
```

And if the folder exists but is empty, we'd get an empty list?

```python
>>> next(os.walk("blahblah"))
('blahblah', [], [])
>>> next(os.walk("blahblah"))[1]
[]
```

Looking in the repo, it seems like this function is mainly used in the context of the [Napari Plugin](https://github.com/search?q=repo%3Aaxondeepseg%2Faxondeepseg%20get_existing_models_list&type=code) for model selection:

https://github.com/axondeepseg/axondeepseg/blob/ccd48b1e9d4738c9f30fe5a1a93e4f283460a3b8/plugins/napari_ADS/_widget.py#L163-L167

Wouldn't returning `None` also result in an error, though? 

```python
>>> ["Select the model"] + None
  File "/opt/pycharm-2024.2.3/plugins/python-ce/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
TypeError: can only concatenate list (not "NoneType") to list
```

So maybe we want to return an empty list instead? 

```python
try:
   models_list = next(os.walk(models_path))[1]
except StopIteration:  # next(os.walk) failed i.e. `models_path` doesn't exist
   models_list = []

File: test/test_download_tests.py [hunk 2]
```diff
@@ -58,7 +58,7 @@ def test_redownload_test_files_multiple_times_works(self):
 
         download_tests(self.tmpPath)
 
-    @pytest.mark.single
+    @pytest.mark.unit
```
joshuacwnewton, line 61: Just to make sure I understand, this isn't actually related to this PR, right? (The convention in ADS is to use the term `unit`, but it looks like this was missed from some previous PR or something like that, right?)

test/test_ads_utils.py
```diff
@@ -102,6 +114,26 @@ def test_get_existing_models_list_returns_known_models(self):
...
+        if not (self.tmp_folder / "models").exists():
+            (self.tmp_folder / "models").mkdir()
+
+        for file in self.model_dir.iterdir():
+            print(file)
+            # Make temp dir
+            shutil.move(file, self.tmp_folder / "models")
+        
+        # Main test
+        assert get_existing_models_list() == None
```
joshuacwnewton, line 129: Assuming the suggested change is made, then maybe this test would be:

```python
models_list = get_existing_models_list()
assert isinstance(models_list, list) and len(models_list) == 0
```

joshuacwnewton (commented):
Thanks for splitting off this change! Apologies for the extra hoops to jump through, but it makes things much easier on this end. :hearts:

AxonDeepSeg/ads_utils.py
mathieuboudreau: >Just to make sure I understand the previous error -- if the folder didn't exist, we'd get a StopIteration error:

>>> next(os.walk("blahblah"))[1]
  File "/opt/pycharm-2024.2.3/plugins/python-ce/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
StopIteration

I don't recall seeing this error message explicitely, but maybe. But yes, this line is where it will error.

This PR exists to solve a problem that isn't here yet, i.e. I only discovered it after trying the pypi route. Basically, currently the installation procedure is guaranteed to always download the models otherwise the installation fails, due to some embedded checks (running the unit tests) and explicit downloading of the data.

When we later switch to offer a pipy install, this won't be the case. And, due to the switch from setup.py to pyproject.toml, there's not really a way to do it either (and it's mainly discouraged to try and get around it); usually they recommend to tell the user to run some kind of `axondeepseg init` to do all that data-related stuff.

So for this case, no models are downloaded, and  in fact, the model folder itself doesn't even exist. So the first time running either the plugin or the segmentations, when the function is called to check the available models,

https://github.com/axondeepseg/axondeepseg/blob/ccd48b1e9d4738c9f30fe5a1a93e4f283460a3b8/AxonDeepSeg/ads_utils.py#L195C4-L195C48

will error because you can't walk a list of files that are in a folder that doesn't exist

test/test_download_tests.py
mathieuboudreau: Yep - thought I'd squeeze this in here since I noticed it

test/test_ads_utils.py
mathieuboudreau: I Iike it - will make the change and push!

AxonDeepSeg/ads_utils.py
mathieuboudreau: >Looking in the repo, it seems like this function is mainly used in the context of the [Napari Plugin](https://github.com/search?q=repo%3Aaxondeepseg%2Faxondeepseg%20get_existing_models_list&type=code) for model selection:

Currently yes, but (this is where splitting up a PR gets messy), I was using it to solve some upcoming issues as well.

To me I see two cases: one where the `models/` folder exists but is empty; to me, returning an empty list here makes sense. But, there is a second (more likely) case, which is that the `models/` folder itself doesn't exist, and to me I feel better returning None vs an empty list for that case (since it indicates that most likely the data download hasn't been initiated since installation).

mathieuboudreau: >Wouldn't returning None also result in an error, though?

>>> ["Select the model"] + None
  File "/opt/pycharm-2024.2.3/plugins/python-ce/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
TypeError: can only concatenate list (not "NoneType") to list

Again, this is a victim of splitting up the PR =)

In my original PR, I added checks prior to these lines such that if no model is there (indicating a pipy or napari install), then to run the download (+ show the user a message). i.e. these lines here

https://github.com/axondeepseg/axondeepseg/pull/873/files#diff-022464199d6b3724f2508950c2156aa6c1d2f07364b4ef30bd9a22382b87cd22R170-R179

But I didn't include them in *this* PR, as this isn't an expected issue for our current installation (which fails if models aren't downloaded succesfully); this is only an issue following when I'll resolve  #881 and/pr #872.

joshuacwnewton: Ahhhhhhh, okay. I'm sorry about missing the context, but thank you for explaining things. LGTM, then. :)

joshuacwnewton (approved):

mathieuboudreau:
Thanks @joshuacwnewton !

>coverage/coverallsExpected — Waiting for status to be reported

Hum - that's odd, there doesn't seem to have been any issue in sending the report to Coveralls: https://github.com/axondeepseg/axondeepseg/actions/runs/13249122543/job/36983321066

joshuacwnewton:
Really odd! That run link points to https://coveralls.io/builds/72172426, which just says "pending completion". Maybe something funky happened with the API call? Weird...

joshuacwnewton:
Ahh... perhaps #876 caused this? I notice in the commit log that the successful runs dropped from 6/6 to 4/4 after that was merged.

image

PR #884, opened 2025-02-11
Log current commit hash

### LINKED ISSUE #877: Log git commit
We need to find an alternative way to what was used before (i.e. pygit2 + __init__.py), as we removed our initial attempt in #876

hermancollin:
## Checklist

## Description
In #876, we removed the logging of the commit hash because the pygit2 dependency caused problems (note that I also noticed this issue on my end - after a fresh install of ADS, I had to manually pip install pygit2 to segment images). This PR re-introduces this feature without using any external package. I also took this opportunity to put the same logging for the morphometrics script.

@Stoyan-I-A  said this about this feature: https://github.com/axondeepseg/axondeepseg/pull/876#discussion_r1948164837
> Would it still be useful to have the version logged in another way? To be fair, I don't think so because we would always encourage users to use the master branch

I think it makes sense to keep the commit hash, because even if we encourage users to use the master branch, realistically they never update the software without being asked to :face_with_diagonal_mouth: 

Also, you'll see that if the repo was not git cloned, I simply put _"unknown commit"_ in the version string. Not much we can do in that situation.

## Linked issues
Resolves #877


AxonDeepSeg/__init__.py
```diff
@@ -10,3 +10,11 @@
 # make sure repo directory (two directories up) is on path (to access config.py)
 __repo_dir__ = os.path.abspath(os.path.join(__file__, os.pardir, os.pardir))
 sys.path.append(__repo_dir__)
+
+# verify that the repo was downloaded with `git clone`
+if os.path.exists(os.path.join(__repo_dir__, ".git")):
+    import subprocess
+    __git_sha1__ = subprocess.check_output(["git", "-C", __repo_dir__, "rev-parse", "HEAD"]).decode("utf-8").rstrip()
```
mathieuboudreau, line 17: What happens in the case that the user doesn't have git installed?

mathieuboudreau (approved):
LGTM!

PR #886, opened 2025-02-13
Transition installation from deprecated setup.py+requirements.txt to pyproject.toml

### LINKED ISSUE #881: Switching from the deprecated setup.py+requirements.txt to pyproject.toml

mathieuboudreau:
## Checklist

## Description

Moving towards doing pipy releases for publishing our software on Napari's plugin hub, it's recommended to use the more modern pyproject.toml and some additional files instead of setup.py+requirements.txt, which is what this PR does. No changes from the user's side are needed during installation.

https://www.bitecode.dev/p/whats-the-deal-with-setuptools-setuppy
https://packaging.python.org/en/latest/guides/modernize-setup-py-project/

## Linked issues

Resolves #881

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,11 +45,15 @@ dependencies =[
     "jupyter",
     "openpyxl",
     "qtconsole<5.4.2",
+    "napari[all]",
     "acvl_utils!=0.2.1",
     "nnunetv2==2.2.1",
     "loguru",
     "torch<2.4.0",
     "pydicom<3",
+    "pytest-qt",
+    "magicgui",
+    "qtpy"
 ]
 
 keywords = ["axon", "myelin", "segmentation", "microscopy"]


pyproject.toml
```diff
@@ -0,0 +1,87 @@
...
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Topic :: Scientific/Engineering :: Image Processing",
+]
+
+dependencies =[
+    "numpy<2",
+    "scipy",
+    "scikit-image",
```
joshuacwnewton, line 33: Diffing the [list of dependencies](https://www.diffchecker.com/cTaOYkdt/), I notice that there are some changes:

 - `napari` -> `napari[all]`
 - added `pytest-qt`, `magicgui`, and `qtpy`

It looks like some of these are specific to the Napari plugin (from `plugins/setup.cfg`). Would it make sense to remove the new dependencies from this PR, and then re-add them in the Napari-specific PR? (Just to show which dependencies got added from merging Napari into ADS?)

Of course this is like, pretty nitpicky and not 100% necessary, but it would create a  nice record of how the transition evolved between PRs.

install_ads
```diff
@@ -635,26 +635,12 @@ fi
 
 ## Install the axondeepseg into the Conda venv
 print info "Installing Python dependencies..."
-# Check if a frozen version of the requirements exist (for release only)
-if [[ -f "requirements-freeze.txt" ]]; then
-  print info "Using requirements-freeze.txt (release installation)"
-  REQUIREMENTS_FILE="requirements-freeze.txt"
-else
```
joshuacwnewton, line 642: I notice that the `requirements-freeze.txt` logic is no longer possible, since the requirements are baked into the `pyproject.toml` file now. (I don't think ADS ever used this logic in the first place, so perhaps this isn't actually relevant for this PR?)

<details>
<summary> Off-topic discussion better suited for a follow-up issue </summary>

But, say that SCT wishes to transition to `pyproject.toml`. Is there any way to keep the behavior of "conditionally overriding the list of dependencies"? 

I ask because there is a risk of some dependency or another updating in the future, and it breaking older versions of ADS. (This is especially relevant for SCT where older versions continue to be used in present-day papers.) Is this sort of reproducibility a concern for ADS?

</details>

pyproject.toml
```diff
@@ -0,0 +1,87 @@
...
+axondeepseg_test = "AxonDeepSeg.integrity_test:main"
+axondeepseg_morphometrics = "AxonDeepSeg.morphometrics.launch_morphometrics_computation:main"
+
+[tool.setuptools]
+packages = ["AxonDeepSeg"]
+package-dir = {AxonDeepSeg = "AxonDeepSeg"}
+include-package-data = true
+
+[tool.setuptools.package-data]
+"*" = ["*.yaml"]
```
joshuacwnewton, line 84: This bit here is a little peculiar to me. Does it recreate the old `setup.py` behavior below?

```python
    package_data={
        "AxonDeepSeg": ['models/default_SEM_model/*',
                        'models/default_TEM_model/*',
                        'models/default_BF_model/*'
                        'data_test/*'],
    },
```

install_ads.bat
```diff
@@ -171,17 +171,11 @@ ads_conda\Scripts\conda create -y -p ads_conda\envs\venv_ads python=3.11 || goto
...
-  set requirements_file=requirements.txt || goto error
-)
 echo:
-echo ### Installing ADS and its dependencies from %requirements_file%...
+echo ### Installing ADS and its dependencies
 rem Skip pip==21.2 to avoid dependency resolver issue (https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3593)
 ads_conda\envs\venv_ads\python -m pip install -U "pip^!=21.2.*" || goto error
-ads_conda\envs\venv_ads\Scripts\pip install -r %requirements_file% || goto error
+ads_conda\envs\venv_ads\Scripts\pip install -e . || goto error
 ads_conda\envs\venv_ads\Scripts\pip install -e . plugins\ --use-pep517 || goto error
```
joshuacwnewton, line 179: This step is duplicated, I think?

tox.ini
```diff
@@ -0,0 +1,32 @@
+# For more information about tox, see https://tox.readthedocs.io/en/latest/
+[tox]
+envlist = py{311,312,313}-{linux,macos,windows}
+isolated_build=true
```
joshuacwnewton, line 4: Are we currently using this `tox.ini` file at all? Even checking the previous "mega" PR, I don't see any changes to the GitHub Actions workflow files related to tox. 

And, searching `tox` in the repo doesn't yield anything either... :thinking:

joshuacwnewton (approved):
The majority of this PR looks good to me, so I'm pre-approving. Most of my comments are either tiny nitpicks or questions for my own understanding. :)

pyproject.toml
mathieuboudreau: Good catch - yeah these are here too early =) Will remove

install_ads
mathieuboudreau: >(I don't think ADS ever used this logic in the first place, so perhaps this isn't actually relevant for this PR?)

Correct, we've not done this.

Despite that, I was able to install and run all the prior major versions of ADS (going back to 2018) with fair success, see

* Repo: https://github.com/axondeepseg/seg_ci
* Comparison doc for versions 1-5: https://docs.google.com/presentation/d/1JptrOQUYWxHI-2fHTRsATfLue8bP6dRcKQFe-PU8WK0/edit?usp=sharing

It's possible that if I were to use the *exact* same versions of all the pypi packages that the results would have differed, but I think this difference would have been quite small. On another note, having tried to fix all the version of softwares for containers (Dockerfiles), I've experienced that most of the time the container will fail to rebuild only one or two years later because of changes on the linux backend (i.e. often apt-gets would disapear or change, with trickle down effects that would break my python env setup).

Most likely I would recommend that *users* find the versions that were installed at their time of use for whatever study (eg pip freeze), so that they can report those and/or save in case of needing to re-run analyses (but ideally, they'd really just create a docker image and reuse that)

install_ads.bat
mathieuboudreau: Good catch - I didn't notice the `.`. in the plugins install line below, I thought that line was just for the plugin. I'll remove this line (178)

tox.ini
mathieuboudreau: This is another of the "recommended" files from napari for napari-hub. I thought maybe they need it on their end, but maybe not? If you want I can remove it for now

mathieuboudreau:
> The majority of this PR looks good to me, so I'm pre-approving. Most of my comments are either tiny nitpicks or questions for my own understanding. :)

Thanks! I've answered most, will do some of the suggested changes!

pyproject.toml
joshuacwnewton: Ahhhh, gotcha!

So, how is `package_data` preserved? Or, is it even needed? (I don't see `models/` or `data_test/` in the repo, so I'm curious if this is a really old/outdated part of ADS' `setup.py`.)

tox.ini
joshuacwnewton: I think we'd only need the file if we're using the `tox` tool to run tests. It could be something worth experimenting with, but if we have no plans to adopt it, then yeah, I don't think it's needed! :)

joshuacwnewton: Just for posterity, though, could you link the guide you were following?

mathieuboudreau: I think it was https://github.com/napari/napari-plugin-template/?tab=readme-ov-file#napari-plugin-template and/or https://napari.org/stable/plugins/testing_and_publishing/deploy.html

mathieuboudreau: > I think we'd only need the file if we're using the `tox` tool to run tests. It could be something worth experimenting with, but if we have no plans to adopt it, then yeah, I don't think it's needed! :)

So I'd have to double check, but I think napari-hub wants to execute our tests at some point or regularly (they specifically ask to have written tests for the plugin), so maybe that's why they want the tox? But yeah, we're not using it internally yet, I'll explore it!

joshuacwnewton: Ahhh, I see now why I had trouble finding the use of `tox`. In the `napari-plugin-template` repo, they currently have it set up the following way:

- For their GitHub Actions workflow [`test.yml`](https://github.com/napari/napari-plugin-template/blob/main/.github/workflows/test.yml), they run the command `python -m pytest -s -v --color=yes`
- Then, in their Python testing script, they call [`tox` via `subprocess`](https://github.com/napari/napari-plugin-template/blob/95d5eceba554115b3e7eda3253acf486f052cd1d/tests/test_create_template.py#L13-L20).

This isn't what I'm used to? (I expected to the workflow file to run the `tox` CLI command directly.) But, maybe Napari has a specific reason for this (e.g. their use of [`headless-gui`](https://github.com/napari/napari-plugin-template/blob/95d5eceba554115b3e7eda3253acf486f052cd1d/.github/workflows/test.yml#L39) testing).

mathieuboudreau: Yep - I didn't dig much deeper, I think they need it so I made it =)

If it's ok I'll resolve this point!

install_ads.bat
mathieuboudreau: Resolved in 3762ae3

pyproject.toml
mathieuboudreau: > package_data

I believe it's needed to include the yaml files that napari will need when it installs our ADS plugin via pypi

install_ads
mathieuboudreau: Yeah - I guess practically, the need has never come up. Maybe another major difference between SCT and ADS is I feel like the backend of SCT is more "stable" than ADS - when we change major versions, the entire backend is often overhauled (from tensorflow (v1, v2) to pytorch (v3) to removing all training code and using ivadomed instead (v4) to now reintroducing *some* of it and using nnunet (v5). We'd never really recommend users to go back to older versions unless it was to reproduce a study for whatever reason (and in that case, there are other issues like the OS, hardware, etc that can and will change some of the underlying calculations/rounding even if the versions are identical). Also, we've found in some cases across versions that different OSes also required additional dependencies installed (which due to their own requirements, could impact other versions that are installed).

Maybe one compromise would be to include a `pip freeze` output as an supplementary file in releases, and do this for each of the three OS (so maybe have it's own pipeline in github actions that would then upload the pip freeze output as an artefact). That way we have that information stored somewhere, and it reduces duplication in the code (so if someone downloads a major release, they're not confused as to which of the two requirements files they should really use)

pyproject.toml
mathieuboudreau: Resolved in 096eac0

joshuacwnewton: It seems that removing `pytest-qt` caused the tests to fail:

```
file /Users/runner/work/axondeepseg/axondeepseg/ads_conda/envs/venv_ads/lib/python3.11/site-packages/napari/utils/_testsupport.py, line 178
  @pytest.fixture
  def make_napari_viewer(
E       fixture 'qtbot' not found
```

I'm surprised why this doesn't fail on the main branch?

joshuacwnewton: Sorry for all the hassle. Feel free to preserve these dependencies then merge as-is. I don't want to hold up the transition any longer than it needs to be due to organizational nitpicks. :sweat:

mathieuboudreau: Probably due to this line that's still in there that I had also removed,

[pytest-qt](https://github.com/axondeepseg/axondeepseg/blob/2334be33a5e4c3bd2f3bdbf459d710de7b067d60/requirements.txt#L25C1-L25C10)

I think you had maybe missed it when comparing, and I removed it since you mentionned it but didn't compare myself. So looks like that's the only very necessary one, but I'll keep the others since they're coming in the next PR anyways

PR #887, opened 2025-02-13
Add checks to run download models if API called without having been downloaded during installation

### LINKED ISSUE #880: Add checks to run download models if API called without having been downloaded during installation
See #873

mathieuboudreau:
## Checklist

## Description

This is not an issue right now for our installation via `./install_ads`, as we explicitely download the models and tests there as a requirement. However, when we'll start offering installation via pipy and napari-hub, we won't be able to (and it's not recommended) download data during the installation (in large part due to #886). Thus, the first time the user encounters either the napari plugin or CLI segment feature, this PR ensures that there is a mechanism to download the necessary models at that time + inform the user.

## Linked issues

Resolves #880


AxonDeepSeg/segment.py
```diff
@@ -143,6 +143,24 @@ def segment_images(
 
     path_images = [convert_path(p) for p in path_images]
     path_model = convert_path(path_model)
+
+    available_models = ads.get_existing_models_list()
+    if available_models is None:
+        available_models = []
+    if path_model.stem not in available_models:
```
joshuacwnewton, line 150: Minor nitpick to preserve the distinction between `None` and `[]` (as you explained to me earlier)

```suggestion
    if not available_models or path_model.stem not in available_models:
```

AxonDeepSeg/segment.py
```diff
@@ -143,6 +143,24 @@ def segment_images(
...
+                # Call download models from the AxonDeepSeg/download_model.py module
+
+                import AxonDeepSeg.download_model as download_model
+                download_model.download_model()
+
+                available_models = ads.get_existing_models_list()
+                assert path_model.stem in available_models
+            except:
+                print('Could not download models, try again.')
+                quit()
```
joshuacwnewton, line 162: Might be good to report the full exception message to help when debugging issues with users?

```suggestion
            except Exception as e:
                raise Exception('Could not download models, try again.') from e
```

Result would look something like this:

```
  File "segment.py", line 159, in <module>
    assert path_model.stem in available_models
AssertionError: path_model.stem not in available models

  File "segment.py", line 4, in <module>
    raise Exception('Could not download models, try again.') from e
Exception: Could not download models, try again.
```

If this is too complicated, then there are [cleaner options](https://stackoverflow.com/a/75549200), but I would also understand just keeping things simple and only printing a simple message.

joshuacwnewton: I'm also curious about the use of `quit()` here, given that usually it's [not recommended](https://www.geeksforgeeks.org/python-exit-commands-quit-exit-sys-exit-and-os-_exit/).

Is it necessary for Napari, maybe?

plugins/napari_ADS/_widget.py
```diff
@@ -161,6 +161,16 @@ def __init__(self, napari_viewer):
         )
 
         self.available_models = ads_utils.get_existing_models_list()
+
+        if self.available_models == None:
+            if self.show_ok_cancel_message('First time opening pluging - models must be downloaded'):
```
joshuacwnewton, line 166: Pluging -> plugin?

joshuacwnewton: Also, maybe the following description better illuminates the issue. Since, technically the model folder could get deleted, but it wouldn't be the first time opening the plugin, so the user might be confused...

```suggestion
            if self.show_ok_cancel_message('Model folder is missing. (Hint: Is this the first time opening the plugin?)\n\nPress 'OK' to download the required models.'):

```

AxonDeepSeg/segment.py
```diff
@@ -143,6 +143,24 @@ def segment_images(
...
+    available_models = ads.get_existing_models_list()
+    if available_models is None:
+        available_models = []
+    if path_model.stem not in available_models:
+            try:
+                print('Model not found, attempting to download')
+                # Call download models from the AxonDeepSeg/download_model.py module
+
+                import AxonDeepSeg.download_model as download_model
+                download_model.download_model()
```
joshuacwnewton, line 156: Tiny tiny tiny nitpick, but why do we call `download_model()` here but `main()` in the plugin?

They seem to have the same functionality (default values of 'generalist', 'light', and './models/{model_name}'). But, there's some duplication here, so I wonder if there's a risk of a minor code change causing the two different methods to become subtly different. (e.g. https://github.com/axondeepseg/axondeepseg/pull/836)

Could it be beneficial to remove the default arguments of `download_model`, and only store the defaults in `main()`, then standardize on `main()` becoming the primary method of calling this function?

It's fine to save this for a follow-up issue of course! It's not entirely relevant for this PR... :sweat_smile:

joshuacwnewton (changes requested):
It mostly looks good! I just have some nitpicks, hehe...

mathieuboudreau:
> Thanks for addressing my comments! This LGTM! 🎉

```
    from ._widget import ADSplugin
E     File "/Users/runner/work/axondeepseg/axondeepseg/plugins/napari_ADS/_widget.py", line 172
E       if self.show_ok_cancel_message('Model folder is missing. (Hint: Is this the first time opening the plugin?)\n\nPress 'OK' to download the required models.'):
E                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   SyntaxError: invalid syntax. Perhaps you forgot a comma?
```
The tests fail here - I think maybe we should be using a double `"` at the start and end of the message instead of a single `'`, since this currently would close the string just before `OK`?

joshuacwnewton:
Whoops! Good catch. Sorry about that.

Issue #849, opened 2024-12-17
Why are some pickled morphometric values changing with scikit 0.25?


mathieuboudreau:
This is a continuation of this PR that pins scikitimage to 0.25.0. Here's the initial discussion,

____________________________

Four days ago, our nightly GithubAction CI tests started failing,

https://github.com/axondeepseg/axondeepseg/actions/runs/12366489905

A single test is failing, one that ensures consistency of the calculated morphometrics values from a test image vs one that was saved when the test was written (or updated),

```

=========================== short test summary info ============================
FAILED test/morphometrics/test_compute_morphometrics.py::TestCore::test_morphometrics_consistency - assert False
[...]

PR #890, opened 2025-02-19
Update skimage and data-testing versions, update morphometrics pkl for consistency check

mathieuboudreau:
## Checklist

## Description

All our tests started failing again due to the data consistency check for our morphometrics,

https://github.com/axondeepseg/axondeepseg/actions/runs/13405546289/job/37444702835

This is due to some changes made in scikit-learn in the version they pushed yesterday (0.25.2) that seems to have tried to revert the issue we first flagged in #849, however when I checked the old pkl files the values dont appear to have reverted to their original value (see https://github.com/scikit-image/scikit-image/pull/7702#issuecomment-2668787118). So this PR pushes an update to our data-testing repo version that now includes a new pkl file that I generated, and updates our requirement for sckit-image to be >=0.25.2

## Linked issues

#849 and https://github.com/scikit-image/scikit-image/pull/7702

### DIFF
--- a/AxonDeepSeg/download_tests.py
+++ b/AxonDeepSeg/download_tests.py
@@ -10,7 +10,7 @@ def download_tests(destination=None):
         destination = convert_path(destination)
         test_files_destination = destination / "__test_files__"
 
-    url_tests = "https://github.com/axondeepseg/data-testing/archive/refs/tags/r20240219.zip"
+    url_tests = "https://github.com/axondeepseg/data-testing/archive/refs/tags/r20250219.zip"
     files_before = list(Path.cwd().iterdir())
 
     if (
--- a/AxonDeepSeg/download_tests.py
+++ b/AxonDeepSeg/download_tests.py
@@ -10,7 +10,7 @@ def download_tests(destination=None):
         destination = convert_path(destination)
         test_files_destination = destination / "__test_files__"
 
-    url_tests = "https://github.com/axondeepseg/data-testing/archive/refs/tags/r20250219b.zip"
+    url_tests = "https://github.com/axondeepseg/data-testing/archive/refs/tags/r20250110.zip"
     files_before = list(Path.cwd().iterdir())
 
     if (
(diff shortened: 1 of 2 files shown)


joshuacwnewton:
> however when I checked the old pkl files the values dont appear to have reverted to their original value (see [scikit-image/scikit-image#7702 (comment)](https://github.com/scikit-image/scikit-image/pull/7702#issuecomment-2668787118)).

I noticed you commented on the PR, but it might be better to comment on the underlying issue?

- https://github.com/scikit-image/scikit-image/issues/7661

Since there is a comment on that issue requesting testers:

>  plan to release 0.25.2 on Monday (2/17). If you have a chance, it would be great if you could test 0.25.2rc0:
> 
>     https://pypi.org/project/scikit-image/0.25.2rc0/

Maybe what we could do for now is specify `scikit-image<25.0`, restore the old 24.x testing values, and only choose to upgrade once that skimage confirms that the results are stable between 24.x and 25.x. (At which point we just avoid 25.1/25.2/etc.)

mathieuboudreau:
>Maybe what we could do for now is specify scikit-image<25.0, restore the old 24.x testing values, and only choose to upgrade once that skimage confirms that the results are stable between 24.x and 25.x. (At which point we just avoid the affected versions 25.0/25.1/25.2/etc.)

Good point! I'll make a new datarelease (because other files were updated between then) and push it

mathieuboudreau:
Hum, not sure why the tests are now failing with both the pkl and scikit image reverted. I'll debug locally and then push an update

mathieuboudreau:
Ok so I had to revert the actual version of the data-testing; it's possible that some of the other changed files impacted the resulting values or because there were generated post-scikit image updated version. Nonetheless, it works now, and I'll make it so that the data-testing master points to r20250110 so that we don't accidentally re-introduce those files in the next version

hermancollin (approved):
Should we open an issue to unpin skimage if they change this behavior?

joshuacwnewton:
> Should we open an issue to unpin skimage if they change this behavior?

Based on https://github.com/scikit-image/scikit-image/issues/7661#issuecomment-2669401941, it seems like 25.2 might have actually fixed things, and we were mistaken in thinking the issue was still with 25.2?

In that case, we could just do `scikit-image!=25.0,!=25.1` to avoid the two versions that contained the buggy behavior.

PR #901, opened 2025-03-06
Create true alias for ads_activate

mathieuboudreau:
## Checklist

## Description

I encountered an issue where our alias `ads_activate` would not work on a new user account on my laptop; after discussing with @joshuacwnewton we found that the culprit is that our current file in the /bin opens in a subprocess, thus not activating it in certain cases. Creating a true alias in the users zsh (or wtv) file resolves this

### DIFF
--- a/install_ads
+++ b/install_ads
@@ -298,7 +298,9 @@ function edit_shellrc() {
     # add line
     echo ""
     # Add ads_activate alias
-    alias ads_activate="source $ADS_DIR/ads_conda/bin/activate $ADS_DIR/ads_conda/envs/venv_ads/venv_ads/"
+    echo "alias ads_activate=\"source $ADS_DIR/ads_conda/bin/activate $ADS_DIR/ads_conda/envs/venv_ads/venv_ads/\""
+    # add line
+    echo ""
   ) >> "$RC_FILE_PATH"
 }
 
--- a/install_ads
+++ b/install_ads
@@ -297,6 +297,8 @@ function edit_shellrc() {
     fi
     # add line
     echo ""
+    # Add ads_activate alias
+    alias ads_activate="source $ADS_DIR/ads_conda/bin/activate $ADS_DIR/ads_conda/envs/venv_ads/venv_ads/"
   ) >> "$RC_FILE_PATH"
 }
 
@@ -653,8 +655,6 @@ for file in "$ADS_DIR/$CONDA_DIR"/envs/venv_ads/bin/*download*; do
   cp "$file" "$ADS_DIR/$BIN_DIR/" || die "Problem creating launchers!"
 done
 cp "$ADS_DIR/$CONDA_DIR/envs/venv_ads/bin/napari" "$ADS_DIR/$BIN_DIR/ads_napari" || die "Problem creating launchers!"
-echo "source $ADS_DIR/ads_conda/bin/activate venv_ads" > "$ADS_DIR/$BIN_DIR/ads_activate"
-chmod +x "$ADS_DIR/$BIN_DIR/ads_activate"
 
 # Activate the launchers, particularly download_model, download_tests, and axondeepseg_test
 export PATH="$ADS_DIR/$BIN_DIR:$PATH"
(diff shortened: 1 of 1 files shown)


install_ads
```diff
@@ -653,7 +653,7 @@ for file in "$ADS_DIR/$CONDA_DIR"/envs/venv_ads/bin/*download*; do
   cp "$file" "$ADS_DIR/$BIN_DIR/" || die "Problem creating launchers!"
 done
 cp "$ADS_DIR/$CONDA_DIR/envs/venv_ads/bin/napari" "$ADS_DIR/$BIN_DIR/ads_napari" || die "Problem creating launchers!"
-echo "source $ADS_DIR/ads_conda/bin/activate $ADS_DIR/ads_conda/envs/venv_ads/" > "$ADS_DIR/$BIN_DIR/ads_activate"
+echo "source $ADS_DIR/ads_conda/bin/activate venv_ads" > "$ADS_DIR/$BIN_DIR/ads_activate"
```
joshuacwnewton, line 656: Just to summarize the context behind our use of `source activate`: 

- Normally, conda recommends using `conda activate` to activate environments.
- But, if we try to use `conda activate venv_ads`, it doesn't work due to a `CondaError`:
    ```console
    dot|joshua@mosaic:~$ $ADS_DIR/ads_conda/bin/conda activate venv_ads
    
    CondaError: Run 'conda init' before 'conda activate'
    ```
- The reason this happens is because we're actively trying **not** to make the ADS conda available via `conda init`. We want it to be its own isolated environment, hidden from the user, only accessible via `ads_activate`. If we were to `conda init`, then that would edit the user's `.bashrc`/`.zshrc` in an ugly way, and we don't want that.
- So, we have to fall back on the legacy activation script (`source activate`), which has its own quirks compared to `conda activate`.

joshuacwnewton: When I check the help for `source activate`, I see that it accepts name or prefix (where "prefix" just means path in conda lingo, apparently):

```console
(venv_sct) dot|joshua@mosaic:~$ source $ADS_DIR/ads_conda/bin/activate --help

ActivateHelp: usage: conda activate [-h] [--[no-]stack] [env_name_or_prefix]

Activate a conda environment.

Options:

positional arguments:
  env_name_or_prefix    The environment name or prefix to activate. If the
                        prefix is a relative path, it must start with './'
                        (or '.\' on Windows).
```

Curious, I also tried using the path again on my own system:

```python
dot|joshua@mosaic:~$ conda --help
conda: command not found
dot|joshua@mosaic:~$ source $ADS_DIR/ads_conda/bin/activate $ADS_DIR/ads_conda/envs/venv_ads
(venv_ads) dot|joshua@mosaic:~$ 
```

And it activated the environment just the same. (Note: I **do not** have conda installed globally.)

So maybe it isn't a path vs name issue?

joshuacwnewton (commented):
Just a summary of our Slack chats :)

install_ads
joshuacwnewton: Indeed it did! `alias` works, while the `ads_activate` _script_ needed `source` to work (since it would activate the venv in a subshell).

mathieuboudreau:
@joshuacwnewton this alias version isn't working for me @hermancollin .

@hermancollin , maybe you could show us your output if any, and details about your system?

mathieuboudreau:
I think that could work, we would need to ensure that the conda / activate functions that are used are in the venv the user has, since the user might not have a system-wide conda installation. @joshuacwnewton any insights?

install_ads
```diff
@@ -297,6 +297,10 @@ function edit_shellrc() {
     fi
     # add line
     echo ""
+    # Add ads_activate alias
+    echo "alias ads_activate=\"source $ADS_DIR/ads_conda/bin/activate $ADS_DIR/ads_conda/envs/venv_ads/venv_ads/\""
```
joshuacwnewton, line 301: Oh. That's embarrassing. `venv_ads` is there twice. I assume that's the issue? @mathieuboudreau

mathieuboudreau:
@joshuacwnewton would it be overkill for us to add a new github actions YAML that tests that these features were installed correctly? i.e. on each os, run something like

`python -c "import AxonDeepSeg"`

and check that it's not found:

```
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'AxonDeepSeg'
```

and then run `ads_activate` and repeat to ensure it is now imported:

```
(venv_ads) (base) mathieuboudreau@Mathieus-MacBook-Pro axondeepseg % python -c "import AxonDeepSeg; print(AxonDeepSeg.__version_string__)"

AxonDeepSeg v5.1.1 - 297ff68f7e45d8da5b6a0b2ad1a49bec3a822ccf
(venv_ads) (base) mathieuboudreau@Mathieus-MacBook-Pro axondeepseg % 
```

PR #903, opened 2025-03-11
Pre-release changes for 5.1.0

mathieuboudreau:
## Checklist

## Description

Tests
* Adds a GitHub actions file to test installation via pipy

Installation
* Restricts ADS to only install with Python 3.11 and 3.12 (our pinned torch version is not compatible with >=3.13)

Documentation
* Added installation documentation for pipy and napari-hub: https://axondeepseg--903.org.readthedocs.build/en/903/

Plugin
* Adds a link to demo image
* Shortens citation
* Sets default model instead of popup to download; default model gets downloaded at initial segmentation if it doesnt exist.
* Changed the notice for the segmentation (may take a few minutes) from a small box in the image window to an actual popup

Tests
* Modified a test to deal with the new pop up box when clicking apply modell

Misc
* Refactors how things are downloaded in ads_utils; old way had become deprecated


hermancollin:
Thank you for these changes. I tried the recommended installation and everything went fine for ADS I think. The problem is that when I pip install napari, the installation doesn't work... but that's a separate issue.

I was thinking since there are more installation options, maybe we could use tabs to structure the doc, like we used to do in ivadomed. It would look like this. Curious to hear your thoughts.

Screenshot_20250314_104347

mathieuboudreau:
> I was thinking since there are more installation options, maybe we could use tabs to structure the doc, like we used to do in ivadomed. It would look like this. Curious to hear your thoughts.

I think that's a great idea! I hadn't thought of that, thanks!

mathieuboudreau:
>So yeah I'm not sure why pip install "napari[all]" gives me a broken install 

Can you elaborate? What's broken / what are the messages?

mathieuboudreau:
> > So yeah I'm not sure why pip install "napari[all]" gives me a broken install
> 
> Can you elaborate? What's broken / what are the messages?

@hermancollin?

hermancollin:
@mathieuboudreau hmm might be only on my side because I forgot to update my laptop since january 😅
I updated it yesterday, let me trying the install again.

hermancollin:
It seems I can't install the latest version of napari. I always get these errors:
```
19:07:41 : WARNING : MainThread : Could not initialize GLX
[1]    207882 IOT instruction (core dumped)  napari
```

I installed/uninstalled ~10 times like a madman and finally succesfully installed napari with conda instead of pip (`conda install -c conda-forge napari pyqt`). With this, the rest of the instructions worked :wave:

mathieuboudreau:
> It seems I can't install the latest version of napari. I always get these errors:
> 
> ```
> 19:07:41 : WARNING : MainThread : Could not initialize GLX
> [1]    207882 IOT instruction (core dumped)  napari
> ```
> 
> I installed/uninstalled ~10 times like a madman and finally succesfully installed napari with conda instead of pip (`conda install -c conda-forge napari pyqt`). With this, the rest of the instructions worked 👋

Hum - I wonder why the conda install would be different. Pinging @joshuacwnewton so he’s aware, maybe we should also see if napari or one of their dependencies pushed any breaking changes

joshuacwnewton:
> ```
> Could not initialize GLX
> ```

Lead me to [this QT forum post](https://forum.qt.io/post/653861) which lead me to [this suggestion](https://github.com/NixOS/nixpkgs/issues/82959#issuecomment-657306112), which is a temporary workaround to disable the GLX rendering backend.

```bash
QT_XCB_GL_INTEGRATION=none
```

More details in [this blog post](https://medium.com/@Spritan/troubleshooting-glx-and-egl-initialization-issues-in-qt-applications-could-not-initialize-glx-86e1e8612c55) which echoes the suggestion. If setting the environment variable works, then you can use it temporarily (or try to troubleshoot why the GLX backend is failing in the first place).

mathieuboudreau:
@hermancollin maybe you could compare the versions of a pip freeze / conda list of both of those environments?

joshuacwnewton:
Maybe we could open a separate issue for debugging this installation issue? I want to dive deeper but I also want to review this PR separately, hehehe. :stuck_out_tongue_closed_eyes:

hermancollin:
oh btw IIUC, when we pip install ADS, do we have access to the CLI tools? If so, I don't know how to test these because `which axondeepseg` always points to the environment created by the installation script.

joshuacwnewton:
Presumably, if you're pip installing ADS into a venv (`python -m venv venv_ads && source venv_ads/bin/activate && pip install napari axondeepseg`), activating the venv should modify the PATH so that the venv's commands are prioritized first? But if the script's commands are prioritized even with an activated venv, then that's an issue!

Temporarily, you can modify your `.bashrc`/`.zshrc` file and comment out the lines created by the `install_ads` script, then open up a new terminal window. That will at least take the script's version of ADS off the path.

hermancollin:
> Presumably, if you're pip installing ADS into a venv (`python -m venv venv_ads && source venv_ads/bin/activate && pip install napari axondeepseg`), activating the venv should modify the PATH so that the venv's commands are prioritized first? But if the script's commands are prioritized even with an activated venv, then that's an issue!
> 
> Temporarily, you can modify your `.bashrc`/`.zshrc` file and comment out the lines created by the `install_ads` script, then open up a new terminal window. That will at least take the script's version of ADS off the path.

Ok that makes sense. I'll create a separate issue to discuss these things

mathieuboudreau:
>Presumably, if you're pip installing ADS into a venv (python -m venv venv_ads && source venv_ads/bin/activate && pip install napari axondeepseg), activating the venv should modify the PATH so that the venv's commands are prioritized first? But if the script's commands are prioritized even with an activated venv, then that's an issue!


I just tried it out. Creating a new conda venv and then pip installing axondeepseg, `which axondeepseg` would still point to the version I installed from the code/repo via `install_ads`, as running that install tool always appends the path to that bin and it was the last entries in my zsh file. If I remove all those instances, and then uninstall and reinstall ads via pip in my conda venv, then ithe CLI tool installed by the pip install are viewable:

```
Successfully installed argparse-1.4.0 axondeepseg-5.0.4
(ads_pip) mathieuboudreau@Mathieus-MacBook-Pro ~ % which axondeepseg
/Users/mathieuboudreau/mambaforge/envs/ads_pip/bin/axondeepseg
(ads_pip) mathieuboudreau@Mathieus-MacBook-Pro ~ % 
```

.github/workflows/pipy_install_tests.yaml
```diff
@@ -0,0 +1,99 @@
...
+        run: |
+          pip install --upgrade pip
+          pip install axondeepseg==5.1.0rc1 --no-cache-dir
+
+      # Step 2a: Modify MKL_THREADING_LAYER (Ubuntu only)
+      # See https://github.com/pytorch/pytorch/issues/37377#issuecomment-629530272
+      - name: Set MKL_THREADING_LAYER
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          echo "MKL_THREADING_LAYER=GNU" >> $GITHUB_ENV
```
joshuacwnewton, line 67: Is this workaround still needed? The linked issue seems to explicitly reference `torch==1.5.0`, which we no longer pin.

.github/workflows/pipy_install_tests.yaml
```diff
@@ -0,0 +1,99 @@
...
+          ls
+
+          rmdir ./AxonDeepSeg/
+
+          python -c "import AxonDeepSeg.download_tests as download_tests; download_tests.main()"
+          python -c "import AxonDeepSeg.download_tests as download_tests; download_tests.main(['-d', './test/'])"
+          python -c "import AxonDeepSeg.download_model as download_model; download_model.main()"
+          python -c "import AxonDeepSeg.download_model as download_model; download_model.main(['-d', './AxonDeepSeg/models'])"
+
+          pytest.exe ./ -v
```
joshuacwnewton, line 99: Just a nitpick, but you could probably get away with deduplicating these steps by having everything but `pytest` be done in a separate step with `shell: bash`, then having separate steps and shells for `pytest` vs. `pytest.exe`.

docs/source/documentation.rst
```diff
@@ -32,56 +32,182 @@ Installation
...
--------------------
+Install Options
+---------------
 
-To install AxonDeepSeg, in a terminal window (macOS or Linux) or Command Prompt (Windows), "clone" AxonDeepSeg's repository (you will need to have ``git`` installed on your system) and then open the directory::
+.. tabs::
 
-    git clone https://github.com/neuropoly/axondeepseg.git
-    cd axondeepseg
+   .. tab:: Install via ``pip`` (recommended)
```
joshuacwnewton, line 40: Just for my own understanding, is there a reason why we've switched away from `install_ads` as the recommended method? (e.g. too many bugs, not a good fit for the ADS project, PyPI install made things easier, etc.)

joshuacwnewton (commented):
I don't have much to add re: the Napari changes, as I haven't quite familiarized myself with the ADS project yet to provide constructive feedback. 

But, I did have one or two thoughts re: the new GHA Workflow! Other than that, LGTM. :)

docs/source/documentation.rst
mathieuboudreau: I think it's mostly because we've had issues recently with some users that lacked a set of skills re: CLI, that I wanted to simplify things for them. To do that, I was kinda aiming to have them install the plugin inside napari using their plugin manager; however, this still requires the `pip install "napari[all]"` step and overall makes it a 2 step process - also, installing through the napari plugin manager requires the users to then shut down napari and restart it due to a quirk on their own side. 

Thus, the truly simplest installation instructions are to download directly via `pip`, as it also skips the github clone / github download step. I think for most users (80%), this will be a sufficient installation, it also locks them in to hopefully more stable versions.

Having said that, now with both @hermancollin and @Stoyan-I-A 's napari crashes when using the pip version, we may need to scrap that idea and have the installation via `install_ads` moved to default, omit the pip method, and only mention that it is also available through the napari plugin manager search

mathieuboudreau:
Ah - I finally just understood where the failing tests were coming from. Those are for the pypi install tests, so they are failing because we haven't pushed a release that contains the new aggregate code in it. I'll do that in a dev version and will re-run the tests.

Issue #957, opened 2026-02-11
Failing test due to redirect code


mathieuboudreau:
@hermancollin while working on the updated download data, did you rename a the data repo name> The URL valid tests are failing because it's giving us code 301 instead of 302, which appears to indicate that.

https://github.com/axondeepseg/axondeepseg/actions/runs/21893557436/job/63204637525

hermancollin:
Yeah I moved the unmyelinated model repo under ads-pipeline, I was curious. The tests weren't failing when we merged, so I don't know why the redirection fails on PyPI

hermancollin:
Ok but the thing that doesn't make sense is that the failing URL was removed from model_cards, so it shouldn't even try it.

This is the failing URL: https://github.com/axondeepseg/model_seg_unmyelinated_tem/releases/download/v1.1.0/model_seg_unmyelinated_sickkids_tem_best.zip

I changed this to:

https://github.com/axondeepseg/axondeepseg/blob/856cb22e026be6b8b8bdfab2e1ff9a0f7372fe6e/AxonDeepSeg/model_cards.yaml#L48

hermancollin:
Also, I can see that this job is running old tests. For example, on L328:

```
test/test_download_models.py::TestCore::test_download_model_cli_throws_error_for_unavailable_model PASSED [ 65%]
```

This test was removed in #955 

Before the tests are downloaded, we have this:
```
HEAD is now at 5b4d98b Release 5.3.0 (#936)
```

So I think the problem is we need to make a new release. But then why are the other jobs running fine? The windows tests are passing on PyPI, but there's something fishy happening before the tests are downloaded:

```
Run git checkout $( git tag | tail -1)
error: pathspec '$(' did not match any file(s) known to git
error: pathspec 'git' did not match any file(s) known to git
error: pathspec 'tag' did not match any file(s) known to git
tail: option used in invalid context -- 1
Invalid switch - "AxonDeepSeg".
```

The `$( ... )` syntax is not parsed properly so git checkout fails, thus staying on the latest commit where the test files are up-to-date. So tests are passing, but that's not even intended behavior 😅

hermancollin:
I made a new release (v5.4.0). Now, the good tests are loaded, but the job still installs v5.3.0.

I think I messed up because `pipy_install_tests.yaml` wasn't updated... I'll open a PR instead

PR #955, opened 2026-01-27
Refactor `download_model` command

hermancollin:
## Checklist

- [x] I've given this PR a concise, self-descriptive, and meaningful title
- [x] I've linked relevant issues in the PR body
- [x] I've applied [the relevant labels](https://intranet.neuro.polymtl.ca/geek-tips/contributing#pr-labels-a-href-pr-labels-id-pr-labels-a) to this PR
- [x] I've added relevant tests for my contribution
- [ ] I've updated the documentation and/or added correct docstrings
- [ ] I've assigned a reviewer
- [x] I've consulted [ADS's internal developer documentation](https://github.com/neuropoly/axondeepseg/wiki/Guidelines-for-contribution) to ensure my contribution is in line with any relevant design decisions

## Description
- As highlighted by users in #829 and #953, the CLI tool to download models is confusing. The `--list` option is the only way for the user to see which models are available, but it's very unclear how to go from there. 
- Also, I want to deprecate the `-t` option. When downloading a model, the `-t` option specifies if the model is a "single-fold" model or an ensemble of models. I think this is a confusing part for users, and it breaks the flow.
- Finally, I changed the format of the "model_cards" file (it was a python file, but there was no point in that)

### Models
Another aim of this PR is to
- add the latest stanford model
- remove the current `unmyelinated-TEM` one - this was trained on a pretty small dataset and I don't think we should put it out there. The stanford one should do a much better job at segmenting unmyelinated axons anyway.

### Cosmetics
I slightly improved the output of the `-l` option to avoid the confusion with the "long-form" model names.
<details><summary>Before</summary>

```
» download_model -l                                                                                                                                                                                      herman@tank
2026-01-27 17:43:08.005 | INFO     | AxonDeepSeg.download_model:main:97 - Printing available models:
2026-01-27 17:43:08.005 | INFO     | AxonDeepSeg.download_model:main:99 - generalist
{'MODEL NAME': 'model_seg_generalist',
 'NUMBER OF CLASSES': 2,
 'OVERVIEW': 'Multi-domain axon and myelin segmentation model trained on TEM, '
             'SEM, BF and CARS data.',
 'TRAINING DATA': 'Training data consists of an aggregation of 6 datasets '
                  'covering multiple microscopy modalities, species and '
                  'pathologies. For a detailed description of the 6 datasets '
                  'used for training, see https://arxiv.org/abs/2409.11552'}
[...]

PR #958, opened 2026-02-11
Release 5.4.0

hermancollin:
## Description

The main change is the refactoring of the `download_model` script in #955.

## Linked issues
#957


hermancollin:
@mathieuboudreau I created this release to update tests on PyPI. 

While investigating #957, I found that there was a problem with the gh workflow. There was a bash command substitution `$(...)` incompatible with windows CMD. I tried to fix this in 3b79b722cb59ccf1361be9461ac9281028e8b1d9 but is there a way to run the PyPI job on this PR to test it?

mathieuboudreau:
> @mathieuboudreau I created this release to update tests on PyPI.
> 
> While investigating #957, I found that there was a problem with the gh workflow. There was a bash command substitution `$(...)` incompatible with windows CMD. I tried to fix this in [3b79b72](https://github.com/axondeepseg/axondeepseg/commit/3b79b722cb59ccf1361be9461ac9281028e8b1d9) but is there a way to run the PyPI job on this PR to test it?

Yes - my worflow is to make it a "Release Candidate" (see here for syntax: https://packaging.python.org/en/latest/discussions/versioning/)  and push that to pypi, and use the release candidate in the test, if they pass, then I'll change it here in the tests and push the real version on pypi

mathieuboudreau:
Let me do that now

mathieuboudreau:
Oh one small caveat that I never explained I think, is that the pypi tests also will only run on master or in branches that begin with "release/". I updated it to test this branch explicitely now

mathieuboudreau (approved):
LGTM! The pipy tests passed with the release candidate - I just pushed the actual release and we can merge after those pass

hermancollin:
Ok thanks for all the insights. Last question - at this point, should we merge before or after publishing the release?

mathieuboudreau:
> Ok thanks for all the insights. Last question - at this point, should we merge before or after publishing the release?

It's already published -> https://pypi.org/project/AxonDeepSeg/5.4.0/

The wheel needs to be pushed to pipy before merging releases, because the tests just before the release is pushed need to be changed to that new release, and if the release wasn't on pipy it would fail and block the merge.

I'll merge it now!

PR #961, opened 2026-02-25
Diameter approximation overlay

hermancollin:
## Checklist

## Description
This feature allows us to visualize our area-based approximations for axon diameter and myelin thickness. A new `-d` option for the `axondeepseg_morphometrics` script triggers an additionnal output: an image suffixed _diameter_overlay_ (name is a WIP), with either circles or ellipses for every axon. For example, this segmentation:
image_seg-axonmyelin
Produces the following overlay for circle axon shape:
image_diameter_overlay-CIRCLE
Or, when using the ellipse axon shape:
image_diameter_overlay-ELLIPSE

We can then make nice visualizations of the diameter approximation. For example, if we overlay this on top of the segmentation and the image:
image_OVERLAY

### DIFF
--- a/AxonDeepSeg/params.py
+++ b/AxonDeepSeg/params.py
@@ -12,6 +12,7 @@ unmyelinated_index_suffix = Path('_uaxon_index.png')    # Colored unmyelinated a
 nnunet_suffix=Path('_seg-nnunet.png')                   # nnunet raw segmentation suffix
 nerve_suffix=Path('_seg-nerve.png')                     # nerve segmentation suffix
 nerve_index_suffix=Path('_nerve_index.png')             # Colored nerve segmentation + the index image
+diameter_overlay_suffix = Path('_diameter_overlay.png') # Diameter overlay with concentric circles
 
 side_effect_suffixes = tuple(
     [
--- a/AxonDeepSeg/params.py
+++ b/AxonDeepSeg/params.py
@@ -91,4 +91,8 @@ column_names_ordered = [
     Morphometrics_Column_Name('myelin_thickness','myelin_thickness (um)'),
     Morphometrics_Column_Name('axonmyelin_area','axonmyelin_area (um^2)'),
     Morphometrics_Column_Name('axonmyelin_perimeter','axonmyelin_perimeter (um)'),
+    Morphometrics_Column_Name('eccentricity'),
+    Morphometrics_Column_Name('orientation'),
+    Morphometrics_Column_Name('fiber_eccentricity'),
+    Morphometrics_Column_Name('fiber_orientation'),
 ]
--- a/AxonDeepSeg/params.py
+++ b/AxonDeepSeg/params.py
@@ -14,16 +14,6 @@ nerve_suffix=Path('_seg-nerve.png')                     # nerve segmentation suf
 nerve_index_suffix=Path('_nerve_index.png')             # Colored nerve segmentation + the index image
 diameter_overlay_suffix = Path('_diameter_overlay.png') # Diameter overlay with concentric circles
 
-side_effect_suffixes = tuple(
-    [
-        str(s) for s in [
-            axonmyelin_suffix, axon_suffix, myelin_suffix, index_suffix, 
-            axonmyelin_index_suffix, unmyelinated_suffix, unmyelinated_index_suffix,
-            nnunet_suffix
-        ]
-    ]
-)
-
 # morphometrics file suffix name
 morph_suffix = Path('axon_morphometrics.xlsx')
 morph_agg_suffix = Path('subject_morphometrics.xlsx')
@@ -31,6 +21,14 @@ unmyelinated_morph_suffix = Path('uaxon_morphometrics.xlsx')
 nerve_morph_suffix = Path('nerve_morphometrics.json')
 instance_im_suffix = Path('_colorized.png')             # Colored instance map of the segmentation
 instance_suffix = Path('_instance-map.png')             # Raw instance map of the segmentation
+
+# All pipeline-generated file suffixes — used by segment_folder() to skip outputs when
+# batch-processing a folder. Built automatically from all module-level Path variables
+# whose names end in '_suffix', so new suffixes are included without manual updates.
+generated_file_suffixes = tuple(
+    str(v) for k, v in list(globals().items())
+    if k.endswith('_suffix') and isinstance(v, Path)
+)
 
 # aggregate morphometrics file suffix name
 agg_dir = Path('morphometrics_agg')
--- a/AxonDeepSeg/postprocessing.py
+++ b/AxonDeepSeg/postprocessing.py
@@ -210,10 +210,42 @@ def generate_rotated_ellipse_points(center_x, center_y, semi_major, semi_minor, 
 def generate_diameter_overlay(stats_dataframe, image_shape, pixel_size, line_width=2, axon_shape='circle'):
     """
     Generate an overlay image with concentric circles or ellipses for each axon.
-    For circles: each axon has two circles: one with axon_diameter and one with axon_diameter + 2*myelin_thickness.
-    For ellipses: each axon has two ellipses with correct major/minor axes and orientation based on eccentricity and orientation columns.
-    
-    :param stats_dataframe: DataFrame containing axon morphometrics with columns: x0, y0, axon_diam, myelin_thickness, eccentricity, orientation, fiber_eccentricity, fiber_orientation
+
+    For circles: each axon has two circles: one with axon_diameter and one with
+    axon_diameter + 2*myelin_thickness.
+
+    For ellipses: the ellipses are NOT fits to the actual mask perimeters. They are
+    reconstructed from second-moment (inertia-tensor) statistics computed by
+    skimage.measure.regionprops (see
+    https://scikit-image.org/docs/stable/api/skimage.measure.html#skimage.measure.regionprops).
+
+    For the inner (axon) ellipse:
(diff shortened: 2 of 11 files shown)


hermancollin:
There is one thing I'm suspicious of. The way I add ellipses is that I add the myelin thickness to both the major and minor semi-axes:

https://github.com/axondeepseg/axondeepseg/blob/24a7568d06979ce4c0c201b8a3a63c6b41e7588c/AxonDeepSeg/postprocessing.py#L273-L275

But I suspect this is not representative of the actual ellipse approximation, because we don't see any variation of myelin thickness along the perimeter of the axon (which we would expect based on @mathieuboudreau's simulations).

mathieuboudreau:
> There is one thing I'm suspicious of. The way I add ellipses is that I add the myelin thickness to both the major and minor semi-axes:
> 
> https://github.com/axondeepseg/axondeepseg/blob/24a7568d06979ce4c0c201b8a3a63c6b41e7588c/AxonDeepSeg/postprocessing.py#L273-L275
> 
> But I suspect this is not representative of the actual ellipse approximation, because we don't see any variation of myelin thickness along the perimeter of the axon (which we would expect based on @mathieuboudreau's simulations).

If I understand how the algorithm works (or how I hope it works), there should be two ellipse estimates; one for the outer axon and one for the outer fiber diameter. Then the g-ratio should simply be computed from the minor axis diameters of both, regardless of if these two "lines" match up together (i.e. the myelin ellipse might be offset relative to the axon ellipse)

We should dig in the code to be sure though

hermancollin:
> If I understand how the algorithm works (or how I hope it works), there should be two ellipse estimates; one for the outer axon and one for the outer fiber diameter. Then the g-ratio should simply be computed from the minor axis diameters of both, regardless of if these two "lines" match up together (i.e. the myelin ellipse might be offset relative to the axon ellipse)

Yes, that is my understanding as well. In **compute_morphometrics.py**, the `evaluate_myelin_thickness_in_px` function takes the properties of the "axon" and "axonmyelin" objects, and spits out the difference in minor axis length. This means that my current implementation does not accurately reflect the approximation, because the ellipses may indeed have different orientations.

hermancollin:
I corrected this behavior, but it required adding 2 columns in the morphometrics dataframe: `fiber_orientation` and `fiber_eccentricity` which provide details for the axonmyelin objects. Now, outer ellipses are drawn with the correct orientation:

before:
Screenshot_20260303_120708


after:
Screenshot_20260303_120634

The weird thing is that according to the "cylinder axon model", oblique cuts should yield a compressed myelin thickness along the minor axis. However, in our case, the thickness is sometimes compressed along the major axis instead. Here is what the whole mask looks like:

image_diameter_overlay_ELLIPSE

mathieuboudreau:
@hermancollin I started reviewing this and pushed a commit to improve documentation/clarity.

One thing I realized while doing so, which isn't specific to this PR but important in its context, is that for the ellipse model, the myelin thickness is defined as the thickness along the minor axis between the elliptical axon and elliptical fibre boundaries. It is **not** an average myelin thickness between 1) the two ellipses or 2) between the two segmented boundaries. I think this is something really important to notify users/developers of; it's basically the assumption that there is an oblique cut that is causing the ellipse, and not a deformation (though possibly still applicable, I would have to test via simulations / your masks more).

Were you aware of this? I added it to the documentation

AxonDeepSeg/postprocessing.py
```diff
@@ -210,10 +210,42 @@ def generate_rotated_ellipse_points(center_x, center_y, semi_major, semi_minor,
...
     Generate an overlay image with concentric circles or ellipses for each axon.
-    For circles: each axon has two circles: one with axon_diameter and one with axon_diameter + 2*myelin_thickness.
-    For ellipses: each axon has two ellipses with correct major/minor axes and orientation based on eccentricity and orientation columns.
-    
-    :param stats_dataframe: DataFrame containing axon morphometrics with columns: x0, y0, axon_diam, myelin_thickness, eccentricity, orientation, fiber_eccentricity, fiber_orientation
+
+    For circles: each axon has two circles: one with axon_diameter and one with
+    axon_diameter + 2*myelin_thickness.
+
+    For ellipses: the ellipses are NOT fits to the actual mask perimeters. They are
```
hermancollin, line 217: this might lead to confusion because "fitting an ellipse" does not have a rigorous definition

hermancollin: the fitting doesn't happen at the perimeter-level but at the area-level

AxonDeepSeg/params.py
```diff
@@ -32,6 +22,14 @@
 instance_im_suffix = Path('_colorized.png')             # Colored instance map of the segmentation
 instance_suffix = Path('_instance-map.png')             # Raw instance map of the segmentation
 
+# All pipeline-generated file suffixes — used by segment_folder() to skip outputs when
+# batch-processing a folder. Built automatically from all module-level Path variables
+# whose names end in '_suffix', so new suffixes are included without manual updates.
+generated_file_suffixes = tuple(
+    str(v) for k, v in list(globals().items())
```
hermancollin, line 29: interesting! I never saw that `globals()` syntax

AxonDeepSeg/postprocessing.py
mathieuboudreau: >the fitting doesn't happen at the perimeter-level but at the area-level

Can you expand on this? For images, isn't that equivalent? Or do you mean the area of the ellipse of the axon will be equivalent to the area of the true axon shape?

hermancollin:
> @hermancollin I started reviewing this and pushed a commit to improve documentation/clarity.

Great!

> One thing I realized while doing so, which isn't specific to this PR but important in its context, is that for the ellipse model, the myelin thickness is defined as the thickness along the minor axis between the elliptical axon and elliptical fibre boundaries. It is **not** an average myelin thickness between 1) the two ellipses or 2) between the two segmented boundaries. I think this is something really important to notify users/developers of; 

Yes, this was my understanding since the beginning. However, I'm not sure about this:

> it's basically the assumption that there is an oblique cut that is causing the ellipse, and not a deformation (though possibly still applicable, I would have to test via simulations / your masks more).

I feel like this assumption breaks down immediately because for every fiber we are "fitting" two independent ellipses with (likely) 2 different orientations. With the _oblique cut assumption_, we expect a certain distribution of myelin around the axon (i.e. narrowest and "correct" thickness value at the minor axis and wider at the major axis). Not only is this not the case in segmentations (where we see a constant thickness all around), but if we fit 2 ellipses independently, we have no guarantee that the 2 ellipses match with "concentric cylinders" that cut the plane at an angle. 

Either way, when we have "elongated" axons, I feel like they often look deformed, not obliquely cut.

AxonDeepSeg/postprocessing.py
mathieuboudreau: >this might lead to confusion because "fitting an ellipse" does not have a rigorous definition

Fair, hmmm, what I meant was "The overlay isn't a failed attempt at fitting the myelin shape to an ellipse", but indicating that we compute the ellipse propreties, then using the centroid overlap the axon and myelin ellipses. It's kinda to explain the unexpected "look" we talked about at the major axis yesterday

mathieuboudreau: Can you think of a better way to explain it?

mathieuboudreau:
> > @hermancollin I started reviewing this and pushed a commit to improve documentation/clarity.
> 
> Great!
> 
> > One thing I realized while doing so, which isn't specific to this PR but important in its context, is that for the ellipse model, the myelin thickness is defined as the thickness along the minor axis between the elliptical axon and elliptical fibre boundaries. It is **not** an average myelin thickness between 1) the two ellipses or 2) between the two segmented boundaries. I think this is something really important to notify users/developers of;
> 
> Yes, this was my understanding since the beginning. However, I'm not sure about this:
> 
> > it's basically the assumption that there is an oblique cut that is causing the ellipse, and not a deformation (though possibly still applicable, I would have to test via simulations / your masks more).
> 
> I feel like this assumption breaks down immediately because for every fiber we are "fitting" two independent ellipses with (likely) 2 different orientations. With the _oblique cut assumption_, we expect a certain distribution of myelin around the axon (i.e. narrowest and "correct" thickness value at the minor axis and wider at the major axis). Not only is this not the case in segmentations (where we see a constant thickness all around), but if we fit 2 ellipses independently, we have no guarantee that the 2 ellipses match with "concentric cylinders" that cut the plane at an angle.
> 
> Either way, when we have "elongated" axons, I feel like they often look deformed, not obliquely cut.

Fair! I get what you mean now.

I guess we just need to be super super clear that the myelin thickness in  these cases isn't the average at all and *basically* (not exactly) a "single line measurement" like some histology folks do; however in this case, that measurement is made on the equivalent ellipses and not the axon itself (is that better though? haha)

mathieuboudreau (approved):
@hermancollin I did some documentation + minor big fixes (tangentially related) + added the overlay in napari when ellipse chosen + some key tests (not comprehensive, but what I felt should be defined behaviour).

On my side, the functionality ran really well - thank you!!

In napari, I set the color to yellow because I felt it gave the better contrast, but feel free to tinker with it.

Have a look at my changes / test them if you like, I'll approve it on my side though!

mathieuboudreau:
@hermancollin did you want me to release a new version with this change? It's the only PR since the prior version

hermancollin:
@mathieuboudreau as you wish. We could also wait for your modifications to the QC viewer and make a release then, as both features relate to the same project.

PR #963, opened 2026-03-17
Fix ellipse axon shape citations

### LINKED ISSUE #962: Incorrect citations in the doc?
I spotted this in the doc, in the section where we describe the difference between circle and ellipse axon shapes: https://axondeepseg.readthedocs.io/en/latest/documentation.html#ellipse

We claim this:

> Studies using Ellipse as axon shape:
> - Payne et al: https://pubmed.ncbi.nlm.nih.gov/21381867/
> - Payne et al: https://pubmed.ncbi.nlm.nih.gov/22879411/
> - Fehily et al: https://pubmed.ncbi.nlm.nih.gov/30702755/

However this might be a misleading reference for "ellipse axon shape". I checked all 3 papers (all from the same group) and in fact they never fit any ellipse. They do measure the shortest diameter, like our Stanford collaborators, but this is done manually using "_line-based measurements_" in ImageJ. In fact, for the myelin thickness, it is quite different: "_minimum myelin thickness (shown in green) [...] was averaged from two areas of compact myelin_" (see Fig below)

Image

Should we correct this, or am I missing something? To be more specific, I can think of one study with ellipse-fitting in segmentations: https://www.nature.com/articles/s41598-021-82575-9

hermancollin:
Fixes #962


mathieuboudreau (approved):
LGTM

Issue #971, opened 2026-04-02
Bug: `axondeepseg` crashes on large images (>178M pixels) with misleading "Background workers died" error


yolaatar:
## Describe the bug

Running `axondeepseg` on whole-slide TEM images larger than ~178M pixels causes a silent crash inside nnUNet's preprocessing subprocess. The reported error (`RuntimeError: Background workers died`) incorrectly suggests an out-of-memory issue. The actual root cause is PIL's `MAX_IMAGE_PIXELS` decompression bomb check firing inside a background worker, where its traceback is swallowed and never surfaced to the user.

## To Reproduce

```bash
axondeepseg \
  -i /path/to/large_image.tif \
  -m model_seg_unmyelinated_stanford_ensemble \
  --gpu-id 0
```

Where `large_image.tif` is any image exceeding ~178M pixels (e.g. 17652 × 12872 px = 227M pixels, which is a typical whole-slide TEM acquisition).

## Expected behavior

Inference runs successfully and saves segmentation masks.

## Actual behavior

The process crashes ~9 seconds after the model is loaded:

```
2026-04-01 20:09:20 | INFO    | AxonDeepSeg.segment:main:271 - AxonDeepSeg v5.4.0 - unknown commit
2026-04-01 20:09:22 | WARNING | AxonDeepSeg.segment:prepare_inputs:114 - ALIC_CC_II_06_P202503071741 will be converted in the expected .png format.
2026-04-01 20:09:30 | INFO    | AxonDeepSeg.apply_model:axon_segmentation:147 - Running inference on device: cuda:0
2026-04-01 20:09:35 | INFO    | AxonDeepSeg.apply_model:axon_segmentation:157 - Model successfully loaded.

Process SpawnProcess-11:

2026-04-01 20:09:39 | ERROR   | AxonDeepSeg.segment:main:298 - An error has been caught in function 'main':
Traceback (most recent call last):
  ...
  File "...\AxonDeepSeg\segment.py", line 169, in segment_images
    axon_segmentation(...)
  File "...\AxonDeepSeg\apply_model.py", line 165, in axon_segmentation
    predictor.predict_from_files(...)
  File "...\nnunetv2\inference\predict_from_raw_data.py", line 342, in predict_from_data_iterator
    for preprocessed in data_iterator:
  File "...\nnunetv2\inference\data_iterators.py", line 109, in preprocessing_iterator_fromfiles
    raise RuntimeError('Background workers died. Look for the error message further up! If there is '

[...]

PR #974, opened 2026-04-23
Add --allow-large-images flag to handle images exceeding PIL pixel limit

yolaatar:
Refs #971


yolaatar:
Added a confirmation popup when the user tries to segment a large image in napari. If the image exceeds PIL's pixel limit, it asks before proceeding.
Screenshot 2026-05-05 at 1 08 28 AM

mathieuboudreau:
@yolaatar please make branches in this repo and not in a fork - I can't push fixes in forks, and we prefer to keep all the work localized here so we can all access it.

I suggest we switch this into a PR of a branch of the same name here, and then create a new PR from that branch. I'll do that now.

PR #978, opened 2026-05-05
Update conda environment creation to include pip

### LINKED ISSUE #977: Failing tests on master since conda 26.3
2 days ago, tests started to fail on master for macOS and ubuntu machines. Prior to failures, we had conda 26.1.1, but now conda 26.3.2 gets installed during tests. Not sure if this is the culprit, but although pip is successfully installed, we get this:

```
/home/runner/work/axondeepseg/axondeepseg/ads_conda/envs/venv_ads/bin/python: No module named pip
```

mathieuboudreau:
Added pip installation to conda environment creation.

Flagged in #977

### DIFF
--- a/.github/workflows/run_tests.yaml
+++ b/.github/workflows/run_tests.yaml
@@ -116,7 +116,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.github_token }}
         run: |
-          curl -sL https://github.com/coverallsapp/coverage-reporter/releases/latest/download/coveralls-darwin.tar.gz | tar -xz
+          curl -sL https://github.com/vtjnash/coveralls-macos-binaries/releases/latest/download/coveralls-macos-x86_64.tar.gz | tar -xz
           ./coveralls report --parallel --flag-name="${{ matrix.os }}" coverage.lcov
       - name: Submit to coveralls
         if: matrix.os != 'macos-26-intel'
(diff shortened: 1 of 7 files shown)


mathieuboudreau:
@joshuacwnewton just tagging you here so you were aware - seems like in the latest miniconda release, either the conda envs no longer "see" the base environment or the venv creation with a python version no longer installs pip by default; either way our conda env installs could no longer see a pip command once inside the venv which failed our CI tests in the past few days

.github/workflows/run_tests.yaml
```diff
@@ -108,7 +108,18 @@ jobs:
...
       # Step 4: Submnit to coveralls
+      # The coverallsapp/github-action installs the reporter via Homebrew on macOS,
+      # which is currently broken on macos-26-intel (Crystal 1.20 vs coveralls formula).
+      # For that runner we download the prebuilt darwin binary directly.
+      - name: Submit to coveralls (macOS Intel)
+        if: matrix.os == 'macos-26-intel'
+        env:
+          GITHUB_TOKEN: ${{ secrets.github_token }}
+        run: |
+          curl -sL https://github.com/coverallsapp/coverage-reporter/releases/latest/download/coveralls-darwin.tar.gz | tar -xz
```
hermancollin: this URL returns Not found

hermancollin: Maybe this?
```suggestion
          curl -sL https://github.com/vtjnash/coveralls-macos-binaries/releases/latest/download/coveralls-macos-x86_64.tar.gz | tar -xz
```

hermancollin: This still fails because `coveralls report` doesn't recognize the `--flag-name` arg

mathieuboudreau: Woops, on it

mathieuboudreau: You know what, it really doesn't matter what mac intel reports the coverage to be on intel, we can just skip it for this runner

hermancollin (approved):
nice!

PR #983, opened 2026-05-15
feat(napari): segmentation progress bar + cancel button

yolaatar:
## Summary

- Adds a `napari.utils.progress` bar that appears in the napari **Activity dock** when the user clicks "Segment image"
- Monkey-patches nnUNet's internal `tqdm` reference with a generator that emits a Qt signal per patch; the signal handler updates the progress bar on the main thread (avoids Qt cross-thread QObject crash)
- Automatically opens the Activity dock at inference start so users don't have to find it manually
- Progress bar is closed cleanly in a `finally` block regardless of success or failure

Closes #979 (progress bar during segmentation)

## How to test

1. Load a large image in napari
2. Select a model and click **Segment image**
3. The Activity dock should open and show a progress bar advancing through nnUNet's sliding-window patches

## Notes

- The monkey-patch is scoped to `ApplyModelThread.run()` and always restored in `finally`, so it does not affect other uses of nnUNet
- Does not require any changes to nnUNet or napari internals

### DIFF
--- a/AxonDeepSeg/ads_napari/_widget.py
+++ b/AxonDeepSeg/ads_napari/_widget.py
@@ -38,6 +38,7 @@ from AxonDeepSeg.visualization.colorization import colorize_instance_segmentatio
 
 import napari
 from napari.utils.notifications import show_info
+from napari.utils import progress as napari_progress
 from .settings_menu_ui import Ui_Settings_menu_ui
 from vispy.util import keys
 import webbrowser
@@ -198,6 +199,9 @@ class ADSplugin(QWidget):
         self.apply_model_thread = ApplyModelThread()
         self.apply_model_thread.model_applied_signal.connect(
             self._on_model_finished_apply
+        )
+        self.apply_model_thread.progress_update.connect(
+            self._on_segmentation_progress
         )
 
         load_mask_button = QPushButton("Load mask")
@@ -485,6 +489,18 @@ class ADSplugin(QWidget):
         else:
             return False
 
+    def _on_segmentation_progress(self, current, total, desc):
+        """Update the napari Activity dock progress bar from the main thread."""
+        pbr = self.apply_model_thread.progress_bar
+        if pbr is None:
+            return
+        if pbr.total != total:
+            pbr.total = total
+            pbr.reset()
+        pbr.set_description(desc)
+        pbr.n = current
+        pbr.events.value(value=current)
+
     def _on_apply_model_button_click(self, layer):
         """Apply the selected AxonDeepSeg model to the active layer of the viewer.
 
@@ -542,6 +558,15 @@ class ADSplugin(QWidget):
         self.apply_model_thread.path_model = model_path
         self.apply_model_thread.gpu_id = self.settings.gpu_id
         self.apply_model_thread.allow_large_images = allow_large_images
+        self.apply_model_thread.progress_bar = napari_progress(
+            total=0, desc="Segmenting image..."
+        )
+        try:
+            sb = self.viewer.window._qt_window.statusBar()
+            if not self.viewer.window._qt_window._activity_dialog.isVisible():
+                sb._toggle_activity_dock()
+        except Exception:
+            pass
         self.show_info_message(
             "Running AI model... This can take a few seconds or minutes. Check the console/terminal for more information."
         )
@@ -1333,6 +1358,7 @@ class ApplyModelThread(QtCore.QThread):
     """
 
     model_applied_signal = Signal()
+    progress_update = Signal(int, int, str)  # (current, total, description)
 
     def __init__(self):
         """Initializes an instance of the class with default values for attributes.
@@ -1353,6 +1379,7 @@ class ApplyModelThread(QtCore.QThread):
         self.gpu_id = -1
         self.task_finished_successfully = False
         self.allow_large_images = False
+        self.progress_bar = None
 
     def run(self):
         """Executes the segmentation process in a separate thread on the selected image layer using the AxonDeepSeg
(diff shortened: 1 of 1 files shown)


hermancollin:
The progress bar is partially hidden by the tools view when the napari window is small (this is really a small nitpick). @yolaatar do you think there would be an easy solution to this?

<details>
Screenshot 2026-05-22 at 1 18 44 PM


## But it looks fine in fullscreen or if the window is big enough:
Screenshot 2026-05-22 at 1 19 54 PM
</details>

Also, when the segmentation is done, the activity dock stays open. Perhaps it would be cleaner to close it automatically?
<details>
Screenshot 2026-05-22 at 1 20 45 PM
</details>

edit: I implemented this last thing in 6957154

AxonDeepSeg/ads_napari/_widget.py
```diff
@@ -542,6 +558,15 @@ def _on_apply_model_button_click(self, layer):
...
         self.apply_model_thread.gpu_id = self.settings.gpu_id
         self.apply_model_thread.allow_large_images = allow_large_images
+        self.apply_model_thread.progress_bar = napari_progress(
+            total=0, desc="Segmenting image..."
+        )
+        try:
+            sb = self.viewer.window._qt_window.statusBar()
+            if not self.viewer.window._qt_window._activity_dialog.isVisible():
+                sb._toggle_activity_dock()
+        except Exception:
```
hermancollin: can you explain what's going on here? why are we ignoring exceptions

yolaatar:
> The progress bar is partially hidden by the tools view when the napari window is small (this is really a small nitpick). @yolaatar do you think there would be an easy solution to this?
> 
> Details
> Screenshot 2026-05-22 at 1 18 44 PM
> ## But it looks fine in fullscreen or if the window is big enough:
> Screenshot 2026-05-22 at 1 19 54 PM
> Also, when the segmentation is done, the activity dock stays open. Perhaps it would be cleaner to close it automatically?
> 
> Details
> Screenshot 2026-05-22 at 1 20 45 PM
> edit: I implemented this last thing in [6957154](https://github.com/axondeepseg/axondeepseg/commit/695715452fa3e2df947c791e77a4b87e3fed2082)

Good catch on both. For the small window, the activity dock position is controlled by napari and I don't see a clean way to fix it on our end. For the dock staying open, thanks for fixing it directly!

AxonDeepSeg/ads_napari/_widget.py
yolaatar: It's actually intentional if it fails the dock just doesn't auto-open, which is fine, segmentation still runs normally. I didn't want to clutter the console with a warning for something that doesn't affect the actual workflow (we can definitely add a debug log if you want).  
Regarding the try/except, it's there because we're poking at private napari internals that could disappear in future versions.

hermancollin:
> For the small window, the activity dock position is controlled by napari and I don't see a clean way to fix it on our end.

@yolaatar is there a reason for using the activity dock specifically? For example, could we put it directly inside the ADS plugin panel?

Also, maybe something to consider is [napari.utils.cancelable_progress](https://napari.org/dev/api/napari.utils.cancelable_progress.html), which would allow us to give an option for the user to cancel the segmentation.

yolaatar:
> @yolaatar is there a reason for using the activity dock specifically? For example, could we put it directly inside the ADS plugin panel?

We could definitely move it to the ADS plugin panel. I first put it there because it felt more natural and integrated to napari (since we're using `napari_progress`,  which hooks into the activity dock natively). I can implement this and it would look exactly the same. Only downside is that it wouldn't be tied to the napari task system, but I guess it's fine. 

> Also, maybe something to consider is [napari.utils.cancelable_progress](https://napari.org/dev/api/napari.utils.cancelable_progress.html), which would allow us to give an option for the user to cancel the segmentation.

I think this would make the ADS panel integration more meaningful -- the cancel button would be better integrated in the UI than sitting in the activity dock.

hermancollin:
ok I found an alternative way to display the progress. Prior to segmentation, the UI looks normal:

<details>
Screenshot_20260525_104500

</details>

Then, when you click on the segment image button, the progress bar is displayed directly inside the ADS plugin panel:

<details>
Screenshot_20260525_104725

Screenshot_20260525_104815


</details>

and then when the segmentation is done, the progress bar disappears. @yolaatar tell me what you think. I took the liberty of commiting the changes on this branch, but we can always revert back if we prefer the Activity Dock solution.

hermancollin:
> > @yolaatar is there a reason for using the activity dock specifically? For example, could we put it directly inside the ADS plugin panel?
> 
> We could definitely move it to the ADS plugin panel. I first put it there because it felt more natural and integrated to napari (since we're using `napari_progress`, which hooks into the activity dock natively). I can implement this and it would look exactly the same. Only downside is that it wouldn't be tied to the napari task system, but I guess it's fine.

We could have both the progress bar inside the menu and the napari_progress, and simply avoid opening the activity dock. This way, the progress would still be tied to napari's task system.

yolaatar:
I was about to implement pretty much the same thing, well done ^^
I can still take care of the cancel button, I have a few ideas on how we can make it work.

hermancollin:
Sounds good!

yolaatar:
Added a Cancel button next to "Segment image". 
<details>
Screenshot 2026-05-25 at 5 23 09 PM
</details>          

It's disabled by default and activates when segmentation starts. 
<details>
Screenshot 2026-05-25 at 4 01 52 PM
</details>                     
Cancellation works by setting a flag that the inference loop checks between patches -- so it interrupts at the next patch boundary rather than mid-patch. If the segmentation is cancelled, masks are not loaded into the viewer.

hermancollin:
That's sick! I'm not sure about the location of the cancel button tho. Maybe a more intuitive UX would be to make the cancel button appear only when the segmentation has started? 

What's your opinion on having the horizontal box layout contain the progress bar on the left and the cancel button on the right, and the whole box appears only during segmentation?

hermancollin:
Here is what I mean:
## Option 1: Cancel button always displayed
<details>

### Idle
Screenshot_20260525_143555

### During segmentation
Screenshot_20260525_143301
</details>

## Option 2: Cancel button with progress bar

<details>

### Idle 
(UI is identical to master branch while idle)
Screenshot_20260525_145224

### During segmentation:
Screenshot_20260525_145412

(and the QHBoxLayout disappears after segmentation)

</details>

If we go with the latter, I have a commit ready to go.

yolaatar:
Option 2 looks much cleaner, you can go ahead with your commit!

hermancollin:
I think this PR is ready for another round of review. To test it, just load any image and segment it!

Stoyan-I-A:
Tested it on my end, looks good
image

Stoyan-I-A (approved):
LGTM

PR #976, opened 2026-05-05
Add --allow-large-images flag to handle images exceeding PIL pixel limit

mathieuboudreau:
Refs https://github.com/axondeepseg/axondeepseg/issues/971

Moved from this PR of a fork: https://github.com/axondeepseg/axondeepseg/pull/974

PR #988, opened 2026-06-15
fix: --allow-large-images flag not applied in get_imshape

yolaatar:
`get_imshape` called `imageio.v3.improps()` without raising PIL's pixel limit first, so images over ~178 Mpx would throw a `DecompressionBombError` even when `--allow-large-images` was passed.

Fix: propagate `allow_large_images` into `get_imshape` and temporarily raise `Image.MAX_IMAGE_PIXELS` before the imageio call.

Refs #976


hermancollin:
wow, I was wondering how we completely missed this in #976. The tests I added were specifically mocking get_imshape 😅 

https://github.com/axondeepseg/axondeepseg/blob/c4d202865e7c08367a04421fa4707b47c47a517a/test/test_ads_utils.py#L193-L211

anyway, I fixed this problem in 2090c3a

hermancollin (approved):
thanks for the fix @yolaatar

Issue #986, opened 2026-06-08
Show download status and available updates in `download_model --list`


yolaatar:
## Summary
 
Right now, `download_model --list` shows info about each model but doesn't tell you whether you already have it installed or whether a newer version is available. Would be nice to have both.
 
## What I'm suggesting
 
1. **Show if a model is installed** — just a simple indicator next to the model name, something like `✓ Installed` or `x Not installed`.
2. **Warn if an installed model is outdated** — if a newer version of a model is available (e.g. retrained on more data), flag it and tell the user to re-download. For example:
   ```
   Model name    generalist
   Status        ✓ Installed (update available — run: download_model -m generalist)
   ```
 
## Why
 
Models get updated over time and there's currently no way to know your local version is stale unless you check manually. This is especially relevant now that we're working on retraining the generalist model.
 
## Notes
 
- Could use a version tag or hash stored in the model metadata, compared against what's on disk after download.
- The remote check could just happen when running `--list` since it already fetches model info anyway.

jcohenadad:
we do have a similar mechanism in SCT-- @joshuacwnewton can point you to the relevant code

<img width="1360" height="638" alt="Image" src="https://github.com/user-attachments/assets/3b81f5c0-8d13-4125-85a6-790a2c7e474b" />

joshuacwnewton:
> we do have a similar mechanism in SCT-- [@joshuacwnewton](https://github.com/joshuacwnewton) can point you to the relevant code

Sure thing! We have two ways of displaying things:

- [**Short table**](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/c07c85c3cad635d94e82cfbe4da1356446ac1141/spinalcordtoolbox/deepseg/models.py#L852-L891) (displayed as an epilog in the argparse help, screenshotted by Julien above)
    * Note that models are grouped by their purpose, and this grouping is defined in the [`TASKS`](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg/models.py#L249-L624) dictionary.
- [**Long table**](https://github.com/spinalcordtoolbox/spinalcordtoolbox/blob/master/spinalcordtoolbox/deepseg/models.py#L894-L955) (displayed when user passes `-list-tasks`)

joshuacwnewton:
(Side note: The "update available" is a nice feature idea, though maybe less relevant for SCT since reinstalling a new version of SCT also reinstalls the models.)

[...]

PR #989, opened 2026-06-15
Show install status in download_model --list

yolaatar:
Refs #986

### DIFF
--- a/AxonDeepSeg/download_model.py
+++ b/AxonDeepSeg/download_model.py
@@ -14,6 +14,7 @@ MODEL_NOT_FOUND_CODE = 1
 DOWNLOAD_ERROR_CODE = 2
 
 MODEL_CARDS_PATH = Path(__file__).parent / 'model_cards.yaml'
+MODEL_VERSION_FILENAME = 'model_version.txt'
 
 def download_model(model_name='generalist', destination=None, overwrite=True):
     '''
@@ -70,6 +71,10 @@ def download_model(model_name='generalist', destination=None, overwrite=True):
 
     shutil.move(folder_name, str(model_destination))
 
+    version = models[model_name].get('version')
+    if version:
+        (model_destination / MODEL_VERSION_FILENAME).write_text(version)
+
     return output_dir
 
 def format_pixel_size(pixel_size) -> str:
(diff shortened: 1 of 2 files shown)


yolaatar:
The current PR covers installed/not installed display and lays the groundwork for version tracking (writing `model_version.txt` on download, `version:` field in `model_cards.yaml`). What's not yet handled is detecting whether an installed model is outdated. Two approaches worth discussing:

**Option A -- fetch `model_cards.yaml` remotely**
Keep the current URL-based release workflow. When releasing a new model, update both the download URL and the `version:` field in `model_cards.yaml`. `download_model --list` fetches the latest `model_cards.yaml` from the ADS repo at runtime and compares the remote `version:` against the locally installed one. Simple, low surface area change.

**Option B -- GitHub Releases API**
Replace hardcoded URLs in `model_cards.yaml` with just the model repo + asset name. At download time and on `--list`, hit the GitHub Releases API to get the latest tag and download URL dynamically. A new model release on the model repo is immediately visible to ADS users with no ADS PR needed. Cleaner long-term, but changes the release workflow more significantly.

Both are reasonable given how infrequently models are released. Happy to go either direction.

hermancollin:
> **Option B -- GitHub Releases API** Replace hardcoded URLs in `model_cards.yaml` with just the model repo + asset name. At download time and on `--list`, hit the GitHub Releases API to get the latest tag and download URL dynamically. A new model release on the model repo is immediately visible to ADS users with no ADS PR needed. Cleaner long-term, but changes the release workflow more significantly.

This will not work, because the latest tag is not a reliable indicator to flag outdated models. For example, both "light" and "ensemble" models are released from the same repo, but might be released independently. If we put a new version of the "light" model in a new release, it will mark the both the light and ensemble versions as outdated. Also, if we make a release of the repository without adding new models (e.g. to point to a specific version in a publication), it will mark the models as outdated.

Option A sounds more feasible, but there is still an annoying problem to solve. If you put a version for model A, it merges together the version of the light and ensemble versions. Maybe have 2 separate versions per model, like `version_light` and `version_ensemble`? Not super elegant but it might do the trick.

PR #1001, opened 2026-08-16
Add multires TEM model

yolaatar:
Adds model_seg_tem-multires, trained across multiple pixel sizes on TEM1+TEM2 for resolution invariance.


hermancollin (approved):
Ahhh windows.

Thanks you for the fix, now it works. Changing the release assets would have been simpler, but now I guess we're covered in case anyone compresses a model to distribute from a Windows machine. LGTM!

PR #1004, opened 2026-08-26
Fix pipy CI on Windows + Mac OS 3.12 runner

mathieuboudreau:
Nightly build have been failing - this fixes two issues; 1) the windows builds that were failing when ADS was installed via pipy, and 2) the Mac OS 3.12 runner that we were previously skipping due to a separate issue.


hermancollin (approved):
lgtm!

PR #1005, opened 2026-09-03
Morphometrics filtering + Axon counting

### LINKED ISSUE #900: Morphometric postprocessing tool
**Describe the problem**
Following this comment https://github.com/axondeepseg/axondeepseg/issues/892#issuecomment-2704567856, it would be nice to have a separate tool to fix this:
- raw morphometrics often contain invalid values, such as gratio >= 1 or NaNs
- also, sometimes it makes sense to remove lines in the morphometric files based on prior knowledge (e.g. exclude axons with diameter inferior to _x_ um, which are most likely artifacts)
- the user can postprocess the morphometrics table himself, but the corrections are not reflected on the segmentation masks

**Proposed solution**
A new entry point or most likely a `axondeepseg_morphometrics` subcommand to filter the morphometrics based on predefined rules. We could develop a way for the user to specify his own rules (`gratio` ∈ **]0,1[**, `axon_diam` > 0.1, `border_touching` = False, etc.)

This would then update the morphometric CSVs and the segmentation masks as well.

### LINKED ISSUE #1003: Missing combined "axonmyelin" mask after 3-class model inference
This is a problem I've noticed some time ago, but it's getting really annoying. When you use a 3-class model (myelinated + unmyelinated axons), the combined "axonmyelin" mask doesn't get created. This is a bug, and it's a problem because the morphometrics script **requires** this image to overlay index numbers onto.

The problem is here: 

https://github.com/axondeepseg/axondeepseg/blob/db6ea61be8fcb6bca36407ea8a14dadd1ab6b177/AxonDeepSeg/apply_model.py#L212

We should actually check if these 2 classes are contained in all output classes instead of checking for identity. Later on,

https://github.com/axondeepseg/axondeepseg/blob/db6ea61be8fcb6bca36407ea8a14dadd1ab6b177/AxonDeepSeg/apply_model.py#L226-L227

So the masks are never merged. This will be a very quick fix.

hermancollin:
## Checklist

## Description
This PR addresses 3 points related to axon counting. The goal was essentially to streamline the process of counting axons, including filtering outliers based on morphometric features.

### 1. Saving "axonmyelin" mask for models with more than 2 output classes
This one is self explanatory. Described in #1003. This was a bug, but it needed fixing for the other 2 features.

### 2. Filtering morphometrics
This was described in #900 but we also discussed it regularly. Filtering the morphometric files is very useful because it can help us spot segmentation artifacts. It is also required to obtain a reliable axon count.

I implemented a new script to do this. The filtering rules are concisely specified in a YAML file. The user can simply modify the values in this config file located in _AxonDeepSeg/morphometrics/filter.yaml_. Currently there are only 5 rules (2 for myelinated axons, 3 for unmyelinated) based on what I was doing on my projects, but we can easily extend it in the future if necessary. This is what the default config file looks like:

```yaml
myelinated:
  - valid-g-ratio-only: True
  - axon-diam-gt: ~

unmyelinated:
  - axon-diam-gt: ~
  - solidity-gt: ~
  - axon-area-lt: ~
```
I purposefully left all values empty because default values would be problematic and, as discussed previously in internal meetings, it's better to let users decide for themselves. The `valid-g-ratio-only` rule is actually enforced by default, because g-ratio values outside of the ]0,1[ range are artifacts. This is the help message for this new script:

```
usage: axondeepseg_filter [-h] -i INPUT [-c CONFIG] [-m] [-o]

options:
  -h, --help            show this help message and exit
  -i INPUT, --input INPUT
                        Path to the input morphometric file or folder containing morphometric files
  -c CONFIG, --config CONFIG
                        Path to the configuration file containing filtering criteria (defaults to AxonDeepSeg/morphometrics/filter.yaml)
  -m, --update_masks    Update the segmentation masks to reflect the filtered morphometrics
  -o, --overwrite       Overwrite the original morphometric files (and segmentation masks if -m is set)
```

By default, new morphometric files will be created with the `_filtered.xlsx` suffix. Optionally, note that the script can modify the segmentation masks to reflect the filtering.

### 3. Count axons
Finally, I added a lightweight script to count axons (and prioritize `_filtered` suffixed files when possible) in every image of a given directory. This is useful because it writes a single CSV file with myelinated and unmyelinated axon counts for every image in the directory. Adapted from [this script](https://github.com/axondeepseg-pipeline/model_seg_unmyelinated_tem/blob/main/scripts/utils/count_axons.py).

```
axondeepseg_count -h  
usage: axondeepseg_count [-h] [-i INPUT_DIR] [-m] [-o OUTPUT_NAME]

Count axons (myelinated and unmyelinated, if applicable).

options:
  -h, --help       show this help message and exit
  -i INPUT_DIR     Path to the folder containing all morphometric files.
  -m, --mask_mode  Toggles mask mode: use masks instead of .xlsx to count axons.
  -o OUTPUT_NAME   Name of the output file.
```

This is an example output:
Screenshot 2026-09-03 at 1 45 33 PM

## Linked issues
Fixes #900 
Fixes #1003

### DIFF
--- a/AxonDeepSeg/morphometrics/count_axons.py
+++ b/AxonDeepSeg/morphometrics/count_axons.py
@@ -1,0 +1,114 @@
+'''
+Finds all morphometric files in a folder and counts the number of axons (myelinated and unmyelinated, 
+if applicable) in each file. The results are saved to a CSV file.
+'''
+from pathlib import Path
+from tqdm import tqdm
+import argparse
+from loguru import logger
+
+import pandas as pd
+from skimage import measure
+
+from AxonDeepSeg.ads_utils import imread
+from AxonDeepSeg.params import generated_file_suffixes
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Count axons (myelinated and unmyelinated, if applicable).')
+    parser.add_argument(
+        '-i',
+        dest='input_dir',
+        type=str,
+        help='Path to the folder containing all morphometric files.'
+    )
+    parser.add_argument(
+        '-m', '--mask_mode',
+        action='store_true',
+        default=False,
+        help='Toggles mask mode: use masks instead of .xlsx to count axons.'
+    )
+    parser.add_argument(
+        '-o',
+        dest='output_name',
+        type=str,
+        default='axon_counts.csv',
+        help='Name of the output file.',
+    )
+
+    args = parser.parse_args()
+    input_dir = Path(args.input_dir)
+    out_name = args.output_name
+    mask_mode = args.mask_mode
+    assert input_dir.exists(), f'Directory {input_dir} does not exist.'
+
+    # find images except for the masks
+    additional_suffixes = [
+        '_seg-axonmyelin_filtered.png', '_seg-uaxon_filtered.png', 
+        '_seg-axon_filtered.png', '_seg-myelin_filtered.png',
+        '_seg-nuclei.png', '_seg-process.png', 
+    ]
+    ignore_suffixes = [str(s) for s in generated_file_suffixes if str(s).endswith('.png')] + additional_suffixes
+    ignore_suffixes = tuple(ignore_suffixes)
+    inputs = [f for f in input_dir.glob('*.png') if not f.name.endswith(ignore_suffixes)]
+
+    counts = {'image': [], 'axon_count': [], 'uaxon_count': []}
+    axon_morph_suffix = '_axon_morphometrics.xlsx'
+    uaxon_morph_suffix = '_uaxon_morphometrics.xlsx'
+
+    total_axon_count = 0
+    total_uaxon_count = 0
+    total_size = 0
+    for img in tqdm(inputs):
+        if not mask_mode:
+            # read morphometric files
+            target_axon_file = str(img.with_suffix('')) + axon_morph_suffix
+            target_uaxon_file = str(img.with_suffix('')) + uaxon_morph_suffix
+            # filtered morphometrics take precedence over unfiltered morphometrics
+            filtered_axon_file = str(img.with_suffix('')) + '_axon_morphometrics_filtered.xlsx'
+            filtered_uaxon_file = str(img.with_suffix('')) + '_uaxon_morphometrics_filtered.xlsx'
+            if Path(filtered_axon_file).exists():
+                target_axon_file = filtered_axon_file
+            if Path(filtered_uaxon_file).exists():
+                target_uaxon_file = filtered_uaxon_file
+
+            axon_count = len(pd.read_excel(target_axon_file))
+            uaxon_count = len(pd.read_excel(target_uaxon_file))
+        else:
+            target_axonmyelin_mask = str(img.with_suffix('')) + '_seg-axonmyelin.png'
+            target_uaxon_mask = str(img.with_suffix('')) + '_seg-uaxon.png'
+            filtered_axonmyelin_mask = str(img.with_suffix('')) + '_seg-axonmyelin_filtered.png'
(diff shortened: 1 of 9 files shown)


AxonDeepSeg/morphometrics/count_axons.py
```diff
@@ -0,0 +1,114 @@
...
+from skimage import measure
+
+from AxonDeepSeg.ads_utils import imread
+from AxonDeepSeg.params import generated_file_suffixes
+
+
+def main(argv=None):
+    parser = argparse.ArgumentParser(description='Count axons (myelinated and unmyelinated, if applicable).')
+    parser.add_argument(
+        '-i',
```
mathieuboudreau, line 20: usually (in our other CLI commands) ,`-i` is used for image so this might be confusing. I'm not sure what other letter would be best though, since we already use d for diameter and f for filename (in morphometrics CLI; though looking at that one, `-o` would probably have been more appropriate)

AxonDeepSeg/morphometrics/count_axons.py
```diff
@@ -0,0 +1,114 @@
...
+        dest='input_dir',
+        type=str,
+        help='Path to the folder containing all morphometric files.'
```
mathieuboudreau: Why have it by default only point at folders? why not have an option to point it directly to one file vs a folder?

docs/source/documentation.rst
```diff
@@ -696,6 +696,26 @@ To aggregate the morphometrics per subject, use the following command::
...
+~~~~~~~~~~~~~~~~~~~~~~~
+Use ``axondeepseg_filter`` to remove invalid morphometric measurements from one file or from all morphometric files in a folder. By default, the rules in ``AxonDeepSeg/morphometrics/filter.yaml`` are used and filtered files are saved
+with the suffix ``_filtered.xlsx``::
+
+   axondeepseg_filter -i folder_with_morphometrics
+
+Use ``--config`` to provide a custom rules file, ``--overwrite`` to replace the original files, or ``--update_masks`` to update the corresponding segmentation masks. 
+A custom ``filter.yaml`` file should contain ``myelinated`` and ``unmyelinated`` rule lists. Supported rules include ``valid-g-ratio-only``, ``axon-diam-gt`` (axon diameter greater than), ``solidity-gt`` (solidity greater than), and ``axon-area-lt`` (axon area less than). Use ``~`` as a threshold to disable that rule. Then, pass your custom rules file to the command line with ``--config``::
+
+   axondeepseg_filter -i folder_with_morphometrics --config my_custom_rules.yaml
```
mathieuboudreau, line 721: It would be helpful here to have a sample of what the custom rules should look like, i.e. an entire mini file as a code block that can be copy-pasted. I am trying to implement the rules files and I don't quite know how it needs to be formatted to setup min/max, etc

AxonDeepSeg/morphometrics/count_axons.py
hermancollin: the input *can* actually be images here

mathieuboudreau: oh ok - then we'd just need to edit the description here/in the doc; the documentation specified that it was a folder input

AxonDeepSeg/morphometrics/count_axons.py
hermancollin: you're right it should also accept a single file. The reason why I did it this way is because it feels like it makes more sense wrt the workflow (i.e. counting axons in a batch at once) but we could allow more granularity

yolaatar (approved):

mathieuboudreau (approved):
LGTM! Thanks!

